PR #8, opened 2018-10-23
Remove eval statements

jan-janssen:


jan-janssen:
@sudarsan1989 Do we have a test for create_surface?

sudarsan1989:
No. There does not seem to be any tests for pyiron at all. Should we create one?

PR #30, opened 2018-12-03
Get nph aniso to read pressure during langevin

liamhuber:
Old string formatting had fix nph aniso reading temperature variables instead of pressure variablse

### DIFF
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -110,7 +110,7 @@ run 0
             if langevin:
                 ensamble = 'nph'
 
-                fix_str = 'all {0} aniso {1} {2} {3}'.format(ensamble, str(temperature), str(temperature), str(delta_temp),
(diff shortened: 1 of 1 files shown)


liamhuber:
K, thanks. Once the checks go through I'll probably proceed with or without Sam's approval ;) as this is a bottleneck for Raynol right now. This is my first change to the PyIron source so I wanted at least one other set of dev eyes on it, also to practice the git flow.

liamhuber:
Actually, this raises a question @jan-janssen... Does having Raynol's PyIron env incorporating this request require a) him cloning this repo and pulling this change down, b) some pushed release from our end for the Garching distribution, or c) nothing special (i.e. Garching will automatically update to the master)?

jan-janssen:
@liamhuber The default would be that he needs his own repo. Still I merged the request and pull the latest changes to cmmc.

liamhuber:
Ok, thanks!

PR #48, opened 2019-01-04
All conda forge

jan-janssen:


jan-janssen:
At least for Linux the builds seem to be 25% faster - 2min.

PR #58, opened 2019-01-08
Potential

samwaseda:
This is actually much less than what is already done by Ahmed but I guess for tomorrow's pyiron meeting this is enough.


jan-janssen:
I would like to merge these changes with https://github.com/pyiron/pyiron/pull/56 so we have all new potential extensions in one module. But maybe that is something we can discuss in the pyiron meeting.

jan-janssen:
In the developer meeting we agreed to move these changes to the developer repository.

PR #59, opened 2019-01-08
Never write in from_hdf() if it is not necessary

jan-janssen:


jan-janssen:
I agree the important point for me was - do not write to the HDF5 file by default, first check if the entry already exists before writing it again.

sudarsan-surendralal:
> I agree the important point for me was - do not write to the HDF5 file by default, first check if the entry already exists before writing it again.

I see. I think this is the only place where we do this?

jan-janssen:
At least this is the place which crashed the calculation for Alexander Shapeev, that is how I found it in the first place.

PR #88, opened 2019-01-17
Langevin velocity corrected

samwaseda:
I also created `job.input.control.set_velocity`

### DIFF
--- a/pyiron/atomistics/job/atomistic.py
+++ b/pyiron/atomistics/job/atomistic.py
@@ -199,7 +199,7 @@ class AtomisticGenericJob(GenericJobCore):
         self._generic_input.remove_keys(['max_iter', 'pressure', 'temperature', 'n_ionic_steps', 'n_print', 'velocity'])
 
     def calc_md(self, temperature=None, pressure=None, n_ionic_steps=1000, time_step=None, n_print=100, delta_temp=1.0,
-                delta_press=None, seed=None, tloop=None, rescale_velocity=True, langevin=False):
+                delta_press=None, seed=None, tloop=None, initial_temperature=True, langevin=False):
         self._generic_input['calc_mode'] = 'md'
         self._generic_input['temperature'] = temperature
         self._generic_input['n_ionic_steps'] = n_ionic_steps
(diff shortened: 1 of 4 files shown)


jan-janssen:
Do we need a separate command for `set_initial_velocity` ? From my perspective it is only reasonable when running MD simulation, therefore it should be part of the `calc_md` command should not it?

samwaseda:
Well I guess the point is it is generally not required to set the initial velocity separately, so I guess the current implementation is good enough. If something like Lifang's case appears (where the initial structure already is not at energy minimum), then I guess it's okay to have a separate command to create the velocity field. Also since the command in LAMMPS itself offers more possibilities, I also wanted to offer more flexibility to the user (which can be seen in further options in the function).

jan-janssen:
But could not we achieve the same by having an integer index for rescaling velocity? It could be 0 if no velocity is set 1 in Lifangs case and 2 by default?

samwaseda:
In my opinion Lifang's case is already extremely special. If we include Lifang's case in the standard `calc_md` implementation, I find it a bit confusing to the user and also I'd wonder why any other temperature value is not set by default.

liamhuber:
I would also like a scalar option for initializing the velocity, and have used a similar technique in my own custom MD codes in the past.

The reason I like it is that I sometimes feel that a factor of 2 is overkill. Far from the melting point it's perfect because it optimizes the partition of energy, but if you're trying to initialize a calculation at 80 or 90% of $T_\mathrm{melt}$, I always get nervous that such high initial velocities may introduce undesired defects in the first few dozen steps.

samwaseda:
Alternatively we can also put something like `initial_temperature` (instead of `rescale_velocity`), whose default value is set to `None`, in which case the velocity is set to twice the target temperature. And if the `initial_temperature` is 0, then no temperature creation and any other value to get a user defined temperature.

I'm saying this partly because what I implemented above carries the risk, that the user sets the initial velocity and then set `calc_minimize()` without setting the `rescale_velocity` to `False`, in which case the velocity field will be overwritten.

liamhuber:
I like the idea of an initial temperature field instead of a rescale. It obscures the physics of energy equipartition a little bit, but if you already know *why* you're setting a higher initial temperature, it makes it easier to choose a good value relative to the approximate melting point.

PR #90, opened 2019-01-18
Typo fix

liamhuber:


liamhuber:
I just refactored the name to have the "c" in "interactive". I don't think the code needs review, so much as I thought people working on interactive stuff (i.e. you, @samwaseda), should be aware that I changed the name in master.

What's the etiquette on little stuff like this? I don't see any reason to wait on patching a little fix into master, but it seems both inane to request a review and reckless to merge anything 100% solo.

PR #93, opened 2019-01-19
Interactive

jan-janssen:


liamhuber:
- typo: `froces` -> `forces` @pyiron/atomistics/master/phonopy.py-204, 206, 207
- documentation: docstrings for pyiron/atomistics/master/structure.py? It looks suspiciously like new code ;)
- typo: `childs` -> `children` @pyiron/base/master/generic.py-322
- [platform compatibility](https://stackoverflow.com/questions/10918682/concatenate-path-platform-independent): Replace `p1 + "/" + p2` with `os.path.join(p1, p2)` @pyiron/base/master/generic.py-184, 188, 213.
- documentation: docstring templated but empty at pyiron/base/master/generic.py-333

jan-janssen:
Thanks for the feedback, about the platform compatibility - that is correct but inside the HDF5 file we always use '/' independent of the operation system notation. 

For the others I opened separate pull requests: 
https://github.com/pyiron/pyiron/pull/100
https://github.com/pyiron/pyiron/pull/96

PR #103, opened 2019-01-23
Mic

samwaseda:


pyiron/base/job/interactive.py
```diff
@@ -263,6 +263,16 @@ def interactive_open(self):
         """
         self.server.run_mode.interactive = True
 
+    def hd_copy(self, hdf_old, hdf_new, exclude_groups = []):
```
jan-janssen: Please add DocStrings for every new function

PR #113, opened 2019-01-28
excluded interactive from exclude_group so that the interactive class…

samwaseda:
… is also saved.

This follows a discussion between Jan and me.


jan-janssen:
We already have a pull request for this: https://github.com/pyiron/pyiron/pull/111/files

PR #115, opened 2019-01-28
Set mixing parameters

samwaseda:


jan-janssen:
@samwaseda Can you add Docstrings to this function?

samwaseda:
you told me it's not necessary to put Docstrings if the function returns only not implemented error

Issue #117, opened 2019-01-29
Error in get_broyden_mixing()


sudarsan-surendralal:
So, I had a calculation which ran successfully but the parsing failed because the OUTCAR parser could not get the broyden mixing mesh

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-340-2919e271a3c5> in <module>()
----> 1 oc.get_broyden_mixing_mesh(ham_repair.working_directory+"/OUTCAR")

/cmmc/u/chandu/programs/pyiron_mpie/pyiron/pyiron/vasp/outcar.py in get_broyden_mixing_mesh(filename)
    562             for line in f.readlines():
    563                 if trigger in line:
--> 564                     return int(line.split()[4])
    565 
    566     @staticmethod

ValueError: invalid literal for int() with base 10: '******'
```
I checked the line and of course it gives out garbage although the calculation is fine
```

Broyden mixing: mesh for mixing (old mesh)
   NGX =123   NGY =103   NGZ = 81
  (NGX  =500   NGY  =420   NGZ  =320)
  gives a total of ****** points

```
Currently we try to parse the line `gives a total of ****** points`. But I realized from other calculations that this is the same as `NGX*NGY*NGZ` where `NGX =123   NGY =103   NGZ = 81`. Is this always the case? Then I can simply parse this line instead.

@jan-janssen I think you need this for your analysis right? Can you comment on this? Have you ever encountered this error?

jan-janssen:
Maybe you can check if it is  '******' and then return 0 ? That should fix the issue. Or maybe even return None that would be more correct.

PR #118, opened 2019-01-29
Try to fix https://github.com/pyiron/pyiron/issues/117

jan-janssen:


jan-janssen:
@sudarsan-surendralal Can you test this fix?

sudarsan-surendralal:
I used exception handling instead of if https://github.com/pyiron/pyiron/pull/118/commits/14d3a87d6d7e0077a26134a15b85e8e7c988c44d. Added a unittest as well

jan-janssen:
@sudarsan-surendralal I guess exception handling is not so nice, if we can identify the error before hand we should not need the exception.

jan-janssen:
It should be because the number is larger than 99999

jan-janssen:
Is OUTCAR_9 just a copy of OUTCAR_1 with the 'Broyden mixing:' line replaced?

jan-janssen:
For the job that crashed can you please copy the following lines: 

```
 Broyden mixing: mesh for mixing (old mesh)
   NGX =  9   NGY =  9   NGZ =  9
  (NGX  = 40   NGY  = 40   NGZ  = 40)
  gives a total of    ****** points 
```

PR #125, opened 2019-02-01
output formats for sphinx updated

samwaseda:


samwaseda:
Sorry I was wrong.

jan-janssen:
> Sorry I was wrong.

So to get this right, the time is LAMMPS is also fs, right?

PR #139, opened 2019-02-12
Langevin fix

liamhuber:
Adds the "zero yes" keyword argument to Lammps Langevin thermostat calls and renames some Lammps MD variables.

The key issue was that the Langevin thermostat adds random forces to the system, and without "zero yes" this can produce a net force on the centre of mass, causing drift over long simulations. Lammps already has the built-in keyword to prevent this. I'm not even making it optional because I don't see why you would ever want your CoM to drift!

I also renamed the `delta_temp` and `delta_press` variables to `temperature_damping` and `pressure_damping`, respectively. This is more in line with our "full name" naming scheme, and also (IMO) more intuitive since they have units of time and not units of $\Delta T$ or $\Delta P$. The original variable names are still usable but print a deprecation warning.

### DIFF
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -144,7 +144,7 @@ class LammpsControl(GenericParameters):
             print("WARNING: `delta_temp` is deprecated, please use `temperature_damping`")
             temperature_damping = delta_temp
         if delta_press is not None:
-            print("WARNING: `delta_temp` is deprecated, please use `temperature_damping`")
+            print("WARNING: `delta_press` is deprecated, please use `pressure_damping`")
             temperature_damping = delta_press
 
         if time_step is not None:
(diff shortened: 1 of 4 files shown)


pyiron/lammps/control.py
```diff
@@ -139,6 +141,12 @@ def calc_md(self, temperature=None, pressure=None, n_ionic_steps=1000, time_step
                                                for the initial temperature.
             langevin (bool): (True or False) Activate Langevin dynamics
         """
+        if delta_temp is not None:
+            warnings.warn("WARNING: `delta_temp` is deprecated, please use `temperature_damping`")
```
jan-janssen: I guess warnings are more suitable than print statements.

pyiron/lammps/control.py
```diff
@@ -139,6 +141,12 @@ def calc_md(self, temperature=None, pressure=None, n_ionic_steps=1000, time_step
...
+            temperature_damping = delta_temp
+        if delta_press is not None:
+            warnings.warn("WARNING: `delta_press` is deprecated, please use `pressure_damping`")
```
jan-janssen: I guess warnings are more suitable than print statements.

PR #157, opened 2019-02-26
check setup created

jan-janssen:


pyiron/base/job/generic.py
```diff
@@ -997,6 +1005,8 @@ def _run_if_new(self, debug=False):
             debug (bool): Debug Mode
         """
         self.validate_ready_to_run()
+        if self.server.run_mode.queue:
```
jan-janssen: Why do you only run the test if `run_mode` queue is selected?

samwaseda: The tests should be performed only when the job is submitted right?

jan-janssen: To have a consistent experience for the user I would recommend that we always run the check.

samwaseda: We decided together that it should be run only when the job is submitted

PR #159, opened 2019-02-27
Outcar

jan-janssen:
Optimise the parser to only read the outcar file once.


jan-janssen:
@sudarsan-surendralal exactly, this way we can still use it to parse files for single entries but when we parse the full file it is only opened once. The next step would be iterating the lines only once, but I guess that is a bit more complex.

PR #160, opened 2019-02-28
get_equivalent_voronoi_vertices can now get length and angle toleranc…

samwaseda:
…e to make in agreement with get_symmetry


jan-janssen:
While I like the function I was wondering if it maybe makes sense to move the analysis of atomistic environments like `neighbors`, `ovito`, `voronoi` and so on to a separate module and only load the functions in the atoms class. At the moment the atoms class is one of the largest and it might be a bit confusing for new developers. @sudarsan-surendralal and @samwaseda what do you think?

samwaseda:
It is extremely true that both the creation and the analysis of structures are done within the same class. So yes, I agree.

PR #163, opened 2019-03-06
Warn about special characters in project names

jan-janssen:


jan-janssen:
`isidentifier()` is not available in Python 2

PR #170, opened 2019-03-12
Md backwards compatibility

liamhuber:
Fixing a bug where forced application of the old `delta_press` didn't work.

Also added rescaling of pressure units to accommodate non-metal Lammps units, and some minor documentation/beautification.

### DIFF
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -167,6 +167,10 @@ class LammpsControl(GenericParameters):
                                                 'pressure': GPa_to_Pa}}
         time_units = lammps_unit_conversions[self['units']['time']]
         pressure_units = lammps_unit_conversions[self['units']]['pressure']
+        # No need for temperature conversion; pyiron and all available Lammps units are both in Kelvin
+        # (well, except unitless Lennard-Jones units...)
+        if self['units'] == 'lj':
+            raise NotImplementedError
 
         if time_step is not None:
             try:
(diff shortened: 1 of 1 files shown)


pyiron/lammps/control.py
```diff
@@ -147,49 +147,66 @@ def calc_md(self, temperature=None, pressure=None, n_ionic_steps=1000, time_step
...
-                                  'real': 1,
-                                  'electron': 1}
-
+        # Conversion factors for transfroming pyiron units to Lammps units
+        fs_to_ps = 1.e-3
+        fs_to_s = 1.e-12
+        GPa_to_bar = 1.e4
+        GPa_to_Pa = 1.e9
+        GPa_to_barye = 1.e10
+        GPa_to_atm = 9869.232667160128  # digits c.f. http://www.kylesconverter.com/pressure/gigapascals-to-atmospheres
```
jan-janssen: Can not we do the unit conversion with scipy constants?

liamhuber: Certainly. I see to do it like this:

`import scipy.constants as sconst

fs_to_ps = sconst.femto / sconst.pico
fs_to_s = sconst.femto / 1.
GPa_to_bar = sconst.giga * 1. / sconst.bar
GPa_to_Pa = sconst.giga 
GPa_to_barye = sconst.giga * 1. / (1.e-6 * sconst.bar)  # Lammps is in "barye"
GPa_to_atm = sconst.giga * 1. / sconst.atm`

...but to me this is quite ugly looking compared to just putting the number. Is there a more elegant way to phrase it?

jan-janssen: I am not aware if there is a more elegant way - so I guess for now we can just implement it like this.

liamhuber: Amended.

PR #176, opened 2019-03-14
Trying to simplify the master jobs

jan-janssen:


jan-janssen:
@sudarsan-surendralal Can you check if this fixes your issues with the SerialMaster?

sudarsan-surendralal:
Yes definitely. But a large job is running right now. Do you think it is safe to merge as the job is running? Or I can wait to crash and then check this.

jan-janssen:
It is only loading the child jobs from HDF5 when they are actually used while before it was always loading them. Still I guess it is save to wait until your current calculation crashes before trying this branch.

PR #179, opened 2019-03-18
Volumertic data

sudarsan-surendralal:
Better and more efficient parsing of the volumetric data in VASP (CHGCAR, LOCPOT etc.)


jan-janssen:
@sudarsan-surendralal I see you removed two of the vasprun.xml files, can you add a comment why they are no longer necessary?

sudarsan-surendralal:
They were redundant. They already exist [here](https://github.com/pyiron/pyiron/tree/master/tests/static/vasp_test_files/vasprun_samples). So this saved us a some memory

PR #193, opened 2019-03-21
Plot3d colours

liamhuber:
Modifications `plot3d` colouring capability.

- Colouring by the scalar field now lets the user define a colourmap (default is blue-red divergent) and limits for which scalars should be mapped to the extrema of the colourmap.
- An explicit per-atom list of colours can also be given (HTML names or hex codes (sort of...some colours come out wrong)). 
- The rest of the functionality (e.g. vector colouring, axes, etc.) should be the same, but I did make code modifications for streamlining and readability.
- Docstrings everywhere.

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1163,18 +1163,24 @@ class Atoms(object):
             show_cell (bool): Whether or not to show the frame. (Default is True.)
             show_axes (bool): Whether or not to show xyz axes. (Default is True.)
             camera (str): 'perspective' or 'orthographic'. (Default is 'perspective'.)
-            spacefill (bool): Whether to use a space-filling or ball-and-stick representation. (Default is True, use space-filling atoms.)
+            spacefill (bool): Whether to use a space-filling or ball-and-stick representation. (Default is True, use
+                space-filling atoms.)
             particle_size (float): Size of the particles. (Default is 1.)
-            select_atoms (ndarray): Indices of atoms to show, either as integers or a boolean array mask. (Default is None, show all atoms)
+            select_atoms (ndarray): Indices of atoms to show, either as integers or a boolean array mask. (Default is
+                None, show all atoms)
             background (str): Background color. (Default is 'white'.)
             color_scheme (str): NGLView color scheme to use. (Default is None, color by element.)
-            colors (ndarray): A per-atom array of HTML color names or hex color codes to use for atomic colors. (Default is None, use coloring scheme.)
+            colors (ndarray): A per-atom array of HTML color names or hex color codes to use for atomic colors.
+                (Default is None, use coloring scheme.)
             scalar_field (ndarray): Color each atom according to the array value (Default is None, use coloring scheme.)
-            scalar_start (float): The scalar value to be mapped onto the low end of the color map (lower values are clipped). (Default is None, use the minimum value in `scalar_field`.)
-            scalar_end (float): The scalar value to be mapped onto the high end of the color map (higher values are clipped). (Default is None, use the maximum value in `scalar_field`.)
+            scalar_start (float): The scalar value to be mapped onto the low end of the color map (lower values are
+                clipped). (Default is None, use the minimum value in `scalar_field`.)
+            scalar_end (float): The scalar value to be mapped onto the high end of the color map (higher values are
+                clipped). (Default is None, use the maximum value in `scalar_field`.)
             scalar_cmap (matplotlib.cm): The colormap to use. (Default is None, giving a blue-red divergent map.)
             vector_field (ndarray): Add vectors (3 values) originating at each atom. (Default is None, no vectors.)
-            vector_color (ndarray): Colors for the vectors (only available with vector_field). (Default is None, vectors are colored by their direction.)
+            vector_color (ndarray): Colors for the vectors (only available with vector_field). (Default is None,
+                vectors are colored by their direction.)
 
             Possible NGLView color schemes:
               " ", "picking", "random", "uniform", "atomindex", "residueindex",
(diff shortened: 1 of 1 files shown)


pyiron/atomistics/structure/atoms.py
```diff
@@ -1058,115 +1058,208 @@ def analyse_phonopy_equivalent_atoms(atoms):
...
+            vector_color (ndarray): Colors for the vectors (only available with vector_field). (Default is None, vectors are colored by their direction.)
+
+            Possible NGLView color schemes:
               " ", "picking", "random", "uniform", "atomindex", "residueindex",
               "chainindex", "modelindex", "sstruc", "element", "resname", "bfactor",
               "hydrophobicity", "value", "volume", "occupancy"
     
         Returns:
-    
+            (NGLWidget): The NGLView widget itself, which can be operated on further or viewed as-is.
```
sudarsan-surendralal: `nglview.NGLWidget` for proper docstring indexing

sudarsan-surendralal: Btw, thank you so much for the detailed and clear docstrings for the`plot3d()`! :)

pyiron/atomistics/structure/atoms.py
```diff
@@ -1213,17 +1227,26 @@ def plot3d(self, show_cell=True, show_axes=True, camera='perspective', spacefill
...
-
-            for elem, num in set(list(zip(parent_basis.get_chemical_symbols(), parent_basis.get_atomic_numbers()))):
-                view.add_spacefill(selection='#' + elem,
-                                   radius_type='vdw',
-                                   color_scheme=color_scheme,
-                                   radius=particle_size * (0.2 + 0.1 * np.sqrt(num)))
+            elemental_data_list = list(zip(elements, atomic_numbers))
+
+            if scalar_field is not None:
+                assert(color_scheme is None)
```
sudarsan-surendralal: Apparently runtime `assert` statements should be minimized [(see here)](http://web.mit.edu/6.005/www/fa16/classes/08-avoiding-debugging/#what_not_to_assert). So maybe an `if` statement followed by raising a `ValueError` may be better?

pyiron/atomistics/structure/atoms.py
```diff
@@ -1058,115 +1058,208 @@ def analyse_phonopy_equivalent_atoms(atoms):
...
         else:
-            scalar_field = (scalar_field-np.min(scalar_field))/(np.max(scalar_field)-np.min(scalar_field))
+            scalar_field = (scalar_field - np.amin(scalar_field)) / np.ptp(scalar_field)
+
         for i, p in enumerate(positions):
             if rotation is not None:
                 p = p.dot(rotation)
                 
-            pdb_str += self._ngl_write_atom(i, elements[i], group=elements[i], num2=i, coords=p, c0=scalar_field[i], c1=0.0)
+            pdb_str += self._ngl_write_atom(i, elements[i],
```
sudarsan-surendralal: Is it possible to avoiding looping over the atoms altogether? For example, the positions can be formated into a string format of your choice using some nice numpy routines

`np.array2string(np.char.mod(('%10.3f', '%7.3f', '%7.3f'), self.positions)).replace("]", "").replace("[", "")`

pyiron/atomistics/structure/atoms.py
```diff
@@ -1058,115 +1058,208 @@ def analyse_phonopy_equivalent_atoms(atoms):
...
         pdb_str += 'MODEL     1\n'
+
         if scalar_field is None:
             scalar_field = np.ones(len(positions))
         else:
-            scalar_field = (scalar_field-np.min(scalar_field))/(np.max(scalar_field)-np.min(scalar_field))
+            scalar_field = (scalar_field - np.amin(scalar_field)) / np.ptp(scalar_field)
+
         for i, p in enumerate(positions):
             if rotation is not None:
```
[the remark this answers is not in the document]
liamhuber: Agreed. It's moved.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1058,115 +1058,208 @@ def analyse_phonopy_equivalent_atoms(atoms):
...
-        return 'ATOM {:>6} {:>4} {:>4} {:>5} {:10.3f} {:7.3f} {:7.3f} {:5.2f} {:5.2f} {:>11} \n'.format(num, species, group, num2, x, y, z, c0, c1, species)
+        return 'ATOM {:>6} {:>4} {:>4} {:>5} {:10.3f} {:7.3f} {:7.3f} {:5.2f} {:5.2f} {:>11} \n'.format(
+            num, species, group, num2, x, y, z, occupancy, temperature_factor, species)
     
     def _ngl_write_structure(self, elements, positions, cell, scalar_field=None):
+        """
+        Turns structure information into a NGLView-readable protein-database-formatted string.
+
+        Args:
+            elements (ndarray/list): Element symbol for each atom.
```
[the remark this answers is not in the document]
liamhuber: You bet! Didn't know how to do it.

pyiron/atomistics/structure/atoms.py
liamhuber: Also done.

np. Glad to hear they're sufficiently clear. One can be detailed, but sometimes more detail doesn't help clarity ;)

pyiron/atomistics/structure/atoms.py
liamhuber: Noted. I actually replaced them with warnings instead.

pyiron/atomistics/structure/atoms.py
liamhuber: Yes, I'm sure it's possible, but I'm actually going to skip this one. I think any structure big enough makes this for-loop give a noticeable performance hit is a structure which will cause problems anyhow. In the meantime, I find the for-loop much more readable than a string-hacked array. If we run into performance issues we can change it later, but it's always been this way so I think we're safe ;)
(Note that I do get some performance hit on my new re-coloring which calls a separate `add_spacefill` for each atom, but it's about 0.5-1s for a 300 atom cell so I'll live with it for now.)

PR #203, opened 2019-04-15
Debug repeat

sudarsan-surendralal:
A very minor comparability debug (list vs numpy)

### DIFF
--- a/pyiron/atomistics/structure/sparse_list.py
+++ b/pyiron/atomistics/structure/sparse_list.py
@@ -184,7 +184,7 @@ class SparseList(object):
         return self.__class__(sliced_dict, default=self._default, length=len(ind_list))
 
     def __setitem__(self, key, value):
-        if isinstance(key, int):
+        if isinstance(key, (int, np.int32, np.int64, np.int_)):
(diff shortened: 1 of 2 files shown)


pyiron/atomistics/structure/sparse_list.py
```diff
@@ -184,7 +184,7 @@ def __getitem__(self, item):
         return self.__class__(sliced_dict, default=self._default, length=len(ind_list))
 
     def __setitem__(self, key, value):
-        if isinstance(key, int):
+        if isinstance(key, (int, np.int32, np.int64, np.int_)):
```
jan-janssen: Is there a specific reason why you need to check for `np.int32, np.int64, np.int_` ? should not `(int, np.integer)` be sufficient ?

sudarsan-surendralal: You're right. I was not aware of this. Changes have been made in the new pull request https://github.com/pyiron/pyiron/pull/204/commits/158f8257aff1ff647909c9a2a31817af9f659681 (all over the codebase)

PR #211, opened 2019-04-30
ref_job gives a warning if the job is not set to interactive

samwaseda:
As discussed [here](https://gitlab.mpcdf.mpg.de/pyiron/pyiron-support/issues/137)

### DIFF
--- a/pyiron/atomistics/job/interactivewrapper.py
+++ b/pyiron/atomistics/job/interactivewrapper.py
@@ -65,6 +65,8 @@ class InteractiveWrapper(GenericMaster):
         Args:
             ref_job (GenericJob): reference job
         """
+        if not ref_job.server.run_mode.interactive:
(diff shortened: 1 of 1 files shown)


samwaseda:
Actually, does it ever make sense that the wrapper is not interactive? -> Can't we simply set default to interactive?

jan-janssen:
Langevin and MonteCarlo both do not use the interactive run mode. At the same time I noticed that SxExtOpt can also be executed in non interactive mode, for example @sudarsan-surendralal did this to see if it accelerates the convergence for VASP and this would be no longer possible with these changes. So the more I think about this I am against this option and my recommendation would be to set the ref_job to interactive for those job classes where it is required.

samwaseda:
So I guess what you mean is this can be merged as it is?

PR #215, opened 2019-05-07
Enable python script execution with multiple cores

jan-janssen:


jan-janssen:
It only works when the cores are defined before the script_path and it does not work with the queuing system after all.

PR #223, opened 2019-05-10
Migration to readthedocs.org

jan-janssen:


jan-janssen:
The documentation is now available at https://pyiron.readthedocs.io/en/readthedocs/ and while the module index still redirects to the old page, it is also available on readthedocs https://pyiron.readthedocs.io/en/readthedocs/py-modindex.html . The advantage of readthedocs is that we can host multiple versions of documentation.

PR #230, opened 2019-05-20
Wrap atoms

sudarsan-surendralal:
Option to wrap atoms before returning them!


liamhuber:
Nice! Looks good to me.

PR #231, opened 2019-05-21
Pressure

samwaseda:
Three pressure components can be set independently. `None` is also allowed as a component, in which case the box length is fixed in this direction.


jan-janssen:
I guess we should have two separate pull requests, one for custom potential and one for the pressure. Still the pressure part looks fine to me.

samwaseda:
I see that I also submitted the custom potential class, but I guess this shouldn't affect anything, since only an independent class is **appended**, so that it shouldn't affect the base

samwaseda:
Good that you noticed it as well :D

liamhuber:
I like it.

I futzed with the logic a little to remove some duplicate code. Sam, do you have a small test notebook to make sure all the ensembles are working as expected, with which you can proof* my changes? If not I'll throw one together, but I don't want to duplicate work if it exists...

*fyi, this is actually not a valid use of "proof", just some Denglish that I think ought to be incorporated into mainstream English.

samwaseda:
[here](https://gitlab.mpcdf.mpg.de/pyiron/pyiron.pages.mpcdf.de/blob/master/meetings/2019-05-22-pyiron-meeting/osamu-waseda/pressure.ipynb)

PR #237, opened 2019-05-27
Store the temperatures as `"temperature"` rather than `"temperatures"`

sudarsan-surendralal:
This is necessary for the sake of consistency.


sudarsan-surendralal:
Yeah, I guess we might need a hack! I'll come up with something before we merge

sudarsan-surendralal:
So I added a long overdue set of unit tests and added backward compatibility for the temperature!

sudarsan-surendralal:
Shall I merge this?

PR #246, opened 2019-06-14
Implement vasp sorter

jan-janssen:


jan-janssen:
That would be perfect and I guess the doctoring has to be updated as well.

PR #253, opened 2019-06-25
Fix busy jobs

jan-janssen:


jan-janssen:
Somehow the `SerialMaster` fails with Python 3.7 ... there seems to be something wrong.

PR #257, opened 2019-07-10
Interactive lammps for water potentials

sudarsan-surendralal:
Modifying interactive LAMMPS to be able to run water potentials. Mainly involving bond creation and overcoming the limitations imposed by LAMMPS when defining them! (Currently works for water potentials but will have to be generalized for other molecules later)


liamhuber:
Nice, it's working fine for me! Thanks very much for the fast turn-around on this, Sudarsan, I'm sure Stefan and Florian will appreciate it.

Before merging it, I would like to suggest a couple simplifications/clean-ups.

First, at line 156, my understanding right now is that `self._interactive_lib_command(line.split('\n')[0])` is called *either* if the style is not full, *or* if the style is full but neither 'kspace' nor 'pair' appear in the `line`.  If so, we should be able to simplify this block as follows:
```
style_full = self.input.control['atom_style'] == 'full'
for line in self.input.potential.get_string_lst():
    if len(line) > 2:
        for potential in potential_lst:
            if potential[0] in line:
                line = line.replace(potential[0], potential[1])
            if not (style_full and ("kspace" in line or "pair" in line)):
                self._interactive_lib_command(line.split('\n')[0])
           # If style was full and 'kspace' or 'pair' was present and we do nothing
    if style_full:
        self._interactive_water_setter()
```

Second, at lines 251, 252, and 254 I'm getting PEP8 warnings; was this new indentation intentional?

Third and finally, I think it's 100% ok that this is working right now only for water and not for all molecules. So that this is clear to future developers, could you please add a short docstring to `_interactive_water_setter` explaining at a high level what it's doing and pointing out that a more general but otherwise similar function is needed for molecules? It might be nice to also add a `# needs to be generalized for all non-water molecules` in the line 157-block where `self._interactive_water_setter()` is called.

Again, it's great that this new functionality is running so soon!!! :D

PR #259, opened 2019-07-11
Pressure axis

Ahmed-Aslam:
Issue : Removing redundant axis from output/pressures. 
It is solved by converting list of array to array of proper dimensions.

### DIFF
--- a/pyiron/lammps/base.py
+++ b/pyiron/lammps/base.py
@@ -266,7 +266,7 @@ class LammpsBase(AtomisticGenericJob):
         Returns:
 
         """
-        print('Collecting',job_name )
+        print('Collecting', self.job_name)
         self.input.from_hdf(self._hdf5)
         if os.path.isfile(self.job_file_name(file_name="dump.h5", cwd=self.working_directory)):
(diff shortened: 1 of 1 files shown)


jan-janssen:
Try to do the conversion before writing to HDF5, just like it is done for the temperature: https://github.com/pyiron/pyiron/blob/pressure_axis/pyiron/lammps/base.py#L432

liamhuber:
As Jan suggests, let's fix the problem at its source.

In principle there's nothing *wrong* with this solution, and it's very pragmatic. However, it covers over the problem by adding an extra line of code. For a small project where you need to move fast and get things done this is an excellent solution. For a large project which needs to be maintained by a whole team of people, it adds one more line of code that is slightly mysterious and needs to be researched and understood if anything goes wrong later. Better then to dig in and remove the problem to begin with :)

liamhuber:
I was reminded today that this principle has a fancy software design name: [technical debt](https://en.wikipedia.org/wiki/Technical_debt)

jan-janssen:
@samwaseda as far as I understand https://github.com/pyiron/pyiron/pull/270 this pull request is no longer required. Can we close this pull request?

PR #260, opened 2019-07-11
values changed for e_tol and f_tol from interactive 

uday-gajera:
e_tol = 0
 f_tol = 0.0001


liamhuber:
@uday-gajera, better in general to have two separate pull requests for these two very distinct issues (the camera and minimization), but since they're both fairly simple fixes adjusting default parameters I don't see a problem merging them together here :)

PR #261, opened 2019-07-11
calc_minimize convergence values changed and function doc stringed.

raynol-dsouza:
calc_minimize convergence values changed and function doc stringed.

### DIFF
--- a/pyiron/lammps/interactive.py
+++ b/pyiron/lammps/interactive.py
@@ -189,7 +189,17 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
         self._reset_interactive_run_command()
         self.interactive_structure_setter(self.structure)
 
-    def calc_minimize(self, e_tol=1e-8, f_tol=1e-8, max_iter=1000, pressure=None, n_print=100):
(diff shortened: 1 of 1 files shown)


liamhuber:
1) Git subtlety: please merge in Uday's changes so that this pull is just for this pull.

2) I'd like some more clarity in the doc strings. 
- Especially, what does 'convergence criterion' mean? For energy I guess it's just magnitude of the difference from one step to the next, but maybe I'm wrong. For forces, is it the max force magnitude among all atoms? Is it the magnitude of the N_atoms x 3 force vector? 
- Pressure default should explicitly indicate that 'None' means constant volume (aka isochoric) calculation will be used. 
- Also, the pressure units are *usually* eV / ang^3, but I guess somewhere there's a place to change pyiron units and there's some way for the user to change this, in which case we try to avoid hard-coding units into the docstrings. (I'm sure I'm guilty of doing this too!) @jan-janssen, can you comment about this possibility of unit changing?
- Finally, and pretty minorly, I would appreciate some more explicit description of what "printing" or "dumping" is. In particular, I guess "dump" is a phrase which will be very informative to Lammps users, but might not be so helpful for people coming from a Vasp background.

PR #262, opened 2019-07-12
Output displacement

samwaseda:
Discussion that I had initiated in gitlab. This of course considers the periodic boundary conditions as well.

### DIFF
--- a/pyiron/atomistics/job/atomistic.py
+++ b/pyiron/atomistics/job/atomistic.py
@@ -670,9 +670,16 @@ class GenericOutput(object):
     @property
     def volume(self):
         return self._job['output/generic/volume']
-    
+
     @property
     def displacements(self):
+        """
+        Output for 3-d displacements between successive snapshots, with minimum image convention.
+        For the total displacements from the initial configuration, use total_displacements
+        This algorithm collapses if:
+        - the ID's are not consistent (i.e. you can also not change the number of atoms)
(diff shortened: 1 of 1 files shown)


samwaseda:
I had already indicated this point, but this algorithm collapses if:

- the ID's are not consistent (i.e. you can also not change the number of atoms)
- there are atoms which move by more than half a box length in any direction within two snapshots (due to periodic boundary conditions)
- the cell size is not constant (NB: technically it doesn't collapse since it simply uses the initial cell size, and anyway you should define yourself what "displacement"  means when the cell size changes)

pyiron/atomistics/job/atomistic.py
```diff
@@ -670,6 +670,21 @@ def unwrapped_positions(self):
...
+    def displacements(self):
+        displacement = np.tensordot(self.positions,
+                                    np.linalg.inv(self._job.structure.cell), axes=([2,0]))
+        displacement -= np.append(self._job.structure.scaled_positions,
+                                  displacement).reshape(len(self.positions)+1,
+                                                        len(self._job.structure), 3)[:-1]
+        displacement -= np.rint(displacement)
+        displacement = np.tensordot(displacement, self._job.structure.cell, axes=([2,0]))
+        return displacement
+
```
sudarsan-surendralal: Remove the whitespace since Codeacy complains!

PR #264, opened 2019-07-14
lammps disable output to command line in interacrtive mode

jan-janssen:


jan-janssen:
I guess we should have `self._disable_lmp_output = True`  instead of `False`. Currently this only introduces the option to disable the output, but from my perspective it should be disabled by default.

liamhuber:
I would be happy with this, but I thought for now it's safer to merge something that doesn't effect the default behaviour. I don't see an issue with making lammps quiet by default, but maybe we can bring it up at the next developers meeting in case there's an application we're both missing.

PR #294, opened 2019-07-24
Workaround genericparameter bug

samwaseda:


jan-janssen:
This is now implemented in https://github.com/pyiron/pyiron/pull/292

PR #307, opened 2019-07-27
Felix hdf

jan-janssen:


jan-janssen:
@FeLoch Can you check if this branch works for you? If that is the case and @sudarsan-surendralal accepts these changes we can merge them into the master branch.

PR #316, opened 2019-08-01
Lmp dump parser

samwaseda:
1. New dump parser implemented (20x faster than the original one)
2. `dump.out` in the unit test is replaced with a new one, which is compatible with the water example given in the test functions
3. `output/generic/positions` are now **wrapped**, which makes more sense since we also have `output/generic/unwrapped_positions`.


tests/lammps/test_lammps.py
```diff
@@ -138,6 +140,40 @@ def test_lammps_water(self):
...
+        self.job_water_dump.structure = water
+        self.job_water_dump.potential = 'H2O_tip3p'
+        self.job_water_dump.calc_md(temperature=350, initial_temperature=350, time_step=1, n_ionic_steps=1000, n_print=200, pressure=0)
+        file_directory = os.path.join(self.execution_path, "..", "static", "lammps_test_files")
+        self.job_water_dump.restart_file_list.append(os.path.join(file_directory, "log.lammps"))
+        self.job_water_dump.restart_file_list.append(os.path.join(file_directory, "dump.out"))
+        self.job_water_dump.run(run_mode="manual")
+        self.job_water_dump.status.collect = True
+        self.job_water_dump.run()
+        positions = np.loadtxt(os.path.join(file_directory, 'positions_water.dat'))
```
sudarsan-surendralal: You have to add `positions_water.dat` and `forces_water.dat` to the repository for this to work

pyiron/lammps/base.py
```diff
@@ -544,6 +545,47 @@ def collect_dump_file(self, file_name="dump.out", cwd=None):
...
+        positions = unwrapped_positions-np.floor(unwrapped_positions)
+        unwrapped_positions = np.einsum('ikj,ilk->ilj', cells, unwrapped_positions)
+        output['unwrapped_positions'] = np.einsum('ijk,kl->ijl', unwrapped_positions, rotation_lammps2orig)
+        positions = np.einsum('ikj,ilk->ilj', cells, positions)
+        output['positions'] = np.einsum('ijk,kl->ijl', positions, rotation_lammps2orig)
+        with self.project_hdf5.open("output/generic") as hdf_output:
+            for k,v in output.items():
+                hdf_output[k] = v
+
+    def collect_dump_file_old(self, file_name="dump.out", cwd=None):
```
sudarsan-surendralal: Now that I think it, let's get rid of `collect_dump_file_old`. We can always get this back from git. Removing this will also increase our code coverage! ;)

PR #329, opened 2019-08-05
bugfix

samwaseda:
If the path contains dots ('.'), the previous version failed.


jan-janssen:
We can merge these changes, still it is not recommended to have `.` in your path because this breaks pyiron, when pyiron tries to navigate through your path using auto completion.

PR #333, opened 2019-08-07
Get majority species

samwaseda:
Number of atoms of the majority species was wrong. And there was no DocString

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -2199,7 +2199,7 @@ class Atoms(object):
         # print "ref_id: ", ref_id_list
         return eq_atoms, trans_vec, rot_vec, id_vec, ref_id_list
 
-    def get_majority_species(self):
+    def get_majority_species(self, return_count=False):
         """
         This function returns the majority species and their number in the box
 
@@ -2212,7 +2212,10 @@ class Atoms(object):
         el_name = list(el_dict.keys())
         if np.sum(np.array(el_num)==np.max(el_num)) > 1:
             warnings.warn('There are more than one majority species')
-        return int(np.max(el_num)), el_name[np.argmax(el_num)]
+        if return_count:
+            return el_name[np.argmax(el_num)], int(np.max(el_num))
+        else:
+            return el_name[np.argmax(el_num)]
 
     def extend(self, other):
         """
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -2201,15 +2201,17 @@ class Atoms(object):
 
     def get_majority_species(self):
         """
-        
-        Returns:
+        This function returns the majority species and their number in the box
+
+        Returns: number of majority atoms, chemical symbol
 
         """
         el_dict = self.get_number_species_atoms()
         el_num = list(el_dict.values())
         el_name = list(el_dict.keys())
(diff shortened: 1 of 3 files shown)


samwaseda:
I followed the order "number of atoms" then "species" but I must say it's more intuitive to have it other way around...

samwaseda:
Actually even better would be to have `return_count` as an option and per default it returns only the majority species.

pyiron/atomistics/structure/atoms.py
```diff
@@ -2201,15 +2201,17 @@ def get_equivalent_atoms(self, eps=1e-5):
...
 
         """
         el_dict = self.get_number_species_atoms()
         el_num = list(el_dict.values())
         el_name = list(el_dict.keys())
-        max_index = np.argsort(el_num)[-1]
-        return max_index, el_name[max_index]
+        if np.sum(np.array(el_num)==np.max(el_num))>1:
+            warnings.warn('There are more than one majority species')
+        return int(np.max(el_num)), el_name[np.argmax(el_num)]
```
sudarsan-surendralal: Wouldn't it be more intuitive to return the species first and the count second?

sudarsan-surendralal: Wouldn't it be more intuitive to return the species first and the count second?

tests/atomistics/structure/test_atoms.py
```diff
@@ -666,6 +670,17 @@ def test_occupy_lattice(self):
         orig_basis.occupy_lattice(H=O_indices[0])
         self.assertEqual(orig_basis.get_chemical_formula(), "Cl31HNa32")
 
+
+    def test_get_majority_species(self):
+        basis = Atoms(symbols=4*['Fe'], positions=np.random.random((4, 3)), cell=np.eye(3))
```
sudarsan-surendralal: Btw you can also initialize like

`Atoms(symbols='Fe4', positions=np.random.random((4, 3)), cell=np.eye(3))`

pyiron/atomistics/structure/atoms.py
```diff
@@ -2201,15 +2201,17 @@ def get_equivalent_atoms(self, eps=1e-5):
...
+
+        Returns: number of atoms of the majority species, chemical symbol
 
         """
         el_dict = self.get_number_species_atoms()
         el_num = list(el_dict.values())
         el_name = list(el_dict.keys())
-        max_index = np.argsort(el_num)[-1]
-        return max_index, el_name[max_index]
+        if np.sum(np.array(el_num)==np.max(el_num))>1:
```
sudarsan-surendralal: pep8 `if np.sum(np.array(el_num)==np.max(el_num)) > 1:`

pyiron/atomistics/structure/atoms.py
```diff
@@ -2201,15 +2201,17 @@ def get_equivalent_atoms(self, eps=1e-5):
 
     def get_majority_species(self):
         """
-        
-        Returns:
+        This function returns the majority species and their number in the box
+
+        Returns: number of atoms of the majority species, chemical symbol
```
sudarsan-surendralal: Enter and tab after Returns

``` 
Returns: 
       list: number of atoms of the majority species, chemical symbol
```

pyiron/atomistics/structure/atoms.py
samwaseda: I totally agree (cf. my comment above)

PR #335, opened 2019-08-08
volume added as an attribute

samwaseda:
in atomistics


samwaseda:
I just realised that `get_volume` was there, so I guess I'm gonna revoke this change (although the unittest is useful so I'll slightly change the pull request)

PR #336, opened 2019-08-08
Add collect_errors to Vasp parser

jan-janssen:


pyiron/vasp/base.py
```diff
@@ -418,8 +418,16 @@ def collect_errors(self):
...
-        self._logger.info("collect_errors() is not yet implemented for VASP")
+        file_name = os.path.join(self.working_directory, 'error.out')
+        if os.path.exists(file_name):
+            with open(file_name, 'r') as f:
+                lines = f.readlines()
+            # If the wrong convergence algorithm is chosen, we get the following error.
+            # https://cms.mpi.univie.ac.at/vasp-forum/viewtopic.php?f=4&t=17071
+            for l in lines:
+                if 'WARNING in EDDRMM: call to ZHEGV failed, returncode =' in l:
+                    self.status.not_converged = True
```
sudarsan-surendralal: But isn't `not_converged` for jobs that finished and didn't converge after the given number of steps? In this case, if I understand correctly, the job crashed right? So shouldn't you write `self.status.aborted = True` instead?

jan-janssen: The strange part is that the VASP OUTCAR looks complete, it is just the command line output which contains these error messages. But at least for my calculation it seems that these results are unusable.

PR #337, opened 2019-08-08
Set relative

samwaseda:
`scaled_positions` had to be abolished for the following reasons:

1. It is not that commonly used
2. It is not possible to keep `positions` and `scaled_positions` synchronised, since when we try to change an element, e.g. via `positions[0,0] = 0.5`, the setter is not called and therefore we have no control over `scaled_positions`.
3. We would like to keep the functionality of `positions` as a numpy array.

Therefore, please use `set_scaled_positions` and `get_scaled_positions` in the future.

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -233,16 +233,16 @@ class Atoms(object):
 
     @scaled_positions.setter
     def scaled_positions(self, positions):
-        self._scaled_positions = positions
         if self.cell is not None:
             self._positions = np.dot(self.cell.T, np.array(positions).T).T
+        self._scaled_positions = positions
 
     @positions.setter
     def positions(self, positions):
-        self._positions = positions
         if self.cell is not None:
             b_mat = np.linalg.inv(self.cell)
             self._scaled_positions = np.dot(b_mat.T, np.array(positions).T).T
+        self._positions = positions
 
     @property
     def species(self):
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -90,7 +90,8 @@ class Atoms(object):
                 cell = np.array(cell)
         self._cell = cell
         self._species = list()
-        self._internal_positions = None
+        self._positions= None
+        self._scaled_positions= None
         self._pse = PeriodicTable()
         self._tag_list = SparseArray()
         self.indices = np.array([])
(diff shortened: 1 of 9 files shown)


jan-janssen:
@samwaseda Can you also upload the notebook with the initial bug?

samwaseda:
You want to see it? Otherwise it’s already included in the unit tests

jan-janssen:
The notebook would be easier.

jan-janssen:
I just realised that ASE also does not have `scaled_positions` and uses `set_scaled_positions` and `get_scaled_positions` as well. So as long as we are compatible, that sounds fine to me.

samwaseda:
lattice = Project('.').create_structure('Al', 'fcc', 4)

basis_absolute = lattice.copy()
basis_absolute.set_absolute()
basis_relative = lattice.copy()
basis_relative.set_relative()

# basis_relative.positions[-1, 0] = 0.5
# basis_absolute.positions[-1, 0] = 0.5
basis_relative.scaled_positions[-1, 0] = 0.5
basis_absolute.scaled_positions[-1, 0] = 0.5

pyiron/atomistics/structure/atoms.py
```diff
@@ -2955,10 +2910,20 @@ def rotate_euler(self, center=(0, 0, 0), phi=0.0, theta=0.0, psi=0.0):
         rcoords = np.dot(a, np.transpose(rcoords))
         # Move back to the rotation point
         if self._is_scaled:
-            self.scaled_positions = np.transpose(rcoords) + center
+            self.get_scaled_positions(np.transpose(rcoords) + center)
```
jan-janssen: Should not this be set_scaled_positions ?

samwaseda: brilliant comment -> I'm gonna fix it

pyiron/atomistics/structure/atoms.py
```diff
@@ -1443,10 +1400,8 @@ def scaled_pos_xyz(self):
         Returns:
 
         """
-        x = self.scaled_positions[:, 0]
-        y = self.scaled_positions[:, 1]
-        z = self.scaled_positions[:, 2]
-        return x, y, z
+        xyz = self.get_scaled_positions()
+        return xyz[0], xyz[1], xyz[2]
```
jan-janssen: Should not it be `xyz[:, 0], xyz[:, 1], xyz[:, 2]`

pyiron/atomistics/structure/atoms.py
samwaseda: so this was not just a problem of this line. This function anyway doesn't seem to work. I'm looking into it more in detail

PR #339, opened 2019-08-09
Add azure script

jan-janssen:
Just playing around with azure


jan-janssen:
All tests are run with Python 3.7 but not with Python 2.7

PR #340, opened 2019-08-09
Use papermill - https://github.com/nteract/papermill

jan-janssen:
Rather than our own class for manipulating input parameters in jupyter notebooks.


samwaseda:
Can you attach a notebook or the commands for me to understand how it works?

jan-janssen:
https://papermill.readthedocs.io/en/latest/usage-parameterize.html#how-parameters-work

samwaseda:
And does this work when a notebook is submitted to the cluster?

jan-janssen:
This part still needs to be done, but it is definitely possible. https://papermill.readthedocs.io/en/latest/usage-execute.html

samwaseda:
But the main idea of `Notebook` class is to pass parameters to the notebook which is submitted. Otherwise I don't think it's particularly important to send parameters from one notebook to another (or even I find it dangerous to work this way).

PR #343, opened 2019-08-12
Set scaled positions

samwaseda:
Problem mentioned [here](https://github.com/pyiron/pyiron/pull/337#pullrequestreview-273537333). Essentially the problem came from the fact that `get_scaled_positions` by default rewrapped the coordinates, whereas the now deprecated `scaled_positions` as property did not.


jan-janssen:
The azure tests are fixed by adding `tqdm` to the requirements https://github.com/pyiron/pyiron/pull/344

PR #350, opened 2019-08-12
Append charge chgcar function

jan-janssen:


FeLoch:
If we append those new functions, is there any need for the already existing `copy_chgcar` and `copy_wavecar`?

jan-janssen:
You are right, those were included for backwards compatibility, but I guess we can remove them now.

pyiron/vasp/base.py
```diff
@@ -1065,15 +1065,37 @@ def restart_from_charge_density(self, snapshot=-1, job_name=None, job_type=None,
...
+            return new_ham
+        else:
+            return new_ham
+
+    def append_charge_density(self, job_specifier=None, path=None):
+        """
+        Append charge density file (CHGCAR)
+
+        Args:
+            job_specifier (str, int): name of the job or job ID
```
sudarsan-surendralal: `job_specifier (str/int)` is the right way

FeLoch: done

PR #355, opened 2019-08-14
restart modified for unfinished jobs to be able to rerun

samwaseda:
I tested it with [this notebook](https://github.com/pyiron/pyiron/files/3500359/restart.ipynb.txt)


jan-janssen:
This disables the `continue_final` functionality ... I am not sure if anybody is currently using it though

samwaseda:
I don't know what the consequence would be, but otherwise it doesn't work for my purpose

jan-janssen:
Maybe we can at least add a flag like `continue` which allows the users to fall back to the old behaviour.

PR #356, opened 2019-08-14
VASP: "ZBRENT: fatal error" capturing 

jan-janssen:
```
 ZBRENT: fatal error in bracketing
     please rerun with smaller EDIFF, or copy CONTCAR
     to POSCAR and continue
```


pyiron/vasp/base.py
```diff
@@ -428,6 +428,9 @@ def collect_errors(self):
                 if 'WARNING in EDDRMM: call to ZHEGV failed, returncode =' in l:
                     self.status.not_converged = True
                     break
+                elif 'ZBRENT: fatal error in bracketing' in l:
+                    self.status.not_converged = True
```
sudarsan-surendralal: Apart from setting the status to `not_converged`, I think we should also write some message in `pyiron.log` using `s.logger.warn`. This can tell the user what to do next if they encounter this error

jan-janssen: Yes, but maybe we should once more look into integrating Custodian and pymatgen because they have quite some more error handlers: 
https://materialsproject.github.io/custodian/_modules/custodian/vasp/handlers.html

FeLoch:
Reopened due to #668.

pyiron/vasp/base.py
```diff
@@ -466,27 +466,37 @@ def collect_errors(self):
         """
         Collects errors from the VASP run
         """
-        num_eddrmm, snap = self._get_eddrmm_info()
+        num_eddrmm, snap_eddrmm = self._get_eddrmm_info()
```
sudarsan-surendralal, line 469: I noticed that we are reading the entire `error.out` file in this function and also do the same thing later on for the `ZBRENT` error. Wouldn't a better way be to read the `error.out` file once and then pass the lines as the argument to this function? This way we avoid unnecessary io.

pyiron/vasp/base.py
```diff
@@ -466,27 +466,37 @@ def collect_errors(self):
...
                     ham_new.run()
                     self._logger.info("Job was restarted with 'ALGO' = 'Normal' to avoid EDDRMM warning.")
 
+        file_name = os.path.join(self.working_directory, "error.out")
+        if os.path.exists(file_name):
+            with open(file_name, "r") as f:
+                lines = f.readlines()
+                for l in lines:
+                    if "ZBRENT: fatal error in bracketing" in l:
+                        self.status.warning = True
```
sudarsan-surendralal, line 497: I think this is misleading. A warning status is assigned when the calculations have converged and some warnings are raised in between. But in your case, the calculation has actually crashed when the ZBRENT error is raised and the outputs are actually not written. So I think we should set the status to `aborted` or `not_converged`. What do you think?

pyiron/vasp/base.py
FeLoch: Sure, I wrote now everything in `collect_errors`. This should also be sufficient for future implementations.

pyiron/vasp/base.py
FeLoch: You are right. `aborted` is the only status to chose as the calculation is not complete. Remember the incomplete `vasprun.xml`.

pyiron/vasp/base.py
```diff
@@ -466,20 +466,54 @@ def collect_errors(self):
...
+            # https://cms.mpi.univie.ac.at/vasp-forum/viewtopic.php?f=4&t=17071
+            lines_where_eddrmm = np.argwhere([eddrmm_error_str in l for l in lines]).flatten()
+            num_eddrmm = len(lines_where_eddrmm)
+            if num_eddrmm > 0:
+                snap_eddrmm = len(np.argwhere(["E0=" in l for l in lines[:lines_where_eddrmm[0]]]).flatten())
+
+            # ZBRENT
+            for l in lines:
+                if zbrent_error_str in l:
+                    zbrent_status = True
```
sudarsan-surendralal, line 502: We could add a `break` statement here which means it will stop the loop once `ZBRENT` is seen.

FeLoch: Fair enough. Done

PR #357, opened 2019-08-14
Coverage test concurrency=multiprocessing option

jan-janssen:


jan-janssen:
I guess 57.706% is the more realistic number, compared to 70.522% ...

PR #361, opened 2019-08-18
Install using Environment variables 

jan-janssen:
Suggestion from @SanderBorgmans


jan-janssen:
Only the installation test seems to fail, because the directories are not created.

```
3574FAIL: test_install (base.settings.test_install.TestInstall)
3575----------------------------------------------------------------------
3576Traceback (most recent call last):
3577  File "/home/travis/build/pyiron/pyiron/tests/base/settings/test_install.py", line 31, in test_install
3578    self.assertTrue(os.path.exists(os.path.join(self.execution_path, 'project')))
3579AssertionError: False is not true
3580
3581----------------------------------------------------------------------
```

SanderBorgmans:
Could this be due to the fact the the directories are created within the init function of Settings instead of within install?

jan-janssen:
@SanderBorgmans yes, you are right. The error message is more like a reminder for myself when I look at this the next time. By the way I was thinking if it is even necessary to have a configuration file, or if we should specify the variables in the configuration file as global variables. I guess for `project_paths` and `resource_paths` this would be fine, only storing the database configuration in an environment variable seems to be a bit dangerous to me. But we anyway plan to develop a pyiron version which does not require a database.

jan-janssen:
@SanderBorgmans pyiron can now be configured using environment variables https://github.com/pyiron/pyiron/pull/394 still I am a bit reluctant to create directories and download the resource files automatically. The user should download the resources manually.

SanderBorgmans:
@jan-janssen manually fetching the resources does seem like the better option, perhaps through some user input

jan-janssen:
@SanderBorgmans That is also possible, after pyiron is installed, the user can import it: 
`import pyiron `
Then pyiron asks:
`"It appears that pyiron is not yet configured, do you want to create a default start configuration (recommended: yes). [yes/no]:"`
If the user confirms this with yes, then the resources are downloaded.

PR #364, opened 2019-08-19
Improve spin cons

FeLoch:
RWIGS is possible to set and get as a dictionary. LAMBDA and RWIGS are mandatory to set in `set_spin_constraint`.

### DIFF
--- a/pyiron/vasp/base.py
+++ b/pyiron/vasp/base.py
@@ -1198,9 +1198,7 @@ class VaspBase(GenericDFTJob):
         Sets the radii of Wigner-Seitz cell. (RWIGS tag)
 
         Args:
-            rwigs_dict (dict): Dictionary of species type and corresponding radius.
-
-        Returns:
+            rwigs_dict (dict): Dictionary of species and corresponding radii.
         """
         species_keys = self.structure.get_number_species_atoms().keys()
         rwigs_keys = rwigs_dict.keys()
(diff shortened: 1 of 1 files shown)


pyiron/vasp/base.py
```diff
@@ -1193,16 +1193,52 @@ def append_wave_function(self, job_specifier=None, path=None):
...
         """
-        Setting thr spin constraints
+        Sets the radii of Wigner-Seitz cell. (RWIGS tag)
 
         Args:
-            direction:
-            norm:
+            rwigs_dict (dict): Dictionary of species type and corresponding radius.
 
         Returns:
```
sudarsan-surendralal: remove `Returns` if not returning anything

pyiron/vasp/base.py
```diff
@@ -1193,16 +1193,52 @@ def append_wave_function(self, job_specifier=None, path=None):
...
+
+        rwigs = [rwigs_dict[i] for i in species_keys]
+        self.input.incar['RWIGS'] = ' '.join(map(str, rwigs))
+
+    def get_rwigs(self):
+        """
+        Gets the radii of Wigner-Seitz cell. (RWIGS tag)
+
+        Returns:
+            rwigs_dict (dict): dictionary of radii
```
sudarsan-surendralal: `dict: dictionary of radii` instead of `rwigs_dict (dict): dictionary of radii`. The docstrings worn't be able to render properly otherwise!

FeLoch: Thanks! Implemented.

PR #369, opened 2019-08-23
Eddrmm handling

FeLoch:
Added ways, how to deal with EDDRMM error in VASP. Three ways are possible:

- ignore: Just ignore error.
- not_converged: set status to 'not_converged' (default).
- restart: set status to 'not_converged' and restarts the job with ALGO = 'Normal' and eddrmm = 'not_converged' from initial structure.

### DIFF
--- a/pyiron/vasp/base.py
+++ b/pyiron/vasp/base.py
@@ -474,18 +474,18 @@ class VaspBase(GenericDFTJob):
                 lines = f.readlines()
             # If the wrong convergence algorithm is chosen, we get the following error.
             # https://cms.mpi.univie.ac.at/vasp-forum/viewtopic.php?f=4&t=17071
-            if not self.get_eddrmm() == "ignore":
+            if not self.get_eddrmm_handling() == "ignore":
                 for l in lines:
                     if "WARNING in EDDRMM: call to ZHEGV failed, returncode =" in l:
-                        if self.get_eddrmm() == "not_converged":
+                        if self.get_eddrmm_handling() == "not_converged":
                             self.status.not_converged = True
                             break
-                        elif self.get_eddrmm() == "restart":
+                        elif self.get_eddrmm_handling() == "restart":
                             self.status.not_converged = True
                             if not self.input.incar["ALGO"].lower() == "normal":
                                 ham_new = self.copy_hamiltonian(self.name + "_normal")
                                 ham_new.input.incar["ALGO"] = "Normal"
-                                ham_new.set_eddrmm()
+                                ham_new.set_eddrmm_handling()
                                 ham_new.run()
                             break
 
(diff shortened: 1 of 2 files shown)


pyiron/vasp/base.py
```diff
@@ -474,10 +474,34 @@ def collect_errors(self):
...
+
+    def copy_hamiltonian(self, job_name):
+        """
+        Copies a job to new one with a different name.
+
+        Args:
+            job_name (str): Job name
+
+        Returns:
+            vasp.vasp.Vasp instance: New job
```
sudarsan-surendralal: `pyiron.vasp.vasp.Vasp: New job` instead of  `vasp.vasp.Vasp instance: New job`. This will render the class path correctly in the docstrings!

pyiron/vasp/base.py
```diff
@@ -724,6 +748,25 @@ def write_magmoms(self):
         else:
             s.logger.debug("No magnetic moments")
 
+    def set_eddrmm(self, status="not_converged"):
```
sudarsan-surendralal: I wonder if it's better to rename this to `set_eddrmm_handling` and `get_eddrmm_handling`? `set_eddrmm` implies you are setting a parameter.

pyiron/vasp/base.py
```diff
@@ -724,6 +748,25 @@ def write_magmoms(self):
...
+        """
+        Sets the way, how EDDRMM warning is handled.
+
+        Args:
+            status (str): new status of EDDRMM handling (can be 'not_converged', 'ignore', or 'restart')
```
sudarsan-surendralal: `str:` instead of `status (str):`

jan-janssen: As an explanation: We are using the Google Style Python Docstrings as explained here https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

pyiron/vasp/base.py
FeLoch: agreed!  Thy type `vasp.vasp.Vasp instance` is several times written in the file (like [here](https://github.com/pyiron/pyiron/blob/1fb11e4c2c2a861a62c149cd1bb182afe4ee1027/pyiron/vasp/base.py#L1275), [here](https://github.com/pyiron/pyiron/blob/1fb11e4c2c2a861a62c149cd1bb182afe4ee1027/pyiron/vasp/base.py#L1303) or [here](https://github.com/pyiron/pyiron/blob/1fb11e4c2c2a861a62c149cd1bb182afe4ee1027/pyiron/vasp/base.py#L1373)). Maybe we should open a new branch to fix it?

pyiron/vasp/base.py
FeLoch: sounds good!

pyiron/vasp/base.py
FeLoch: yes, agreed, but since 'status' is an argument my way is correct, isn't it?

FeLoch:
changed!

PR #371, opened 2019-08-24
bugfix for new path name in rewrite_hdf5

FeLoch:
bugfix for wrong path name


FeLoch:
the last commit is made accordingly #329

PR #390, opened 2019-09-05
Added MO visualization function (not for very large structures)

jan-janssen:
@SanderBorgmans I moved the changes to the gaussian module to a separate pull request so we can discuss them in more detail.


pyiron/gaussian/gaussian.py
```diff
@@ -60,6 +63,60 @@ def from_hdf(self, hdf=None, group_name=None):
...
             self.structure = Atoms().from_hdf(hdf5_input)
 
+    def visualize_MO(self,index):
+        n_MO = self.output.scf_density.shape[0]
+        assert index > 0 and index <= n_MO
+        assert len(self.output.numbers) < 50 # check whether structure does not become too large for interactive calculation of cube file
+
+        #make cube file
+        path = self.path+'_hdf5/'+self.name+'/input'
+        out = subprocess.check_output(
```
jan-janssen: This is very specific to the Gent cluster, so we should think about a way how we can generalise this option. Maybe the cube file can be generate at the end of every job?

pyiron/gaussian/gaussian.py
```diff
@@ -145,53 +225,53 @@ def fchk2dict(fchk):
     # check job type, for now implement basics (SP=single point, FOpt = full opt, Freq = frequency calculation)
     if not fchk.command.lower() in ['sp','fopt','freq']:
         raise NotImplementedError
-    
+
```
jan-janssen: To have the generic output consistent with other codes take a look at https://pyiron.github.io/source/faq.html#which-output-quantities-are-stored-in-pyiron

pyiron/gaussian/gaussian.py
```diff
@@ -2,14 +2,16 @@
...
 
 from pyiron.base.generic.parameters import GenericParameters
 from pyiron.atomistics.structure.atoms import Atoms
-from pyiron.atomistics.job.atomistic import AtomisticGenericJob
+from pyiron.atomistics.job.atomistic import AtomisticGenericJob,GenericOutput,Trajectory
 from molmod.io.fchk import FCHKFile
-from molmod.units import amu
+from molmod.units import amu,angstrom,electronvolt
+from molmod.periodic import periodic
+import tamkin
```
jan-janssen: `tamkin` is currently not part of the conda-forge channel, so if we want to use it I recommend adding it to conda-forge as well.

SanderBorgmans: Good idea, the tamkin version that I import here is a reduced python 3 version that can be found on my github.

pyiron/gaussian/gaussian.py
SanderBorgmans: Both the MO's and the NMA functionality should be something that we have to generalise. 

For the MO part, I could not figure out how to load the complete cube file into nglview so this is what I came up with. However always creating the cube file is not very efficient, since we do not require it on a regular basis and it takes extremely long for extended systems (like MOF clusters). That is why I would prefer a post processing option for the cube file. Perhaps we could link it in some way to the gaussian executable?

pyiron/gaussian/gaussian.py
SanderBorgmans: I'll take a look at it today.

pyiron/gaussian/gaussian.py
jan-janssen: I started the submission to conda forge https://github.com/conda-forge/staged-recipes/pull/9429

pyiron/gaussian/gaussian.py
SanderBorgmans: If I understand correctly, I would change the properties that correspond to the final step to another group, e.g. structure. Should I then add the orbital and density information into a subgroup dft of the structure group?

jan-janssen:
@SanderBorgmans I looked for alternative gaussian parsers, have you tested https://github.com/cclib/cclib/blob/master/cclib/parser/gaussianparser.py ?

jan-janssen:
Replaced by https://github.com/pyiron/pyiron/pull/545

PR #394, opened 2019-09-07
Settings update 

jan-janssen:


jan-janssen:
@SanderBorgmans This allows you to configure the pyiron settings as environment variables instead of using the `.pyiron` config file.

SanderBorgmans:
@jan-janssen  Looks good! I will be on holiday until the 18th so I can't test it right away.

jan-janssen:
> @jan-janssen Looks good! I will be on holiday until the 18th so I can't test it right away.

No problem, I still plan to work on a version which can submit jobs to a cluster where pyiron is installed, but no database is available, just for the case it is not possible to install jupyterhub and a database on your cluster.

SanderBorgmans:
Sounds good! What would you recommend for our exercise lessons? Should we keep updating the pyrion version regularly from the github master branch, or rather use the version I currently have forked and await stable conda versions?

jan-janssen:
The question is what is easier for the students. On the one hand I believe an anaconda based version would allow them to install pyiron also on their own hardware and maybe play around with something when they do not have access to the cluster. On the other hand if they have access all the time, it might be easier to have a customised version for your course which gives you the flexibility to push emergency changes even without us.

jan-janssen:
I added a test for the function which uses the environment variables.

PR #395, opened 2019-09-08
Allow for singular matrices to be set as cell value

sudarsan-surendralal:
Necessary for clusters and 2D structures


samwaseda:
What is this good for?

sudarsan-surendralal:
For calculations of clusters (not supercells) with `cell` values of `[[1, 0, 0], [0, 1, 0], [0, 0, 0]]` etc without pbc on all three directions.

PR #396, opened 2019-09-09
pyiron without a database 

jan-janssen:
Based on https://github.com/pyiron/pyiron/pull/394

### DIFF
--- a/pyiron/base/database/filetable.py
+++ b/pyiron/base/database/filetable.py
@@ -1,0 +1,274 @@
+import numpy as np
+import os
+import pandas
+import datetime
+import h5io
+import sys
+from pyfileindex import PyFileIndex
+
+
+def filter_function(file_name):
+    return '.h5' in file_name
+
+
+class FileTable(object):
+    def __init__(self, project):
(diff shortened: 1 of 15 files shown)


jan-janssen:
By adding the lines: 
```
PROJECT_CHECK_ENABLED = False
DISABLE_DATABASE = True
```
To your `.pyiron` config you can disable the use of the pyiron database.

jan-janssen:
In combination with https://github.com/pyiron/pysqa/pull/20 it is now possible to execute jobs on remote clusters with do not support daemon processes.

pyiron/base/database/filetable.py
```diff
@@ -0,0 +1,365 @@
+import numpy as np
+import os
+import pandas
+import datetime
+import h5io
+import sys
+from six import with_metaclass
+from pyfileindex import PyFileIndex
```
samwaseda: URGENT: This creates ModuleNotFoundError: No module named 'pyfileindex'

jan-janssen: Should be fixed now

PR #398, opened 2019-09-11
Implement multiprocessing

jan-janssen:


jan-janssen:
Following @samwaseda recommendation to use `multiprocessing` for parallel jobs.

PR #399, opened 2019-09-11
Stress

samwaseda:
This was related to the fact that `extract_compute` (as well as `extract_atoms`) was **not** ordered according to the structure definition, i.e. it has to be reordered using `id` with `extract_atoms`.

### DIFF
--- a/pyiron/lammps/interactive.py
+++ b/pyiron/lammps/interactive.py
@@ -563,12 +563,12 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
             self._interactive_lib_command("compute st all stress/atom NULL")
             self._interactive_lib_command("run 0")
             self.interactive_cache["stress"] = []
-        ID = self._interactive_library.extract_atom("id", 0)
-        ID = np.array([ID[i] for i in range(len(self.structure))])-1
-        ID = np.arange(len(ID))[np.argsort(ID)]
+        id_lst = self._interactive_library.extract_atom("id", 0)
+        id_lst = np.array([id_lst[i] for i in range(len(self.structure))])-1
+        id_lst = np.arange(len(id_lst))[np.argsort(id_lst)]
         ind = np.array([0, 3, 4, 3, 1, 5, 4, 5, 2])
         ss = self._interactive_library.extract_compute("st", 1, 2)
-        ss = np.array([ss[i][j] for i in range(len(self.structure)) for j in range(6)]).reshape(-1, 6)[ID]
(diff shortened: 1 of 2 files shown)


pyiron/lammps/interactive.py
```diff
@@ -563,23 +563,16 @@ def interactive_stress_getter(self):
...
-                        j + (j != k) * (k + 2)
-                    ]
-                    for i in range(len(self.structure))
-                    for j in range(3)
-                    for k in range(3)
-                ]
-            ).reshape(len(self.structure), 3, 3)
-            / 1.602e6
-        )
+        ID = self._interactive_library.extract_atom("id", 0)
```
jan-janssen: `id` is a python function and writing it in capital letters is even less pythonic - can not we use something like `id_lst` or `atom_ids` ?

pyiron/lammps/interactive.py
```diff
@@ -563,23 +563,16 @@ def interactive_stress_getter(self):
...
+        ID = np.array([ID[i] for i in range(len(self.structure))])-1
+        ID = np.arange(len(ID))[np.argsort(ID)]
+        ind = np.array([0, 3, 4, 3, 1, 5, 4, 5, 2])
+        ss = self._interactive_library.extract_compute("st", 1, 2)
+        ss = np.array([ss[i][j] for i in range(len(self.structure)) for j in range(6)]).reshape(-1, 6)[ID]
+        ss = ss[:, ind].reshape(len(self.structure), 3, 3)/1.602e6
```
jan-janssen: Please do not use numbers like `1.602e6 `- but instead define the contents using `scipy.constants`

jan-janssen: That seems to be the wrong constant: 
```
from scipy import constants
constants.eV*constants.bar*constants.angstrom**3
``` 
equals `1.6021766208000003e-44`

samwaseda: `1/constants.eV*constants.bar*constants.angstrom**3`

jan-janssen: Previously you had `1/1.602e6` now it is `1/1.6021766208000003e-44` - so it seem like there is a factor missing.

samwaseda: `1/constants.eV*constants.bar*constants.angstrom**3 = 6.241509125883258e-07`

samwaseda: `1/1.602e6 = 6.242197253433208e-07`

jan-janssen: It is still not the same like `1.602e6` which you had previously.

jan-janssen: Now I get it. Looks fine to me.

PR #407, opened 2019-09-13
lammps compatible to old versions

jan-janssen:
Otherwise the Lammps 2017 crashes.


samwaseda:
What kind of error message do you get?

jan-janssen:
```
ERROR: Illegal atom_modify command (../atom.cpp:581)
Last command: atom_modify map yes
```

samwaseda:
Hm interesting. I'm okay with not having this line, but I'm very sure that we'll need it at some point.

PR #414, opened 2019-09-20
Centering coordinates only along periodic directions

sudarsan-surendralal:


sudarsan-surendralal:
> 
> 
> Shouldn't it wrap in the periodic directions? Something like `get_scaled_positions(wrap=False)[self.pbc]`?

`get_scaled_positions()` already takes care of that. It should still return a `3x3` matrix.

PR #415, opened 2019-09-20
Spx potential

jan-janssen:


jan-janssen:
@samwaseda recommended to use the VASP potentials if available and only fall back to the JTH potentials if the VASP potentials are not available. So please do not merge this pull request for now.

PR #416, opened 2019-09-25
Disable resolving symbolic links

jan-janssen:
@SanderBorgmans Can you check if this resolves your issue with symbolic links?

### DIFF
--- a/pyiron/base/settings/generic.py
+++ b/pyiron/base/settings/generic.py
@@ -581,9 +581,4 @@ def convert_path(path):
     Returns:
         str: absolute path in POSIX format
     """
-    if not (sys.version_info.major < 3 and os.name == "nt"):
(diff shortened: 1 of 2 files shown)


SanderBorgmans:
@jan-janssen Hi Jan, as I said during the skype call, this indeed solves the issue that the path on the node itself is altered. However, relative paths remain relative, is this a problem? I'll do some testing and will send you a mail with my findings soon.

jan-janssen:
@SanderBorgmans I only removed the `resolve` part, but kept the `absolut`, so to my understanding this should solve the issue.

jan-janssen:
@SanderBorgmans Does this solve your issue with the absolute paths? If yes we can merge it into the master branch.

SanderBorgmans:
@jan-janssen I think it does. But I won't be sure until Tuesday, I have another node reservation then.

PR #419, opened 2019-09-27
GenericParameters - use literal_eval rather than eval

jan-janssen:


jan-janssen:
Maybe use this one instead of https://github.com/pyiron/pyiron/pull/418

PR #426, opened 2019-10-07
Remove database check

jan-janssen:


jan-janssen:
@SanderBorgmans Did you try this one? It should help to speed up the performance issues with `job.status` and `job.status.finished `especially for SQLite databases.

SanderBorgmans:
@jan-janssen I'll update it ASAP

SanderBorgmans:
@jan-janssen It seems the job status it not refreshing at all now?

SanderBorgmans:
> @jan-janssen It seems the job status it not refreshing at all now?

Only when I reload the job object it registers whether it has finished or not

jan-janssen:
> Only when I reload the job object it registers whether it has finished or not

Yes, I guess in this way you have to call `job.status.refresh_status()` to refresh the status from the database. At the same time we reduce the number of requests send to the database. I guess it really depends on the kind of job. ... 

Just to clarify this, the reason why you check the job status is to check the status of jobs submitted to the queue, right?

jan-janssen:
The job table directly accesses the database, so the job table is always up to date.

PR #427, opened 2019-10-08
Set reservations for all child jobs in a master job

jan-janssen:


SanderBorgmans:
Looks like a nice solution! I will check whether this works for my notebook.

SanderBorgmans:
@jan-janssen it seems to be working as expected!

PR #431, opened 2019-10-10
fix meam potentials

jan-janssen:


jan-janssen:
This requires https://github.com/pyiron/pyiron-resources/pull/6

PR #437, opened 2019-10-26
Fix lammps water

sudarsan-surendralal:
Fixing some errors noticed during the simulation of ions in water

### DIFF
--- a/pyiron/base/generic/parameters.py
+++ b/pyiron/base/generic/parameters.py
@@ -1004,4 +1004,11 @@ class GenericParameters(PyironObject):
         elif len(i_line_lst) == 1:
             return i_line_lst[0], multi_word_lst[0]
         else:
-            raise ValueError("Multiple occurrences of key_name " + key_name, i_line_lst)
+            error_msg = list()
+            error_msg.append("Multiple occurrences of key_name: " + key_name + ". They are as follows")
+            for i in i_line_lst:
+                error_msg.append("dataset: {}, {}, {}".format(i,
+                                                              self._dataset["Parameter"][i],
+                                                              self._dataset["Value"][i]))
+            error_msg = "\n".join(error_msg)
+            raise ValueError(error_msg)
--- a/pyiron/base/generic/parameters.py
+++ b/pyiron/base/generic/parameters.py
@@ -1004,11 +1004,4 @@ class GenericParameters(PyironObject):
         elif len(i_line_lst) == 1:
             return i_line_lst[0], multi_word_lst[0]
         else:
-            for i in i_line_lst:
-                print(
-                    "dataset: ",
-                    i,
-                    self._dataset["Parameter"][i],
-                    self._dataset["Value"][i],
-                )
             raise ValueError("Multiple occurrences of key_name " + key_name, i_line_lst)
--- a/pyiron/lammps/structure.py
+++ b/pyiron/lammps/structure.py
@@ -416,41 +416,41 @@ class LammpsStructure(GenericParameters):
         molecule_lst, bonds_lst, angles_lst = [], [], []
 
         # Using a cutoff distance to draw the bonds (1.5 A for water) instead of the number of neighbors
-        # num_atoms_in_molecule = 3
         neighbors = self._structure.get_neighbors(cutoff=1.5)
-        # print "neighbors: ", neighbors.distances
         id_mol = 0
         indices = self._structure.indices
+        o_indices = self._structure.select_index("O")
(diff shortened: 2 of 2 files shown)


pyiron/base/generic/parameters.py
```diff
@@ -1004,11 +1004,4 @@ def _find_line(self, key_name):
         elif len(i_line_lst) == 1:
             return i_line_lst[0], multi_word_lst[0]
         else:
-            for i in i_line_lst:
```
jan-janssen: Is there a specific reason why you remove these lines? I guess they are quite helpful for debugging, so we should at least add them to the logger.

pyiron/lammps/structure.py
```diff
@@ -406,58 +406,55 @@ def structure_full(self):
...
-        molecule_lst, bonds_lst, angles_lst = [], [], []
+        species_translate_list = [self.potential.get("group {} type".format(el.Abbreviation))
+                                  for el in self.structure.species]
 
-        # species_lst = structure.get_species_objects()
-        # for id_el, el in enumerate(structure.species):
-        #     el.id = id
-        # el_lst = structure.get_chemical_elements()
+        # Drawing bonds only for water molecules
+        molecule_lst, bonds_lst, angles_lst = [], [], []
```
jan-janssen: If it is specific for water, we should give an error message if any other molecules are set. I guess we had this issue with Markus before and maybe Marcel has had a similar experience. The important part is to make it clear what is supported and what is not supported.

pyiron/lammps/structure.py
```diff
@@ -489,10 +487,11 @@ def structure_full(self):
             id_atom, id_mol, id_species = atom
             # print id_atom, id_mol, id_species
             x, y, z = coords[id_atom]
-            el_id = self._structure.species[id_species].Abbreviation
+            ind = np.argwhere(np.array(species_translate_list) == id_species).flatten()[0]
```
jan-janssen: Same here, emphasise that only water molecules are supported.

pyiron/lammps/structure.py
sudarsan-surendralal: Not sure how to do this since we can't identify if an atom belongs to a molecule or not. This info is not available from our atoms class. For example, an atom may be an ion in solution and not a molecule. So if we raise an error message, it might be highly misleading. If more molecule types are to be supported, then they really have to be implemented on a case by case basis, unfortunately. Btw Marcels' system will work now with this fix.

pyiron/lammps/structure.py
sudarsan-surendralal: This works for all species in the potential config. Not just water molecules

pyiron/base/generic/parameters.py
sudarsan-surendralal: Fixed here https://github.com/pyiron/pyiron/pull/437/commits/7a860d969037f8d2537967d3d75245e09ba4baa0. Is this what you mean?

pyiron/lammps/structure.py
jan-janssen: I am still confused because you explicitly select O and H indices, so should not we raise error messages for other molecules?

pyiron/lammps/structure.py
```diff
@@ -406,58 +406,55 @@ def structure_full(self):
...
-        num_atoms_in_molecule = 3
-        neighbors = self._structure.get_neighbors(
-            num_neighbors=num_atoms_in_molecule + 2
-        )
-        # print "neighbors: ", neighbors.distances
+        # Using a cutoff distance to draw the bonds instead of the number of neighbors
+        neighbors = self._structure.get_neighbors(cutoff=2)
         id_mol = 0
         indices = self._structure.indices
+        o_indices = self._structure.select_index("O")
```
jan-janssen: This part seems to be explicit for water.

pyiron/lammps/structure.py
sudarsan-surendralal: Yeah, but that is just to estimate if they are any O-H bonds in the structure. So if there are any O atoms in the structure, the function will check if any H atoms are there in its vicinity. If so, it will draw bonds. The atoms will be written even in the absence of any bonds.

jan-janssen:
@sudarsan-surendralal do we have any tests for the water bonds and maybe some other kind of bond, like the glass bonds?

sudarsan-surendralal:
Unfortunately no. We would have to check with the group leaders to see if we can get access to some notebooks!

PR #440, opened 2019-10-30
Lammps bonds

sudarsan-surendralal:
Defining a generalized way of drawing bonds in LAMMPS through pyiron

### DIFF
--- a/pyiron/lammps/base.py
+++ b/pyiron/lammps/base.py
@@ -947,7 +947,6 @@ class Input:
         with hdf5.open("input") as hdf5_input:
             self.control.to_hdf(hdf5_input)
             self.potential.to_hdf(hdf5_input)
-            hdf5_input["bond_dict"] = self.bond_dict
 
     def from_hdf(self, hdf5):
         """
@@ -961,7 +960,8 @@ class Input:
         with hdf5.open("input") as hdf5_input:
             self.control.from_hdf(hdf5_input)
             self.potential.from_hdf(hdf5_input)
-            self.bond_dict = hdf5_input["bond_dict"]
+            if "bond_dict" in hdf5_input.list_nodes():
+                self.bond_dict = hdf5_input["bond_dict"]
 
 
 def to_amat(l_list):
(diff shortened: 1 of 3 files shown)


pyiron/lammps/base.py
```diff
@@ -920,6 +968,13 @@ class Input:
     def __init__(self):
         self.control = LammpsControl()
         self.potential = LammpsPotential()
+        self.bond_dict = dict()
```
jan-janssen: Should not the bond dict be empty by default?

sudarsan-surendralal: I thought about this a little bit. I think it is important to have some pre-defined default values for `bond_dict` just as an example. Once again, these bonds will only be drawn if water molecules are present (not just O or H atoms). The users can modify/play with these parameters if they want to. I don't think it is a good idea to force the user to define these bond parameters every time water is present since a certain bit of convenience is necessary.

jan-janssen: We can have an example as comment, or have a separate function the user has to call, but we should not have any bonds by default, because then the behavior for water is different compared to other molecules.

sudarsan-surendralal: But that would not be ideal. If the bonds are not drawn by default, the system would simply run the calculation without any error messages (since technically it is still a valid potential) which is highly misleading. This is dangerous since many people already use the water example on our website. So for the sake of backward compatibility, I think we should allow just these default bond parameters at least for water (most of the simulations using this style use water anyway). I would rather that users who do not want water bonds to be drawn (who are very rare) to use the `define_bonds()` function explicitly. Besides `define_bonds()` has very LAMMPS specific parameters (eg. bond/angle type) which I think only users who know what they are doing should modify.

jan-janssen: It all comes down to the zen of python `Special cases aren't special enough to break the rules.` You could raise a warning that no bonds were defined.

sudarsan-surendralal: I disagree. I don't think this violates the zen of python.

sudarsan-surendralal:
@jan-janssen can we merge this? Of course, we can extend this anytime. But for now, I think it is important to get these changes in the master branch since Jörg needs them for his calculations.

PR #441, opened 2019-11-06
pyscal to analyse steinhardt parameter

jan-janssen:


jan-janssen:
@liamhuber I added the option to provide q as input parameter, even though the current implementation is mainly focused on identifying solid and liquid phases, I definitely see the option to extend it for machine learning.

PR #450, opened 2019-11-21
Vcsgc adaptations

liamhuber:
- Implements the special Lammps fix for variance-constrained semi-grand canonical Monte Carlo/MD as a `calc` for Lammps jobs.
  - This is only for non-interactive jobs; I don't think this fix is available in the interactive library
- Under the hood we now also read the indices alongside xyz coordinates from the Lammps output
  - If new indices are encountered, new species are added to `self.structure` accordingly so that they can be used

This is my first crack at writing tests, so I'd appreciate if someone could pay some special criticism to the `_test` file changes!


jan-janssen:
I see no reason why this should not work with interactive jobs - the lammps version is the same - still I agree it makes more sense to use it in non-interactive mode.

liamhuber:
@jan-janssen do I need to do anything special to get the tests to run again? The failures are now resolved when I run the lammps `test_base.py` checks on Garching.

liamhuber:
Regarding the interactive capability, I don't know what to say. If the interactive library itself will have no trouble binding to the new fix then it ought to work. It doesn't though.

Annoyingly it doesn't fail and give a useful error message, it just doesn't do anything (except what appears to be a `calc_static`).

Does the interactive master (@samwaseda) have any insight?

MWE:
```python
from pyiron.project import Project
pr = Project('tmp')

job = pr.create_job(pr.job_type.Lammps, 'interactive_nonstarter')
job.structure = pr.create_ase_bulk('Al', orthorhombic=True).repeat(15)
job.structure[0] = 'Mg'
job.potential = job.list_potentials()[0]
job.calc_vcsgc(
    mu = {'Al': 0, 'Mg': -5},
    target_concentration = {'Al': 0.5, 'Mg': 0.5},
    mc_step_interval = 1,
    temperature = 300,
    n_ionic_steps = 10,
    n_print=1
)

interactive=False

if interactive:
    job.interactive_open()
    job.run()
    job.interactive_close()
else:
    job.run()

print([np.mean(job.get_structure(i).indices) for i in np.arange(len(job.output.positions))])
```

Output for regular:

`[0.00014814814814814815, 0.09348148148148148, 0.18, 0.25437037037037036, 0.3248888888888889, 0.389037037037037, 0.444, 0.49274074074074076, 0.5013333333333333, 0.5017777777777778, 0.5017777777777778]`

Output for interactive:

`[0.00014814814814814815]`

samwaseda:
(NB: I'm making this comment without having taken a look at the code)

Static `calc_md`: pyiron parses the output files every `n_print` steps. E.g. if `n_ionic_steps=10` and `n_print=1`, then it gives 10 output values
Interactive calc_md`: pyiron gives only the last step back (essentially it considers one `n_ionic_steps` MD step as one interactive run/step)

samwaseda:
Probably you have to change this function (i.e. you have to add `if self._generic_input["calc_mode"] == "vcsgc":`)

https://github.com/pyiron/pyiron/blob/16051780e8baa14cf6b42bc889f6c1fd092a0a27/pyiron/lammps/interactive.py#L324

liamhuber:
@jan-janssen The tests are now running, but I expect one failure:

```
4205ERROR: test_vcsgc_integration (lammps.test_base.TestLammps)
4206----------------------------------------------------------------------
4207Traceback (most recent call last):
4208  File "/home/travis/build/pyiron/pyiron/tests/lammps/test_base.py", line 470, in test_vcsgc_integration
4209    self.job_vcsgc_integration.run()
4210  File "/home/travis/build/pyiron/pyiron/pyiron/base/job/generic.py", line 660, in run
4211    self._run_if_new(debug=debug)
4212  File "/home/travis/build/pyiron/pyiron/pyiron/base/job/generic.py", line 1303, in _run_if_new
4213    self._create_job_structure(debug=debug)
4214  File "/home/travis/build/pyiron/pyiron/pyiron/base/job/generic.py", line 1567, in _create_job_structure
4215    self.write_input()
4216  File "/home/travis/build/pyiron/pyiron/pyiron/lammps/base.py", line 324, in write_input
4217    self.input.potential.copy_pot_files(self.working_directory)
4218  File "/home/travis/build/pyiron/pyiron/pyiron/lammps/potential.py", line 93, in copy_pot_files
4219    if self.files is not None:
4220  File "/home/travis/build/pyiron/pyiron/pyiron/lammps/potential.py", line 88, in files
4221    raise ValueError("Was not able to locate the potentials.")
4222ValueError: Was not able to locate the potentials.
```

I guess this is because `self.job_vcsgc_input.potential = self.job_vcsgc_input.list_potentials()[0]` is finding zero available potentials for the AlMg system I'm testing on. Please advise.

jan-janssen:
To me it seems like you are trying to run a lammps calculation as part of the test. This is not recommended as it is going to take a lot of resources. Instead we recommend adding a notebook example - those are executed using https://github.com/pyiron/pyiron-release-testing . For the unit tests we commonly just write the input_files and check that they are consistent with what we expect. As you did not change anything related to the read and write procedures I would believe it is enough to compare if the right flags are written to the lammps generic parameters.

liamhuber:
@samwaseda thanks for the pointer; I modified the clause in `run_if_interactive` to `if self._generic_input["calc_mode"] in ["md", "vcsgc"]:` since VCSGC is basically just glorified MD anyhow, and added a new `interactive_indices_getter` for interactive Lammps. Works like a charm now.

pyiron/atomistics/job/atomistic.py
```diff
@@ -473,6 +473,10 @@ def trajectory(
 
         """
         cells = self.output.cells
+        if len(self.output.indices) != 0:
+            indices = self.output.indices
+        else:
+            indices = self.get("output/generic/indices")
```
jan-janssen: Does this work with old calculation? once which were calculated before this update?

pyiron/atomistics/job/atomistic.py
```diff
@@ -654,6 +676,8 @@ def get_structure(self, iteration_step=-1, wrap_atoms=True):
             snapshot.cell = self.output.cells[iteration_step]
         snapshot.positions = self.output.positions[iteration_step]
         indices = self.output.indices
+        if len(indices) == 0:
+            indices = self['output/generic/indices']
```
jan-janssen: I am a bit confused about this construction, would not it be easier to update `self.output.indices` to fall back to `self['output/generic/indices']` if it exists?

pyiron/lammps/interactive.py
```diff
@@ -539,7 +519,29 @@ def update_potential(self):
         self._interactive_lib_command(self.potential.Config[0][1])
 
     def interactive_indices_getter(self):
-        return super(LammpsInteractive, self).interactive_indices_getter().tolist()
+        lammps_indices = np.array(self._interactive_library.gather_atoms("type", 0, 1))
+        indices = self.remap_indices(lammps_indices)
+        return indices.tolist()
+
+    def interactive_indices_setter(self, indices):
```
jan-janssen: Just for my understanding, this function was only moved, but nothing changes, right?

pyiron/lammps/interactive.py
```diff
@@ -539,7 +519,29 @@ def update_potential(self):
         self._interactive_lib_command(self.potential.Config[0][1])
 
     def interactive_indices_getter(self):
-        return super(LammpsInteractive, self).interactive_indices_getter().tolist()
```
jan-janssen: Can you test if these changes have an impact on the performance of the interactive mode? If that is the case it might make sense to check the calculation mode in the generic_input: `self._generic_input["calc_mode"] in ["vcsgc"]`

liamhuber: I did two MD tests locally on my laptop, both 6750 Al atoms at room temperature.

| n_steps | n_print | updating | static |
|---------|---------|----------|--------|
| 1000    | 10      | 2m44s    | 2m35s  |
| 100     | 1       | 2m20s    | 2m22s  |

I'm comfortable leaving it 'always on'. 

Non-interactively this takes 17s, so interactive is still super slow, but with some thousands of atoms this doesn't seem to make a big difference.

pyiron/lammps/interactive.py
liamhuber: Yes, just to make the getter and setter code adjacent; they were previously quite far apart.

pyiron/atomistics/job/atomistic.py
liamhuber: Agreed. I had previously been unable to figure out how to do that, but it was in the `pyiron/job/interactive.GenericeInteractiveOutput.indices` property, which needed `prop=GenericOutput.indices.fget` as an argument to `_lst_from_property`.

pyiron/atomistics/job/atomistic.py
liamhuber: Good catch. It does now; if the indices aren't there I just cover by using the structure's indices at each frame.

PR #451, opened 2019-11-22
Add an easy function for checking the git hash of each pyiron module

liamhuber:
As discussed with @jan-janssen [on the private gitlab board](https://gitlab.mpcdf.mpg.de/pyiron/pyiron-support/issues/229)


liamhuber:
I don't see what about python 2.7 was failing...

jan-janssen:
It looks like Python 2.7 is not able to resolve the dependencies.

PR #455, opened 2019-11-24
Do not recreate restart job if it exists

sudarsan-surendralal:


pyiron/base/job/generic.py
```diff
@@ -1198,7 +1198,7 @@ def restart(self, snapshot=-1, job_name=None, job_type=None):
             job_name = "{}_restart".format(self.job_name)
         if job_type is None:
             job_type = self.__name__
-        if job_type == self.__name__:
+        if job_type == self.__name__ and job_name not in self.project.job_table().job.tolist():
```
jan-janssen: This fails if the job_table is empty. Maybe we should have a general function on the project level to check if a job name exists already.

jan-janssen: There is already the project list nodes function, but I guess also that one fails when the job table is empty:
https://github.com/pyiron/pyiron/blob/master/pyiron/base/project/generic.py#L701

PR #459, opened 2019-11-25
Static methods need no self argument

liamhuber:
derp


jan-janssen:
I see the urgency of this change, but please ask at least one person to review your changes before merging with the master branch.

PR #462, opened 2019-11-27
Remove EnCUT

jan-janssen:
Default: ENCUT	= largest ENMAX on the POTCAR file
https://www.vasp.at/wiki/index.php/ENCUT

### DIFF
--- a/tests/vasp/test_vasp.py
+++ b/tests/vasp/test_vasp.py
@@ -67,7 +67,7 @@ class TestVasp(unittest.TestCase):
         self.assertEqual(self.job.potential, self.job._potential)
 
     def test_plane_wave_cutoff(self):
-        self.assertIsInstance(self.job.plane_wave_cutoff, (float, int))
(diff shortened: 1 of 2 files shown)


tests/vasp/test_vasp.py
```diff
@@ -67,7 +67,7 @@ def test_potential(self):
         self.assertEqual(self.job.potential, self.job._potential)
 
     def test_plane_wave_cutoff(self):
-        self.assertIsInstance(self.job.plane_wave_cutoff, (float, int))
+        # self.assertIsInstance(self.job.plane_wave_cutoff, (float, int))
```
sudarsan-surendralal: I think we should keep this line. Just use

`self.assertIsInstance(self.job.plane_wave_cutoff, (float, int, type(None))`

jan-janssen: My assumption would be that the tests fail with those changes, but let's give it a try.

jan-janssen: I have to correct myself - it worked, just missing a `)`

PR #463, opened 2019-11-29
Lammps triclinic repairs and extension

liamhuber:
Key features:

- Convert to and from triclinic cells for Lammps
- Allow non-diagonal pressures (i.e. Pxy, Pxz, and Pyz) during Lammps relaxation and minimization
- Fun extra: synchronized `calc_minimize` docstrings to a single place then linked it in other places (plus tests to make sure the defaults stay validly described)

I had originally intended just to get triclinic cells working, e.g. for running primitive cells (there was a conversion in place but it had a bug), but wound up implementing a more complete binding to cell shape changes.

### DIFF
--- a/pyiron/lammps/base.py
+++ b/pyiron/lammps/base.py
@@ -555,20 +555,14 @@ class LammpsBase(AtomisticGenericJob):
                 hdf_output[k] = np.array(v)
 
     def calc_minimize(
-        self, e_tol=0.0, f_tol=1e-2, max_iter=100000, pressure=None, n_print=100
+            self,
+            e_tol=0.0,
+            f_tol=1e-4,
+            max_iter=100000,
+            pressure=None,
+            n_print=100
     ):
-        """
-
-        Args:
-            e_tol:
-            f_tol:
-            max_iter:
-            pressure:
-            n_print:
-
-        Returns:
-
-        """
+        # Docstring set programmatically -- Ensure that changes to signature or defaults stay consistent!
         super(LammpsBase, self).calc_minimize(
             e_tol=e_tol,
             f_tol=f_tol,
@@ -583,6 +577,7 @@ class LammpsBase(AtomisticGenericJob):
             pressure=pressure,
             n_print=n_print,
         )
+    calc_minimize.__doc__ = LammpsControl.calc_minimize.__doc__
 
     def calc_static(self):
         """
--- a/pyiron/lammps/base.py
+++ b/pyiron/lammps/base.py
@@ -559,7 +559,8 @@ class LammpsBase(AtomisticGenericJob):
             f_tol=1e-4,
             max_iter=100000,
             pressure=None,
-            n_print=100
+            n_print=100,
+            style='cg'
     ):
         # Docstring set programmatically -- Ensure that changes to signature or defaults stay consistent!
         super(LammpsBase, self).calc_minimize(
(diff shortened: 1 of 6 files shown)


pyiron/lammps/control.py
```diff
@@ -101,23 +101,60 @@ def load_default(self, file_content=None):
...
+            style ('cg'/'sd'/other values from Lammps docs): The style of the numeric minimization, either conjugate
+                gradient, steepest descent, or other keys permissible from the Lammps docs on 'min_style'. (Default
+                is 'cg' -- conjugate gradient.)
+        """
+        # This docstring is a source for the calc_minimize method in pyiron.lammps.base.LammpsBase.calc_minimize and
+        # pyiron.lammps.interactive.LammpsInteractive.calc_minimize -- Please ensure that changes to signature or
+        # defaults stay consistent!
+        max_evaluations = 100 * max_iter
+        GPA_TO_BAR = spc.giga * 1.0 / spc.bar
+        if self["units"] != "metal":
```
sudarsan-surendralal: Is there a reason why you restrict `calc_minimize()` only for calculations with `metal` units? Currently, pyiron users also use LAMMPS with `real` units (water calculations for example). So maybe a better way to do this is to define something like

```
pressure_conv = {"metal":  spc.giga * 1.0 / spc.bar, "real":  spc.giga * 1.0 / spc.atm} 
if self["units"] not in ["metal", "real"]:
    raise NotImplementedError("`calc_md()` does not support the {} LAMMPS unit yet!".format(self["units"]))
``` 

And then later on

"""
str_press += str_axis + str(press * pressure_conv[self["units"]])
"""

pyiron/lammps/control.py
```diff
@@ -101,23 +101,60 @@ def load_default(self, file_content=None):
...
+
         if pressure is not None:
             if type(pressure) == float or type(pressure) == int:
                 pressure = pressure * np.ones(3)
             str_press = ""
-            for press, str_axis in zip(pressure, [" x ", " y ", " z "]):
+            for press, str_axis in zip(pressure, [" x ", " y ", " z ", " xy ", " xz ", " yz "][:len(pressure)]):
                 if press is not None:
-                    str_press += str_axis + str(press * 1.0e4)
+                    str_press += str_axis + str(press * GPA_TO_BAR)
```
sudarsan-surendralal: See earlier comment

pyiron/lammps/base.py
```diff
@@ -972,10 +984,19 @@ def _get_lammps_structure(self, structure=None, cutoff_radius=None):
         else:
             lmp_structure.cutoff_radius = self.cutoff_radius
         lmp_structure.el_eam_lst = self.input.potential.get_element_lst()
+
+        def structure_to_lammps(structure):
+            """Converts a structure to the Lammps coordinate frame"""
```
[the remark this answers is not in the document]
liamhuber: Fixed

pyiron/lammps/control.py
liamhuber: The reason was because I was lazy and didn't want to look too closely at the units :p ...Now I did [look more closely](https://github.com/pyiron/pyiron/issues/465) ;)

This restriction is lifted here though, and the conversion dictionaries are typed out for using in a bit more thorough treatment of units elsewhere in the future.

PR #464, opened 2019-12-02
Fix vasprun

sudarsan-surendralal:
A minor fix in order to parse element types


sudarsan-surendralal:
@janj shall I merge this?

PR #469, opened 2019-12-03
Enmax finder

liamhuber:
The [VASP wiki](https://www.vasp.at/wiki/index.php/ENCUT) states 

> For consistency reasons we still recommend to specify the cutoff manually in the INCAR file and keep it constant throughout a set of calculations.

e.g. if you're looking at the segregation of various solutes to a particular defect, then you should use the highest ENCUT among host and all solutes across all calculations, even that of your pure-host defect. 

Thus, it's nice to be able to quickly assess what that cutoff value is for a given set of elements.

### DIFF
--- a/pyiron/vasp/potential.py
+++ b/pyiron/vasp/potential.py
@@ -6,6 +6,7 @@ import os
 import pandas
 from pyiron.base.settings.generic import Settings
 from pyiron.atomistics.job.potentials import PotentialAbstract
+from pyiron.vasp.base import Potcar
 
 __author__ = "Jan Janssen"
 __copyright__ = (
@@ -272,3 +273,38 @@ def find_potential_file(file_name=None, xc=None, path=None, pot_path_dict=None):
                     if file_name in file_lst:
                         return os.path.join(path, file_name)
     raise ValueError("Either the filename or the functional has to be defined.")
+
+
+def get_enmax_among_species(species_lst, return_list=False, xc="PBE"):
+    """
+    Given a list of species symbols, finds the largest applicable encut.
+
+    Args:
+        species_lst (list): The list of species symbols.
+        return_list (bool): Whether to return the list of all ENMAX values (in the same order as `species_lst` along with
+            the largest value). (Default is False.)
+        xc ("GGA"/"PBE"/"LDA"): The exchange correlation functional for which the POTCARs were generated. (Default is "PBE".)
+
+    Returns:
+        (float): The largest ENMAX among the POTCAR files for all the species.
+        [optional](list): The ENMAX value corresponding to each species.
+    """
+    enmax_lst = []
+    vpf = VaspPotentialFile(xc=xc)
+
+    for symbol in species_lst:
+        potcar_file = find_potential_file(
+            path=vpf.find_default(symbol)['Filename'].values[0][0],
+            pot_path_dict=Potcar.pot_path_dict
+        )
+        with open(potcar_file) as pf:
+            for i, line in enumerate(pf):
+                if i == 14:
+                    encut_str = line.split()[2][:-1]
+                    enmax_lst.append(float(encut_str))
+                    break
+
+    if return_list:
+        return max(enmax_lst), enmax_lst
+    else:
+        return max(enmax_lst)
--- a/pyiron/vasp/potential.py
+++ b/pyiron/vasp/potential.py
@@ -274,12 +274,12 @@ def find_potential_file(file_name=None, xc=None, path=None, pot_path_dict=None):
     raise ValueError("Either the filename or the functional has to be defined.")
 
 
-def get_enmax_among_species(species_lst, return_list=False, xc="PBE"):
+def get_enmax_among_species(symbol_lst, return_list=False, xc="PBE"):
     """
     Given a list of species symbols, finds the largest applicable encut.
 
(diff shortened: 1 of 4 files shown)


liamhuber:
Tests are passing on my machine and Garching. I think the failure has to do with the POTCARs (un)availability on the remote testing server, but I'm not sure how to follow up.

The error:
```
ERROR: test_get_enmax_among_species (vasp.test_potential.TestPotential)
  File "/home/vsts/work/1/s/tests/vasp/test_potential.py", line 17, in test_get_enmax_among_species
    float_out = get_enmax_among_species(['Al'], return_list=False)
  File "pyiron/vasp/potential.py", line 300, in get_enmax_among_species
    pot_path_dict=pot_path_dict
  File "pyiron/vasp/potential.py", line 274, in find_potential_file
    raise ValueError("Either the filename or the functional has to be defined.")
ValueError: Either the filename or the functional has to be defined.
```

How I understand the cause: 

In
```python
vpf = VaspPotentialFile(xc=xc)

for symbol in symbol_lst:
    potcar_file = find_potential_file(
        path=vpf.find_default(symbol)['Filename'].values[0][0],
        pot_path_dict=pot_path_dict
    )
```

`vpf.find_default(symbol)['Filename'].values[0][0]` is returning `None`?

jan-janssen:
We are not allowed to publish the VASP potentials, all we can publish are reduced templates. Maybe you can modify the template for iron and then use it in your example https://github.com/pyiron/pyiron/blob/master/tests/static/vasp/potentials/potpaw_PBE/Fe/POTCAR

liamhuber:
Fair enough.
Thanks for the pointer! I think I see how to modify it and will take a crack at it in the morning.

pyiron/vasp/potential.py
```diff
@@ -272,3 +272,41 @@ def find_potential_file(file_name=None, xc=None, path=None, pot_path_dict=None):
...
+        symbol_lst (list): The list of species symbols.
+        return_list (bool): Whether to return the list of all ENMAX values (in the same order as `species_lst` along with
+            the largest value). (Default is False.)
+        xc ("GGA"/"PBE"/"LDA"): The exchange correlation functional for which the POTCARs were generated. (Default is "PBE".)
+
+    Returns:
+        (float): The largest ENMAX among the POTCAR files for all the species.
+        [optional](list): The ENMAX value corresponding to each species.
+    """
+    from pyiron.vasp.base import Potcar
```
jan-janssen: We should do the import at the top of the file - that is recommended by PEP8. The only point where we import in the function is when it is a dependency which might not be available on all systems, like `ovito`.

liamhuber: No can do (trace below). This introduces a cyclic dependency between potential.py and base.py. 

I think the real solution is to change the architecture. There are a few choices:
- `get_enmax_among_species` could live directly in in base.py. I don't like this -- potential stuff should be in the potential module.
- `get_enmax_among_species` could not be a method of the vasp job type. I don't like this -- it needs to be tab-completable from somewhere the user will have in her notebook.
- `pot_path_dict` could be moved from belonging to the `Potcar` class to somewhere else, e.g. inside the potential module -- this makes a lot of sense to me.
- The `Potcar` class could be moved into the potential module -- this makes some sense to me, but then `Incar`, `Kpoints`, etc would need to be moved out of base for consistency.

I'm perfectly happy to bend a little Pep8 to delay resolving this issue.

jan-janssen: @sudarsan-surendralal Can we move the `POTCAR` class to a separate file? That should resolve this issue. https://github.com/pyiron/pyiron/blob/enmax_finder/pyiron/vasp/base.py#L2184

liamhuber: @jan-janssen @sudarsan-surendralal if we're going to move it, is there a reason not to simply move it into potential.py?

jan-janssen: I am also fine with that.

sudarsan-surendralal: Yes moving it to `potential.py` is a good idea

liamhuber: Ok, I refactored it over. All the vasp tests run fine on Garching; we'll see how it turns out on remote.

liamhuber: If all the tests pass, I guess we're all satisfied and can merge it in.

pyiron/vasp/potential.py
```diff
@@ -272,3 +277,187 @@ def find_potential_file(file_name=None, xc=None, path=None, pot_path_dict=None):
...
+
+def get_enmax_among_species(symbol_lst, return_list=False, xc="PBE"):
+    """
+    Given a list of species symbols, finds the largest applicable encut.
+
+    Args:
+        symbol_lst (list): The list of species symbols.
+        return_list (bool): Whether to return the list of all ENMAX values (in the same order as `species_lst` along with
+            the largest value). (Default is False.)
+        xc ("GGA"/"PBE"/"LDA"): The exchange correlation functional for which the POTCARs were generated. (Default is "PBE".)
```
sudarsan-surendralal: `xc (str): The exchange correlation functional for which the POTCARs were generated can have values of "GGA"/"PBE"/"LDA". (Default is "PBE".)`

PR #470, opened 2019-12-04
sphinx restart 

samwaseda:
SPHInX can now be relaunched when a simulation does not converge within the allocated computation time (e.g. 4 days). It had already been possible to create a restart job by `job.restart()`, but it did not reflect the final results automatically (i.e. the output files of the not converged simulation were not parsed). With this fix, a single  `job.restart()` will parse the output files of the not-converged simulation **and** it sets the status to `not_converged` (instead of `aborted` which had been the case before).

### DIFF
--- a/pyiron/sphinx/base.py
+++ b/pyiron/sphinx/base.py
@@ -336,6 +336,12 @@ class SphinxBase(GenericDFTJob):
         from_charge_density=True,
         from_wave_functions=True,
     ):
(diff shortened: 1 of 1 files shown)


samwaseda:
Here's an example notebook

[restart_sphinx.ipynb.txt](https://github.com/pyiron/pyiron/files/3920929/restart_sphinx.ipynb.txt)

PR #471, opened 2019-12-04
fix openkim interface

jan-janssen:


jan-janssen:
Required for https://github.com/pyiron/pyiron-resources/pull/7

samwaseda:
Doesn't it make more sense to do something like this?

```
if len(self._df["Filename"])>0 and list(self._df["Filename"])[0] != ['']
```

At least I prefer `is not None` to putting nothing...

samwaseda:
Sorry I guess it should have been `len(self._df["Filename"][0])>0`

PR #472, opened 2019-12-05
sxphonon fixes

jan-janssen:


samwaseda:
I can't really review this one - I don't know what it does.

jan-janssen:
I copied functions from the Sphinx code, as it is using the Sphinx output parser. Copying code is definitely not the best way to do it, but it was the quickest way to fix the Sphinx output parser for sxphonons.

PR #476, opened 2019-12-06
avoid named pipes (because otherwise decompression fails)

samwaseda:


jan-janssen:
Is not this already included in https://github.com/pyiron/pyiron/pull/475 ?

samwaseda:
Short answer: yes.

What I wanted to do:
- in SPHInX: pipes are deleted
- in general: pipes are not deleted by not compressed

But as it turned out in both cases pipes are deleted in this implementation. Then I thought it might actually be better to simply delete all pipes whenever `compress` is called. So I guess we can keep this one and I remove the pipe deletion in [#475](https://github.com/pyiron/pyiron/pull/475)

jan-janssen:
@samwaseda Should we merge this or close the issue? As far as I understand it is already included.

samwaseda:
I guess we should rather have this one - named pipes make compression fall flat in any case, so it makes sense to prevent it beforehand for any code.

PR #481, opened 2019-12-12
always remove nve

jan-janssen:


jan-janssen:
You are right, I just have the feeling it is confusing for LAMMPS users to have commands in there which are not used. So maybe it is more clear if we handle the special cases like fire separately.

liamhuber:
No objections from my end. I have a mild preference to add functionality for fire-like things back in later if/when needed.

jan-janssen:
@samwaseda Can you comment on this, or can I merge it?

PR #482, opened 2019-12-13
Sphinx interactive energy type

liamhuber:
Let's return a float instead of np.array(float).


samwaseda:
Why array? It returns one value for each `get energy`

liamhuber:
Yes. This is exactly my problem. Instead of returning the one value as a float, it returns it as a float wrapped in an array. Will it break the code elsewhere to get rid of this array wrapping? Was there some rational for wrapping it to start with?

samwaseda:
But this one expects one value

https://github.com/pyiron/pyiron/blob/72e4c16a4c9d79344e1011a90e4cebcbe302ad44/pyiron/atomistics/job/interactive.py#L182

samwaseda:
Oh sorry I thought you were doing exactly the opposite

liamhuber:
I don't understand if we disagree about anything or not.

Note:
```python

foo = np.array(1)
bar = np.array(2)
baz = []
for x in [foo, bar]:
    baz.append(x)
print(baz)
print(np.array(baz))

> [array(1), array(2)]
> [1 2]
```

liamhuber:
Ahhh, good! No, I am doing the opposite of the opposite ;)

liamhuber:
I think my example a couple comments above probably shows how we got away without noticing this for so long though.

samwaseda:
I guess I had never used this one as such, so you're probably right

liamhuber:
Cool. I'll let the tests run, but provided they come back green I'll merge.

PR #484, opened 2019-12-21
number of steps in calc_minimize changed

samwaseda:
I increased `max_iter` because when a particularly large box is to be minimised, 1e5 was not always enough. Besides I don't see any particular reason why it should not be very large.


jan-janssen:
When the calculation does not finish within the number of steps, its status is set to not-converged right? So shouldn’t we try for calculations to fail as quickly as possible, in particular for large systems more steps might mean it takes forever and the user can always take the final structure and minimize from there. It is just like the electronic steps in DFT, I believe the number should be sufficiently small, so if electronic convergence is an issue, people have to set this parameter manually, which gives other users of the same protocol a hint that the electronic convergence is a parameter which might require more specific tuning when switching the materials system. Ideally we would have an algorithm to set these convergence parameter depending on the system but until we have such an algorithm, I am in favor of failing quickly, rather than wasting computing resources. 

But that is just my opinion, @liamhuber and @sudarsan-surendralal what is your take on this?

jan-janssen:
@samwaseda by the way I noticed that the azure tests started failing recently dude to some issue with Sphinx, maybe you can take a look at the logs and fix it in a separate pull request?

samwaseda:
> When the calculation does not finish within the number of steps, its status is set to not-converged right? So shouldn’t we try for calculations to fail as quickly as possible, in particular for large systems more steps might mean it takes forever and the user can always take the final structure and minimize from there. It is just like the electronic steps in DFT, I believe the number should be sufficiently small, so if electronic convergence is an issue, people have to set this parameter manually, which gives other users of the same protocol a hint that the electronic convergence is a parameter which might require more specific tuning when switching the materials system. Ideally we would have an algorithm to set these convergence parameter depending on the system but until we have such an algorithm, I am in favor of failing quickly, rather than wasting computing resources.
> 
> But that is just my opinion, @liamhuber and @sudarsan-surendralal what is your take on this?

I have following four arguments for my defence:

1. The number of steps in conjugate gradient grows with the degree of freedom, i.e. if a system with 1e5 atoms requires 2e5 steps, it most certainly fails if there are 1e6 atoms. So you don't have to have chosen a bad system for CG to fail in this sense, as the number of atoms simply prevents it from converging.
2. Increasing the force/energy convergence accuracy by a factor of 10 will most likely increase the number of CG steps by 10, but increasing the number of maximum iterations does not affect most of the systems. It changes the real number of steps only in special cases.
3. It is true that the philosophy of pyiron is to offer a minimalistic parameter set. But then in my opinion it should be considered for the force/energy convergence, because it is possible to estimate the final errors on various thermodynamic quantities (e.g. energy, pressure etc.) from the force/energy residuals, but it is absolutely impossible to estimate those errors from the number of steps (i.e. you can certainly not tell "the error on total energy will be around XXX if the number of steps is 1e5", while it's possible to say "the error on total energy will be around XXX if the force convergence is 1e-2 eV/A").
4. I don't think 1e6 steps in LAMMPS is outrageously large. In contrast, I might rather say 1e5 is very small.

samwaseda:
> @samwaseda by the way I noticed that the azure tests started failing recently dude to some issue with Sphinx, maybe you can take a look at the logs and fix it in a separate pull request?

Hmmm it does not look like something I changed recently. I guess the thermodynamic constants given in scipy is not the same for the case that failed. But again I don't know how this problem could be solved...

jan-janssen:
Instead of comparing to a string you could compare to `str(scipy.constants.physical_constants["Hartree energy in eV"][0])`.

samwaseda:
> Instead of comparing to a string you could compare to `str(scipy.constants.physical_constants["Hartree energy in eV"][0])`.

I submitted it [here](https://github.com/pyiron/pyiron/pull/486). I'll merge as soon as it passes all the tests.

samwaseda:
> @samwaseda by the way I noticed that the azure tests started failing recently dude to some issue with Sphinx, maybe you can take a look at the logs and fix it in a separate pull request?

So this problem has been solved

liamhuber:
First, I want to use this as an opportunity to plug the [simultaneous perturbation stochastic approximation](https://en.wikipedia.org/wiki/Simultaneous_perturbation_stochastic_approximation) for minimization problems, which is beautiful and doesn't scale with the number of DoF's in the way CG does. I'm not suggesting that we should abandon CG and use this (it's anyhow not implemented in Lammps), I just think it's really cool math and everyone should know about it.

Second, I agree with @samwaseda that for Lammps 1e6 is not an outrageous number of steps. If Lammps is going to fail to converge, I am suspicious that 99% of the time it will be because things are so sever that the actual Lammps process will crash and we won't lose any computing resources. Anyhow, Lammps runs are generally not so expensive.

Finally, I'd like to raise a bit of a philosophical question: do we even *want* **different defaults for the same physics**. I think the answer will be 'yes', and that the 'pyironic' thing to do is to choose specialized defaults for each program we bind to. @jan-janssen's point about failing fast is extremely important for DFT calculations, and even 1e5 steps is crazy. If I think for a moment about the synchronization of defaults these are my conclusions:

Pros:
- It is truer to the pyironic maxim of not caring what code is interpreting your system
- May be easier to debug what's gone wrong if the code which is passing silently is always doing exactly the same thing 
- The user is forced to think a bit about what they're doing and how their specific calculation differs from the pyiron 'standard'

Cons:
- Impossible to satisfy all codes; users would need to think about what they are doing and write specific values for more input fields
- The practical reality of maintaining perfect consistency may be difficult; perhaps complete inconsistency is less of a headache for users than near-consistency

Ultimately I think I come down on the side of the status quo -- keep defaults optimized on a per-code basis. But I wanted to stop for a minute and think about the higher-level philosophy we're embracing by doing this. I think the core tension is between the pyironic values of 'code ambivalence' and 'users don't need to provide much input'.

samwaseda:
@liamhuber I agree that that's a valuable discussion (and we should keep thinking about it over and over again), but I'd say here the statement **different defaults for the same physics** doesn't fully apply, because there is no physics in conjugate gradient, as long as it does not converge. Also, 1e5 is already different from VASP and SPHInX, so in this regard changing it to 1e6 does not make the things worse.

PR #486, opened 2019-12-23
Reflecting Jan's comment on the unittest for SPHInX

samwaseda:


samwaseda:
I have the feeling that I'm opening Pandora's box now...

PR #489, opened 2020-01-06
Interactive input speedup

samwaseda:
Now the input has essentially the same structure. This has two important implications:

1. It can be made faster (up to 4 times, combined with the output stuff)
2. If the input is not defined (e.g. cell in DFT), it can be removed beforehand, which allows for a more secure input handling.

I add a notebook that I used for testing:

[interactive_speedup.ipynb.txt](https://github.com/pyiron/pyiron/files/4026409/interactive_speedup.ipynb.txt)

### DIFF
--- a/pyiron/atomistics/job/interactive.py
+++ b/pyiron/atomistics/job/interactive.py
@@ -32,6 +32,10 @@ class GenericInteractive(AtomisticGenericJob, InteractiveBase):
         self._interactive_fetch_completed = True
         self._interactive_species_lst = np.array([])
         self._periodic_table = PeriodicTable()
+        self.interactive_input_functions = {'index': self.interactive_index_organizer,
+                                            'cell': self.interactive_cell_organizer,
+                                            'positions': self.interactive_positions_organizer,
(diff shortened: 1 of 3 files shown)


liamhuber:
One thing I notice is that if a particular type of output isn't there once, it will never be found. I can't think of a code where this causes an issue, I just wanted to bring attention to the behaviour in case anyone else can think of a time it might be problematic. 

Looks good!

samwaseda:
> One thing I notice is that if a particular type of output isn't there once, it will never be found. I can't think of a code where this causes an issue, I just wanted to bring attention to the behaviour in case anyone else can think of a time it might be problematic.
> 

Yes! This is something I didn't like much as well, but this question goes beyond the changes that I made here - we have to talk about how to handle output data with different lengths.

liamhuber:
@samwaseda I guess it's a little disconcerting that the "speedup" pull request failed the Azure pipeline because of a timeout ;) I think you and Jan were talking about this as a general issue somewhere else? As long as it doesn't concern you I'm happy to have it merged.

jan-janssen:
I triggered a rerun on the failed job.

Issue #491, opened 2020-01-07
Pressure not properly rotated


samwaseda:
https://github.com/pyiron/pyiron/blob/77053fa5b0bddcc1f6f0e3c7a15b234985a165ba/pyiron/lammps/base.py#L544

If the upper triangle matrix is defined for Lammps, the pressure coordinates will be wrong -> it has to be rotated just in the same way as positions or velocities

jan-janssen:
@samwaseda As far as I understand @liamhuber has been working on this.

PR #496, opened 2020-01-14
Rotate pressures

liamhuber:
As per Sam's observation in [issue 491](https://github.com/pyiron/pyiron/issues/491).

I agree with the issue that this is low priority and this is just a naive fix -- I haven't tested it yet. @samwaseda, do you have a notebook where this issue came up that we could use to test this fix?


liamhuber:
Yes I think you're right.

I think I can also test this change numerically by doing two zero-pressure finite-T calculations with cubic and primitive cells; if the conversion is right the mean pressure tensor should come out to zero for both, but if we're converting wrong the primitive cell will do something else.
  This comment is basically just a reminder to myself though, since I don't think I'll get to it today. I'll try to get to it before January finishes though.

liamhuber:
Ok, I think it's working. The resulting pressure tensor is the target pressure tensor, and *as far as I can tell* playing around with cubic and primitive Al and applying some pressures, the correct pressure is being applied. I'm only, like, 85% sure of this though and I'd need to think more how to cook up a test to get that to 100%.

In the future I'd like to extract the unit conversions dictionary to a separate file and remove the remaining bar->GPa hard coding, but that can wait.

samwaseda:
Why did you add `rotation_matrix` as an argument?

liamhuber:
Because `control` had no access to `_prism.R` but still needed that information to convert the pyiron-coordinate-frame pressures to the Lammps-coordinate-frame.

samwaseda:
oh right that makes sense

liamhuber:
Yeah, it might be nice to refactor some stuff eventually, but for now it's good.

I'll get the failed tests all worked out prior to merging.

Thanks for the feedback and hints, @samwaseda!

samwaseda:
Since codacy doesn't like `@`, you can also think about using:

```
np.einsum('ji,njk,kl->nil', matrix, pressure, matrix)
```

`einsum` is anyway my favourite matrix multiplier, because it's really clear what does what.

liamhuber:
> Since codacy doesn't like `@`, you can also think about using:

I suspect this is simply a matter of 

> `einsum` is anyway my favourite matrix multiplier, because it's really clear what does what.

Really? I actually find `einsum` very difficult to follow. Based on the [length of the doc page](https://docs.scipy.org/doc/numpy/reference/generated/numpy.einsum.html), I would argue that my perspective will wind up being more common (although it is, admittedly one based on ignorance). In general, `einsum` encapsulates a variety of other tools, I would suggest using the more specific tools, e.g. `matmul`, `trace`, etc, wherever possible to improve readability.

samwaseda:
The problem with other functions is that it's not really clear which axis is multiplied with which axis. For example, in this case the rotation matrix is indeed 3x3, but what if it becomes nx3x3? With `np.einsum` you can easily switch between axes, but for other functions I never know what does what, especially when it for example comes to trace or transpose.

Well, other than that, it might be also my personal taste; anyway I write all algebraic operations in the Einstein notation and when I put them in the code it's essentially copying what I wrote on the paper.

liamhuber:
Good point, we're actually applying the matrix multiplication to one matrix and one tensor. When its (n x m x m) and (m x m) I still think it's unambiguous and preferable to use `matmul` and friends. Once both objects are tensors I guess we're stuck moving over to `einsum`.

jan-janssen:
@samwaseda and @liamhuber Can you take a look at this again and either fix the tests or close the pull requests?

samwaseda:
Did anything change? This looks good to me

jan-janssen:
@samwaseda The tests were not working

liamhuber:
I'm afraid the earliest I'd be able to look at it is the first Friday in June. @samwaseda if you can't fix the tests sooner we can just close it, but let's make sure there's a `bug` issue still open with a link to the branch because the underlying problem will still be there.

samwaseda:
> @samwaseda The tests were not working

I sometimes wonder how hopelessly stupid I must be appearing whenever I ask questions without fully reading what other people were saying.

Anyway, I guess I corrected the errors and as soon as the unit tests finish, I'll merge this one.

PR #497, opened 2020-01-14
Job copy logic

liamhuber:
The main thing is that copying jobs didn't work when the project value was explicitly set, but the same as the reference project. This is now not a problem and is equivalent to leaving the project value set to None.


pyiron/base/job/generic.py
```diff
@@ -482,58 +482,81 @@ def copy(self):
...
             else:
                 new_location = self.project_hdf5.__class__(
                     self.project, new_job_name, h5_path="/" + new_job_name
                 )
-            new_generic_job._name = new_job_name
+
             new_generic_job.project_hdf5.copy_to(new_location, maintain_name=False)
             new_generic_job.project_hdf5 = new_location
-            if new_database_entry:
-                new_generic_job.save()
```
jan-janssen: I am a bit confused - how does it work without the save function?

jan-janssen: Ok, I see the save part is below.

jan-janssen:
In general the code looks fine to me. But I realised that it took me quite some time to figure out again why we have the different cases and how we should handle them. One way to simplify this process would be to add some comments, copy within the same project and so on. I am not sure what is the best approach, maybe it even makes sense to have different functions for the different cases, that is what I did in the parallel master https://github.com/pyiron/pyiron/blob/master/pyiron/base/master/parallel.py#L553 . Basically the functions, copy, copy_to and rename share certain functionality and they have not been very clear for a long time, in particular when it comes to master jobs with childs nested within them. So we definitely need some clean up for this functionality.

jan-janssen:
To answer the question about the `from_hdf` functions, this is done within the `copy` function now.

liamhuber:
Yeah, you're right. I looked back at it and it's still too complicated. I gave it another pass and completely got rid of the nested conditions (which are the devil, IMO) by relying on `or` to define a couple new variables.

I can't seem to get `test_copyto.py` to run on my local machine, I think because it's actually creating jobs and stuff and I don't have the right paths set up? My "integration tests" -- i.e. the protocols where I'm actually extensively using `copy_to` are all running fine though, so I thought it's worth throwing at the CI.

liamhuber:
I also found a bug where one logical branch had `maintain_name=has_new_name` as an argument for `project_hdf5.copy_to`, but I think `maintain_name=not has_new_name` must be correct.

liamhuber:
@jan-janssen these failures don't look to me like they're related to the copying code, are they meaningful to you?

liamhuber:
Nevermind, I see the connection. I'm working on it...

liamhuber:
Ok, in my passion for simplified logic I actually borked the functionality on the last pass. Now the logic still avoids nested if clauses, but I think I've restored functionality.

I say think because I'm having trouble running the tests for this code on my local machine, but copying and pasting one of the previously-failing tests into a notebook shows it working fine. Let's see what the CI says this time around.

jan-janssen:
For running the tests locally can you try to add the `pyiron/tests` folder to your `top_level_paths` in the `.pyiron` file?

liamhuber:
Yeah, bet you're right. Actually I moved my repositories around recently so it's even possible that I had this earlier and forgot to fix it! I had to come home a bit early today but I'll give it a go first thing tomorrow.

I also see exactly why these tests failed and it's a simple one-line fix, so tomorrow should finally be merge day ;)

liamhuber:
@jan-janssen ok, the tests on the code are working. The coverage change is trivial. The azure pipeline fails trying to install something called "deeptools". If this is unrelated to my changes, let's merge them.

jan-janssen:
@liamhuber Yes, that is unrelated

PR #498, opened 2020-01-15
Lmp average

samwaseda:
Now the average of anything can be calculated. For example, if user wants to calculate potential energy square, it suffices to write: `job.input.control.measure_mean_value('pe^2', 10, name='energy_pot_sq')`, where `name` is what is going to be stored in the hdf5 file (preceded by `mean_`).

### DIFF
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -175,20 +175,21 @@ class LammpsControl(GenericParameters):
     def load_default(self, file_content=None):
         if file_content is None:
             file_content = (
-                "units               metal\n"
-                + "dimension           3\n"
-                + "boundary            p p p\n"
-                + "atom_style          atomic\n"
-                + "read_data           structure.inp\n"
-                + "include             potential.inp\n"
-                + "fix___ensemble      all nve\n"
-                + "variable___dumptime equal 100\n"
-                + "dump___1            all custom ${dumptime} dump.out id type xsu ysu zsu fx fy fz vx vy vz\n"
-                + "dump_modify___1     sort id format line \"%d %d %20.15g %20.15g %20.15g %20.15g %20.15g %20.15g %20.15g %20.15g %20.15g\"\n"
-                + "thermo_style        custom step temp pe etotal pxx pxy pxz pyy pyz pzz vol\n"
-                + "thermo_modify       format float %20.15g\n"
-                + "thermo              100\n"
-                + "run                 0\n"
+                "units                 metal\n"
+                + "dimension             3\n"
+                + "boundary              p p p\n"
+                + "atom_style            atomic\n"
+                + "read_data             structure.inp\n"
+                + "include               potential.inp\n"
+                + "fix___ensemble        all nve\n"
+                + "variable___dumptime   equal 100\n"
+                + "variable___thermotime equal 100\n"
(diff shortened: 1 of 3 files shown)


samwaseda:
This is rather meant to be used by power users, who know how variables are defined in LAMMPS and have to reduce the number of output data by making LAMMPS calculate average values internally. Otherwise I assume it makes more sense for normal users to launch LAMMPS calculation with a low `n_print` value and do the statistics afterwards.

pyiron/lammps/control.py
```diff
@@ -641,13 +641,14 @@ def calc_vcsgc(
             append_if_not_present=True
         )
 
-    def measure_mean_value(self, key, every=1, repeat=None, freq=None):
+    def measure_mean_value(self, key, every=1, repeat=None, freq=None, name=None):
         """
             Args:
```
liamhuber: Please add the first comment + first sentence of your second comment from the 'conversation' section of the pull request to the docstring. They might need light formatting, but I think they really clearly communicate the intention of this function.

pyiron/lammps/control.py
```diff
@@ -641,13 +641,14 @@ def calc_vcsgc(
...
                 key (str): property to take an average value of (e.g. 'energy_pot' v.i.)
                 every (int): number of steps there should be between two measurements
                 repeat (int): number of measurements for each output (default: n_print/every)
                 freq (int): output frequency (default: n_print)
```
liamhuber: I see that this line wasn't code edited in this pull request, but I think we should take the opportunity to remove explicit reference to `n_print`. My concern is that if we do a variable name change in `calc_md` these docs could get out of date. Of course, in principle refactoring the variable name might catch this usage, but since it's in documentation I'm nervous.

Rather, could we simply refer to either the actual code used below (i.e. `self['thermo']`), or --and this would be my preference -- to the logical intent, e.g. `freq (int): Output frequency. (Default is None, which uses the same output frequency as the underlying calculation.)`

liamhuber: Further, can we change this from `freq` to `period`?

samwaseda: `freq` follows the convention from [this page](https://lammps.sandia.gov/doc/fix_ave_time.html)

samwaseda: ok then I'll define a variable inside LAMMPS for `n_print` and will replace `freq` by that.

liamhuber: Hmmph. Lammps is wrong. :p 

My preference would be to call the variable something like `period` and make a comment in the docstring that Lammps (*cough*wrongly*cough*) calls it `freq`, but I would also accept leaving it as `freq` and being very explicit in the docstring that it has units of period.

samwaseda: Now, I entirely got rid of `Nfreq`, because I don't see why this output should be performed with a different frequency than the output of all the other properties.

tests/lammps/test_interactive.py
```diff
@@ -84,9 +84,10 @@ def test_interactive_execute(self):
             [
                 "fix ensemble all nve",
                 "variable dumptime equal 100",
+                "variable thermotime equal 100",
```
liamhuber: I like.

PR #499, opened 2020-01-15
Hdf

samwaseda:
`__setitem__` is defined for hdf. In particular, user can directly write things like `job['user/output/acceptance_ratio'] = 0.3`. As both `user` and `output` are not predefined in pyiron, this new function creates groups recursively.

### DIFF
--- a/pyiron/base/job/core.py
+++ b/pyiron/base/job/core.py
@@ -801,6 +801,34 @@ class JobCore(PyironObject):
         del self._master_id
         del self._status
 
+    def _create_groups_recursively(self, hdf, key, value):
+        """
+        This function creates a new group if a group is not defined, simply passes to the
+        next level of hdf5 hierarchy or stores data if the key is already in the lowest level
+
+        args:
+            hdf (str): (partial) path to data
+            key (str): hdf key
+            value (anything): value to store
+        """
+        if key[0] not in hdf.list_groups():
+            hdf.create_group(key[0])
+        with hdf.open(key[0]) as hdf_group:
+            if len(key)==2:
+                hdf_group[key[1]] = value
+            else:
+                self._create_groups_recursively(hdf_group, key[1:], value)
+
+    def __setitem__(self, key, value):
+        """
+        stores data
+        
+        args:
+            key (str): key to store in hdf (full path)
+            value (anything): value to store
+        """
+        self._create_groups_recursively(self._hdf5, key.split('/'), value)
+
     def __getitem__(self, item):
         """
         Get/ read data from the HDF5 file
--- a/pyiron/base/job/core.py
+++ b/pyiron/base/job/core.py
@@ -822,7 +822,7 @@ class JobCore(PyironObject):
     def __setitem__(self, key, value):
         """
         stores data
-        
+
         args:
             key (str): key to store in hdf (full path)
             value (anything): value to store
--- a/pyiron/base/job/core.py
+++ b/pyiron/base/job/core.py
@@ -843,8 +843,8 @@ class JobCore(PyironObject):
             key (str): key to store in hdf (full path)
             value (anything): value to store
         """
-        if key.startswith('output/') or key.startswith('input/'):
-            raise ValueError('output and input cannot be changed by user -> use a different path')
+        if not key.startswith('user/'):
+            raise ValueError("user defined paths+values must begin with user/, e.g. job['user/key'] = value")
         self._hdf5[key] = value
 
     def __delitem__(self, key):
--- a/pyiron/base/job/core.py
+++ b/pyiron/base/job/core.py
@@ -827,6 +827,8 @@ class JobCore(PyironObject):
             key (str): key to store in hdf (full path)
             value (anything): value to store
         """
+        if key.startswith('output/') or key.startswith('input/'):
+            raise ValueError('output and input cannot be changed by user -> use a different path')
         self._create_groups_recursively(self._hdf5, key.split('/'), value)
 
     def __getitem__(self, item):
--- a/pyiron/base/job/core.py
+++ b/pyiron/base/job/core.py
@@ -801,36 +801,6 @@ class JobCore(PyironObject):
         del self._master_id
         del self._status
 
-    def _create_groups_recursively(self, hdf, key, value):
-        """
-        This function creates a new group if a group is not defined, simply passes to the
-        next level of hdf5 hierarchy or stores data if the key is already in the lowest level
-
-        Args:
-            hdf (str): (partial) path to data
-            key (str): hdf key
-            value (anything): value to store
-        """
-        if key[0] not in hdf.list_groups():
-            hdf.create_group(key[0])
-        with hdf.open(key[0]) as hdf_group:
-            if len(key)==2:
-                hdf_group[key[1]] = value
-            else:
-                self._create_groups_recursively(hdf_group, key[1:], value)
-
-    def __setitem__(self, key, value):
-        """
-        Stores data
-
-        Args:
-            key (str): key to store in hdf (full path)
-            value (anything): value to store
-        """
-        if key.startswith('output/') or key.startswith('input/'):
-            raise ValueError('output and input cannot be changed by user -> use a different path')
-        self._create_groups_recursively(self._hdf5, key.split('/'), value)
-
     def __getitem__(self, item):
         """
         Get/ read data from the HDF5 file
@@ -864,6 +834,18 @@ class JobCore(PyironObject):
             with open(file_name) as f:
                 return f.readlines()
         return None
+
+    def __setitem__(self, key, value):
+        """
+        Stores data
+
+        Args:
+            key (str): key to store in hdf (full path)
+            value (anything): value to store
+        """
+        if key.startswith('output/') or key.startswith('input/'):
+            raise ValueError('output and input cannot be changed by user -> use a different path')
+        self._hdf5[key] = value
 
     def __delitem__(self, key):
         """
--- a/pyiron/base/job/core.py
+++ b/pyiron/base/job/core.py
@@ -806,7 +806,7 @@ class JobCore(PyironObject):
         This function creates a new group if a group is not defined, simply passes to the
         next level of hdf5 hierarchy or stores data if the key is already in the lowest level
 
-        args:
+        Args:
             hdf (str): (partial) path to data
             key (str): hdf key
             value (anything): value to store
@@ -821,9 +821,9 @@ class JobCore(PyironObject):
 
     def __setitem__(self, key, value):
         """
-        stores data
+        Stores data
 
-        args:
+        Args:
             key (str): key to store in hdf (full path)
             value (anything): value to store
         """
--- a/tests/base/job/test_hdf5content.py
+++ b/tests/base/job/test_hdf5content.py
@@ -47,6 +47,11 @@ class DatabasePropertyIntegration(unittest.TestCase):
             job_inspect.content.output.__repr__(), job_inspect["output"].__repr__()
         )
 
+    def test_setitem(self):
+        self.ham['user/output/some_value'] = 0.3
+        self.assertEqual(self.ham['user/output/some_value'], 0.3)
+        with self.assertRaises(ValueError):
+            self.ham['input/value'] = 1
 
 if __name__ == "__main__":
     unittest.main()


pyiron/base/job/core.py
```diff
@@ -801,6 +801,36 @@ def __del__(self):
...
+            if len(key)==2:
+                hdf_group[key[1]] = value
+            else:
+                self._create_groups_recursively(hdf_group, key[1:], value)
+
+    def __setitem__(self, key, value):
+        """
+        stores data
+
+        args:
```
liamhuber: Nitpick: 'args' should be capitalized, and capitalization for the head of the docstring is consistent with the majority of the code.

samwaseda:
You mean why we need this?

liamhuber:
Yes, what is the point? Why do I want to mess directly with the HDF5? Assuming I want to do this sort of thing, why not have some area of the python object that I can mess with and which is automatically saved and loaded?

samwaseda:
It follows [this discussion](https://gitlab.mpcdf.mpg.de/pyiron/pyiron-support/issues/151)

samwaseda:
One example was exactly the acceptance ratio when I was doing a Monte Carlo simulation, which was an output value but was not stored in pyiron, but still it's a property that should be assigned to a job.

Another example was when I was calculating C-C interactions in Fe, where I consider the symmetries and therefore needed the degeneracy of each configuration for the Boltzmann statistics later on. Of course I could store the info in an independent variable, but if I know that one specific configuration appears for example 4 times, then it makes sense that the job knows this info (so that I can simply multiply exp(-E/kBT) by 4) without risking to mess up with independent variables.

liamhuber:
Sounds good. I still find it very strange to modify the HDF directly. Can we not attach it to the job directly under a `user` or `user_data` attribute?

@dgehringer wrote some [nice code](https://github.com/pyiron/pyiron_contrib/blob/master/pyiron_contrib/protocol/utils/dictionaries.py#L110) for writing complex datatypes to hdf over in `pyiron_contrib`. I guess we could just use this and add something like `self.user_data.to_hdf()` to `GenericJob.to_hdf()`.

samwaseda:
I agree. That's actually the reason why I forbade to edit `output` and `input`, so `job['output/some_data'] = 1` will raise an error

pyiron/base/job/core.py
```diff
@@ -801,6 +801,36 @@ def __del__(self):
         del self._master_id
         del self._status
 
+    def _create_groups_recursively(self, hdf, key, value):
```
jan-janssen: OK, now that I see the functions maybe it makes more sense to move them to `hdfio` . I am sorry for the confusion.

samwaseda:
I just realised that what I wanted to do was already there via `job._hdf5[key] = value`. But still I keep the first version that I proposed above while the code has just been made a bit slimmer.

jan-janssen:
So currently you write directly to use HDF5 root level `/` rather than `/user` ? This might be an issue for Master jobs which also store their childs there.

samwaseda:
> So currently you write directly to use HDF5 root level `/` rather than `/user` ? This might be an issue for Master jobs which also store their childs there.

ok that's an important remark... So better make `__setitem__` store the data in `['user/'+key]`?

jan-janssen:
Yes, I think that would be better.

samwaseda:
The problem with this is that it could be confusing if the user stores the data via:

```
job['output/some_value'] = some_value
```

And then there's `job['output']`, but it's not there that `some_value` is stored, but `job['user/some_value']`. Another possibility is to force the user to choose a key that starts with `user`, i.e. you always have to choose `job['user/output/some_value']`.

jan-janssen:
Or we simply require the user to add `user` in front - then we have to tell him in the documentation that only `job['user/..']` is writable.

liamhuber:
Users can't add attributes to the `job.output` object can they? Aren't these issues resolved by having a `job.user` object that users can read to and write from? I also still think it's really weird to have people directly working with the hdf5 file. Is there a reason I don't see to avoid this setup?

samwaseda:
> Users can't add attributes to the `job.output` object can they? Aren't these issues resolved by having a `job.user` object that users can read to and write from? I also still think it's really weird to have people directly working with the hdf5 file. Is there a reason I don't see to avoid this setup?

`job.user` will not allow for structured data storage, e.g. `job['user/input/some_value']` won't possible, i.e. it does not create new groups or nodes. And then I find it important that part of hdf5 file is available to the user, especially if we know that it does not conflict with the rest of what pyiron stores.

jan-janssen:
I hope I can clarify this a bit. The GenericMaster class can contain other nested jobs within itself. In this case the jobs are stored under the root level next to `output` and `input`. At the moment users could already write data to the HDF5 file but they would have to use `project_hdf`. Now what Sam is working on is exposing this interface under `job.user`. So instead of calling `job.project_hdf['output/generic/energy_tot'] = np.array([])` we can then use `job.user['test']` which equals `job.project_hdf['user/test'] `

liamhuber:
Thanks Jan, that helps clarify some stuff.

> So instead of calling `job.project_hdf['output/generic/energy_tot'] = np.array([])` we can then use `job.user['test']` which equals `job.project_hdf['user/test']`

I still don't understand why we can't use `job.user.test`, which would be saved to (and could in principle be read from as desired) `job.user['test']`.

My understanding is that `job.user.test` would not exist in the proposed solution. My assertion is that such an object-oriented interface instead of *only* being able to access the hdf file directly with `job.user['test']` is (1) desirable, and (2) achievable.

samwaseda:
Ok then do I understand correctly, that in principle it is not a problem that the user can store data in `job['user/..']`, only it should be accessed via `job.user` instead of manipulating the HDF5 directly? I guess this is at some point desirable, but with or without this pull request, not everything is accessible by functions in the current implementation, and in particular in the `inspect` mode there's no way to use functions.

Another concern of mine which I haven't mentioned is that I'm not sure if it's intuitively clear to new users what `job.user` means. Furthermore, if data storage is extended even more in the future, we'll have more attributes like this. I think it's not quite wrong to have this dictionary format (i.e. `job['output']` etc.) which essentially should mean it's data storage.

samwaseda:
I was talking with Jan and another argument came to my mind: Since now we know that `job._hdf5` and `job.project_hdf5` can actually change everything in the HDF5 file, I think it's important to create a functionality, which allows the user to store data, before they even notice that there are `job._hdf5` etc. I think a pyiron beginner would intuitively try out something like `job['output/..'] = some_value`. Then they get an error message saying *Use `job['user/..']` to store data*, then they won't look for other possibilities to store data and therefore won't risk editing important content. I'm fine with the idea that the error message is going to be replaced at some point by *Use `job.user` instead*.

liamhuber:
I think I'm still not being clear enough. I have zero problem with people accessing `job['user/stuff']`, I just don't think that should be the *exclusive* way of accessing `stuff`. We can set `job.input.foo = 5` and later `print(job.output.bar)`. We can also `print(job['output/bar'])`. This is all great. 

If we now have a place for users to add custom stuff that gets saved to hdf5, it should look as much like that as possible. i.e. we should let people write *all* of the following: `job.user.foo = 5`, `print(job.user.foo)`, `print(job['user/foo']`. 

It doesn't *have* to happen in this pull request, but with `_generic_to_hdf` already available I don't understand why it *shouldn't* happen in this pull request. "Ok, but I don't want to code that right now" is a totally fine reason to me, but first I want to have what I'm requesting be understood ;)

Finally, I agree `.user` might now be the best field name. Maybe `.user_data` or `.user_storage`? I'm certainly open to alternatives to `.user`.

samwaseda:
It seems to me that `_generic_to_hdf` allows the user to store complex data, but it doesn't look like it automatically generates attributes from the tags, or does it do that as well? (i.e. if there's `job['user/stuff']`, `job.user.stuff` will be available automatically?)

liamhuber:
`_generic_to_hdf` and `_generic_from_hdf` are methods of `IODict`. `IODict` does other things to, but in this comment I'll just make use of those methods and you can imagine adding a *similar* class as `job.user` (or whatever you want to call it.). 

Our protocols use an `IODict` for their `output` field, among other things. Right now there are no restrictions stopping the user from writing to that `output` field, so I'll just use this as an example to show how you can write to it, save, and load it. Here is a small example (assuming you have `pyiron_contrib` cloned and in your path):

```python
from pyiron import Project
import numpy as np
pr = Project('generic_hdf_demo')
pr.remove_jobs()

has_generic = pr.create_job(pr.job_type.ProtocolMD, 'has_generic')
has_generic.output.foo = 1
has_generic.output.bar = ['a', 'b', 'c']
has_generic.output.baz = {'baz1': 1, 'baz2': 'a', 'baz3': np.random.rand(4)}

print(has_generic.output.bar)
> ['a', 'b', 'c']

# Save the job to hdf5 (which calls the IODict's to_hdf method) and it's all there
has_generic.save()
print(has_generic['output/foo/'])
> 1
# For lists the hdf5 tags are based on the index
print(has_generic['output/bar/'])
> {'groups': [], 'nodes': ['FULLNAME', 'TYPE', 'i_0', 'i_1', 'i_2']}
print(has_generic['output/bar/i_1'])
> b
print(has_generic['output/baz'])
> {'groups': [], 'nodes': ['FULLNAME', 'TYPE', 'baz1', 'baz2', 'baz3']}
print(has_generic['output/baz/baz3'])
> [0.64446157 0.91924444 0.94277681 0.71080347]

# We can load the job later and it's all there
loaded = pr.load('has_generic')
# We can access it like an object
print(loaded.output.bar)
> ['a', 'b', 'c']
print(loaded.output.baz['baz2'])  # baz is just some dict, so we need square brackets
> a
# And we can access the hdf5 directly
print(loaded['output/baz/baz3'])
> [0.64446157 0.91924444 0.94277681 0.71080347]
```

I'm not saying it's *exactly* as simple as adding `self.user = IODict()` to `GenericJob.__init__` and `self.user.to_hdf(group='user')` to `GenericJob.to_hdf`, because IODict does other stuff we don't need (mostly to do with pointers, which are only an issue for the protocols.) But I think it should be close to this easy.

liamhuber:
Again, the credit for this utility is to @dgehringer.
I think we might run into trouble if the user tries to save some strange class type, but primitive data types, ndarrays, `Atoms`, lists, and dictionaries are all totally fine.

I forgot to show, but you can nest things without any trouble
```python
...
has_generic.output.boa = {'depth1': {'depth2': {'list': ['a', 'b', 'c']}}}
print(has_generic['output/boa/depth1/depth2/list/i_2'])
> c
...
print(loaded.output.boa)
> {'depth1': {'depth2': {'list': ['a', 'b', 'c']}}}

samwaseda:
Ok but then my implementation is not incompatible with `_generic_to_hdf`, since mine does not do more than creating the group `user`; it probably should be replaced at some point by a more object oriented method like `_generic_to_hdf`. Also, I'm not trying to encourage people to use this new functionality. On the contrary, with this I might simply discourage people from editing important information in pyiron, since they will be allowed to edit only `user` in this case.

liamhuber:
> Ok but then my implementation is not incompatible with `_generic_to_hdf`, since mine does not do more than creating the group `user`; it probably should be replaced at some point by a more object oriented method like `_generic_to_hdf`.

Can we replace it now? I still don't see why we don't just trim down `IODict` to a simpler class and make the two line changes to `GenericJob`. If it's more complicated than that I'm ok leaving it off, but if we agree the object/hdf duality provided by `_generic...` is a superior solution and simple enough to implment, let's just use it.

> Also, I'm not trying to encourage people to use this new functionality. On the contrary, with this I might simply discourage people from editing important information in pyiron, since they will be allowed to edit only `user` in this case.

Fair enough. I wouldn't encourage it either, but I don't want to discourage it by making it look very different from how the rest of the job works (i.e. with object/hdf duality).

samwaseda:
> Can we replace it now? I still don't see why we don't just trim down `IODict` to a simpler class and make the two line changes to `GenericJob`. If it's more complicated than that I'm ok leaving it off, but if we agree the object/hdf duality provided by `_generic...` is a superior solution and simple enough to implment, let's just use it.

I meant the way people store data should be replaced, but as you said I think it's good to offer the possibility of `job['user/key'] = value`. So if the `_generic_to_hdf` is added, we can replace the error message in `__setitem__` by *Use `job.user` instead*, but the implementation of this pull request should stay in place. After all, this pull request is literally only about defining `__setitem__` - which has been absent so far - to tell the user how user-defined data can be stored in the job hdf5. I don't think we'll have to change anything later on.

liamhuber:
>  So if the `_generic_to_hdf` is added, we can replace the error message in `__setitem__` by _Use `job.user` instead_, but the implementation of this pull request should stay in place. 

Yep, good point. I'm happy now.

PR #502, opened 2020-01-17
Drop py27

jan-janssen:


jan-janssen:
It seems like Appveyor does not support Python 3.8 - at least not the miniconda version we are using for Python 3.7

jan-janssen:
The conda packages seem to be not sufficient for Python 3.8 - travis gets stuck in resolving the dependencies.

PR #509, opened 2020-01-21
Lmp temperature

samwaseda:
This implementation allows the user to define a slope of temperature during MD. I personally think it would be a very rare case, but still I guess it's right to offer this possibility.


pyiron/lammps/control.py
```diff
@@ -411,11 +411,14 @@ def calc_md(
 
         # Transform temperature
         if temperature is not None:
+            temperature = np.array([temperature], dtype=float).flatten()
+            if len(temperature)==1:
+                temperature = np.array(2*temperature.tolist())
```
liamhuber:
```
elif len(temperature) > 2:
    raise ValueError("At most two temperatures can be provided (for a linearly ramping target temperature), but got {}".format(len(temperature))
```

samwaseda: Brilliant!

samwaseda:
> Oh, one more change: The docstring for `control.calc_md` will need to be changed. I'm pretty sure I set it up so that the other Lammps `calc_md` calls just pull their docstrings from there, so you should only need to update it in the one place.

I saw the same DocString in `base.py` and `control.py`, so I modified both of them (I don't know if this was necessary or not), but it's true that it makes sense to automatically copy one to the other (which I don't know how to do though)

liamhuber:
>  but it's true that it makes sense to automatically copy one to the other (which I don't know how to do though)

It mostly makes sense. The problem I have is that the methods in `control` sometimes have extra arguments, e.g. in our recent discussion over at the [pressure pull request](https://github.com/pyiron/pyiron/pull/496), where I added that `rotation_matrix` argument.

The *how* is actually super simple. I thought I'd done it for `calc_md`, but apparently I only did it for `calc_minimize`. If you look at `pyiron.lammps.base.LammpsBase.calc_minimize`, right after the function definition is the line `calc_minimize.__doc__ = LammpsControl.calc_minimize.__doc__`. That's it. When you're using the code, `?` automatically pulls up the root `__doc__`. If you're actually working with the codebase you need to go somewhere else to see the docstring, but in pycharm (and I imagine most other editors) that's as simple as ctrl+clicking on the assigned docstring.

For those two minor downsides, I think it's worth centralizing our docstring maintenance.

PR #511, opened 2020-01-22
Water analysis

sudarsan-surendralal:
Using correct variable names!


sudarsan-surendralal:
This was written ages ago so there might not be good justification for this! But if I remember, set_neighborhood() is called by the user when the bond radius is manually modified and the user wants to recalculate the neighbor list. That's why this is public.

samwaseda:
But it's recalculated anyway automatically when `bond_cutoff` is called

PR #513, opened 2020-01-23
Pot per atom sam

Alaukiksaxena:
This function calculates the potential energy of a particular atom in a given structure.


samwaseda:
The only one thing could be that in pyiron potential energy is generally called `energy_pot`, so it might be more consistent to call it `self.input.control.energy_pot_per_atom` (although I'm fine with `self.input.control.pot_per_atom` too)

PR #516, opened 2020-02-08
Lammps mpi lib

jan-janssen:
For some reason the interface is unstable, some times it works some times it does not. This seem to be related to the Lammps commands used, like `change_box remap`.


jan-janssen:
The instability bug is fixed by reducing buffer size.

jan-janssen:
It works for 2000 atoms while the previous implementation failed at 1000 atoms - still both is not a reasonable size for parallelisation.

jan-janssen:
After several hours of testing I realised that the issue is in mpich - switching to openmpi solves the issue and already the previous implementation works.

Issue #520, opened 2020-02-13
bug in set_empty_states in sphinx


FeLoch:
Empty states can never be set!

https://github.com/pyiron/pyiron/blob/010444ae03d099909a0ae94ddd5086d26ae6528e/pyiron/sphinx/base.py#L577

FeLoch:
pull request open #521

jan-janssen:
I do not get it, what was wrong with the previous code? if the number is smaller than `0` it raises and issue and if it is `None` nothing happens.

FeLoch:
If `n_empty_states` is not None it raises an error, too. Therefore it always raises an error.

jan-janssen:
Got it, it should have been is `not None and `instead of is `not None or `

FeLoch:
Moreover, the user is never able to restore `"auto"` for the emtpy states.

PR #521, opened 2020-02-13
bugfix set_empty_states

FeLoch:
See #520


jan-janssen:
@samwaseda Can you take a look? For me it looks fine.

samwaseda:
Looks great :) But was it really a bug?

Maybe a small unit test...

samwaseda:
> Looks great :) But was it really a bug?
> 

ok now I understood what was the problem in the previous version

samwaseda:
Very good!

PR #526, opened 2020-02-18
Apply strain

samwaseda:
Apply a strain with a single number / 3-dim vector / epsilon matrix.

With `return_box` only the returned box will be strained

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -3121,14 +3121,15 @@ class Atoms(object):
         epsilon.reshape(3,3)
         if epsilon.min()<-1.0:
             raise ValueError("Illegal strain value")
-        structure_copy = self.copy()
+        if return_box:
+            structure_copy = self.copy()
+        else:
+            structure_copy = self
         cell = structure_copy.cell.copy()
         cell = np.matmul(epsilon+np.eye(3), cell)
         structure_copy.set_cell(cell, scale_atoms=True)
         if return_box:
             return structure_copy
-        else:
-            self = structure_copy
 
     def get_initial_magnetic_moments(self):
         """
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -3106,6 +3106,30 @@ class Atoms(object):
             else:
                 self.selective_dynamics[atom_ind] = [False, False, False]
 
+    def apply_strain(self, epsilon, return_box=False):
+        """
+            Args:
+                epsilon (float/list/ndarray): epsilon matrix. If a single number is set, the same strain
+                                              is applied in each direction. If a 3-dim vector is set, it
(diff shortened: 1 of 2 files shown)


pyiron/atomistics/structure/atoms.py
```diff
@@ -3106,6 +3106,31 @@ def set_constraint(self, constrain):
...
+                                              is applied in each direction. If a 3-dim vector is set, it
+                                              will be multiplied by a unit matrix.
+                return_box (bool): whether to return a box. If set to True, only the returned box will 
+                                   have the desired strain and the original box will stay unchanged.
+        """
+        epsilon = np.array([epsilon]).flatten()
+        if len(epsilon) == 3 or len(epsilon) == 1:
+            epsilon = epsilon*np.eye(3)
+        epsilon.reshape(3,3)
+        if epsilon.min()<-1.0:
```
liamhuber: nit: `if epsilon.min() < -1.0`  whitespace

pyiron/atomistics/structure/atoms.py
```diff
@@ -3106,6 +3106,31 @@ def set_constraint(self, constrain):
...
+            raise ValueError("Illegal strain value")
```
liamhuber: nit: Can be more informative and let user know why it was illegal, i.e. too negative

jan-janssen:
Looks good to me, maybe you can also update Murnaghan master to use this new method https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/master/murnaghan.py#L309

samwaseda:
> Very nice. I would only add that I'd like to see this method called in `Atoms.__[r]mul__` and `Atoms__imul__` under the condition that the multiplying value is a 3x3 array or matrix. (And I guess we'd need to subtract `np.eye(3)` off of the multiplying value in that case).

That would mean that `cell` has to get `__mul__`, right? I guess that would slightly deviate from what is written in ASE, so I might want to avoid it (as I'm envisaging a convergence to ASE at some point....)

liamhuber:
> That would mean that `cell` has to get `__mul__`, right?

Cell is type `numpy.ndarray` and already has the correspondning `__mul__` behaviour. If you mean we need to make a custom `Cell` class and modify it's `__mul__` behaviour, then no, I don't see why we'd need to do that. we can just do something like:

```python
Atoms.__mul__(self, other):
    # other stuff, then
    self.cell = np.matmul(self, other)
   # more other stuff
```

Right now as it is, people can directly modify a structure's cell, but it does exactly that. It modifies the cell and leaves the atoms alone. e.g.)

```python
struct = pr.create_ase_bulk('Al', cubic=True)
struct.cell *= 2
```

Just gives you a unit cell of atoms taking up one eights of a cubic box.

I can just as easily do `struct.positions *= np.pi` and get similar nonsense.

I'm only proposing that we give sophisticated behaviour to the `Atoms` class itself. If people want to mess around with the constituent arrays by themselves, then on their heads be it.

PR #527, opened 2020-02-19
Kpoints line mode sphinx

FeLoch:
Added Line mode analogous to VASP. Also small PEP8 corrections were done :)


pyiron/sphinx/base.py
```diff
@@ -614,22 +616,57 @@ def _set_kpoints(
...
+                        + ", "
+                        + str(center_shift[2])
+                        + "]"
+                )
+        elif scheme == "Line":
+            if n_trace is None:
+                raise ValueError("n_trace has to be defined")
+            high_symmetry_points = self.structure.get_high_symmetry_points()
+            if high_symmetry_points is None:
+                raise ValueError("high_symmetry_points has to be defined")
```
samwaseda: Is it up to the user to define `high_symmetry_points`?

jan-janssen: I thought they are given by `high_symmetry_points = self.structure.get_high_symmetry_points()` - but maybe the user should have the option to overwrite those.

samwaseda: I'm fine with `high_symmetry_points = self.structure.get_high_symmetry_points()`, but then the error message is misleading.

pyiron/sphinx/base.py
```diff
@@ -614,22 +616,57 @@ def _set_kpoints(
...
-                + ", "
-                + str(center_shift[1])
-                + ", "
-                + str(center_shift[2])
-                + "]"
+        if scheme == "MP":
+            self._kpoints_odict = None
+            if mesh is not None:
+                self.input["KpointFolding"] = (
+                        "[" + str(mesh[0]) + ", " + str(mesh[1]) + ", " + str(mesh[2]) + "]"
```
samwaseda: How about writing something like `np.array2string(mesh, separator=',')`?

pyiron/sphinx/base.py
FeLoch: The user has to define the `high_symmetry_points` for the `structure` object. Since they are structural quantities, they should not be defined within `set_kpoints`.

FeLoch: @samwaseda you're right. I'll change the message.

pyiron/sphinx/base.py
FeLoch: Sure, but I didn't implement that function.

pyiron/sphinx/base.py
```diff
@@ -637,7 +637,7 @@ def _set_kpoints(
                 raise ValueError("n_trace has to be defined")
             high_symmetry_points = self.structure.get_high_symmetry_points()
             if high_symmetry_points is None:
-                raise ValueError("high_symmetry_points has to be defined")
+                raise ValueError("no 'high_symmetry_points' defined for 'structure'.")
```
samwaseda: but does `self.structure.get_high_symmetry_points()` only returns `self.structure.high_symmetry_points`? Isn't the function then redundant ?

FeLoch: I will rewrite the `high_symmetry_points` part in near future. For me it is fine at the moment.

pyiron/sphinx/base.py
FeLoch: done

PR #533, opened 2020-02-23
Fix for SerialMaster

jan-janssen:
Enforce run_mode modal when submitting to the queue - it is important that the master process never stops, otherwise the queuing system is going to kill the job. At least modern queuing systems like Slurm do. @sudarsan-surendralal this should fix your issue with the SerialMaster.


sudarsan-surendralal:
Thanks, now it works! I am going to merge this.

PR #535, opened 2020-02-26
property setter for cell removed

samwaseda:
Following last week's discussion

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1075,12 +1075,12 @@ class Atoms(object):
         return len(self)
 
     def set_absolute(self):
-        warnings.warn("set_relative is deprecated as of 2020/02/26. It is not guaranteed from v. 0.3")
+        warnings.warn("set_relative is deprecated as of 2020/02/26. It is not guaranteed from v. 0.3", DeprecationWarning)
         if self._is_scaled:
             self._is_scaled = False
 
     def set_relative(self):
-        warnings.warn("set_relative is deprecated as of 2020/02/26. It is not guaranteed from v. 0.3")
+        warnings.warn("set_relative is deprecated as of 2020/02/26. It is not guaranteed from v. 0.3", DeprecationWarning)
         if not self._is_scaled:
             self._is_scaled = True
 
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -251,7 +251,6 @@ class Atoms(object):
             else:
                 self.set_cell(value)
 
-
     @property
     def species(self):
         """
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -548,10 +548,10 @@ class Atoms(object):
                 self.dimension = hdf_atoms["dimension"]
                 self.units = hdf_atoms["units"]
 
-                self.cell = None
+                self._cell = None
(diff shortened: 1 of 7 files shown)


samwaseda:
I'm wondering if `set_relative` makes sense hereafter ... ?

jan-janssen:
We are breaking ASE compatibility with this change: 

```
from ase.build import bulk
structure = bulk('Cu', 'fcc', a=3.6, cubic=True)
structure.cell *= 0.5
structure.cell
```

This works in ASE and this is no longer working with pyiron if we adopt your changes, correct?

samwaseda:
That's true, but the problem is that in pyiron it's just an attribute and in ASE it's an object, right? I guess in the current implementation it's not possible to make `__imul__` available while the property setter is disabled

samwaseda:
In the end, I restored `cell` but put a deprecation warning on `set_relative` and `set_absolute`, since they do not mean anything anymore

jan-janssen:
Here is one of the examples where we still use the _is_scaled functionality: 
https://github.com/pyiron/pyiron/blob/master/pyiron/lammps/interactive.py#L108

pyiron/atomistics/structure/atoms.py
```diff
@@ -239,6 +239,8 @@ def cell(self):
         numpy.ndarray: A size 3x3 array which gives the lattice vectors of the cell as [a1, a2, a3]
 
         """
+        if self._cell is None:
```
jan-janssen: Is not this redundant? `if self._cell is None`, then return `self._cell` returns `None`

samwaseda: yes this because redundant again :D

pyiron/atomistics/structure/atoms.py
```diff
@@ -552,10 +555,10 @@ def from_hdf(self, hdf, group_name="structure"):
                 self.dimension = hdf_atoms["dimension"]
                 self.units = hdf_atoms["units"]
 
-                self.cell = None
+                self._cell = None
                 if "cell" in hdf_atoms.list_groups():
                     with hdf_atoms.open("cell") as hdf_cell:
-                        self.cell = hdf_cell["cell"]
+                        self._cell = hdf_cell["cell"]
```
jan-janssen: I guess for consistency we should always use `set_cell` instead of `self._cell`.

pyiron/atomistics/structure/atoms.py
```diff
@@ -693,7 +694,9 @@ def center(self, vacuum=None, axis=(0, 1, 2)):
         translation = np.zeros(3)
         for i in axes:
             nowlen = np.sqrt(np.dot(c[i], c[i]))
-            self.cell[i] *= 1 + longer[i] / nowlen
+            cell = self.cell
```
jan-janssen: or `cell = self.cell.copy()` as cell is an numpy array you still have a link when you use `cell = self.cell`

tests/atomistics/structure/test_atoms.py
```diff
@@ -299,21 +299,23 @@ def test_cell(self):
         )
         self.assertTrue((CO.get_cell() == np.identity(3)).all())
         self.assertTrue((CO.cell == np.identity(3)).all())
-        CO.cell[2][2] = 10.0
-        self.assertTrue(CO.cell[2, 2] == 10.0)
+        cell = CO.cell
```
jan-janssen: Again `cell = CO.cell.copy()`

jan-janssen:
If the tests finish without errors this can be merged.

PR #543, opened 2020-03-09
Warning status

liamhuber:
Add a new status 'warning' to jobs, and use it when Vasp faces EDDRMM issues.

### DIFF
--- a/pyiron/vasp/base.py
+++ b/pyiron/vasp/base.py
@@ -471,7 +471,7 @@ class VaspBase(GenericDFTJob):
         warning_string = "EDDRMM warnings occured {} times, first in ionic step {}."
         status_string = "Status is switched to 'warning'."
 
-        if not snap is None:
+        if snap is not None:
             if self.get_eddrmm_handling() == "ignore":
                 self._logger.warning(warning_string.format(num_eddrmm, snap))
             elif self.get_eddrmm_handling() == "warn":
--- a/pyiron/vasp/base.py
+++ b/pyiron/vasp/base.py
@@ -1760,7 +1760,15 @@ class Input:
             if "vasp_dict" in hdf5_input.list_nodes():
                 vasp_dict = hdf5_input["vasp_dict"]
                 if "eddrmm_handling" in vasp_dict.keys():
-                    self._eddrmm = vasp_dict["eddrmm_handling"]
+                    self._eddrmm = self._eddrmm_backwards_compatibility(vasp_dict["eddrmm_handling"])
+
+    @staticmethod
+    def _eddrmm_backwards_compatibility(eddrmm_value):
+        """On 9-03-2020, the EDDRMM flag 'not_converged' was switched to 'warn'."""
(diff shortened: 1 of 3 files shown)


pyiron/vasp/base.py
```diff
@@ -468,24 +468,18 @@ def collect_errors(self):
         """
         num_eddrmm, snap = self._get_eddrmm_info()
 
+        warning_string = "EDDRMM warnings are ignored. EDDRMM occures {} times, first in ionic step {}."
```
FeLoch: shall be "warnings are ignored" be always printed? Even if  `self.get_eddrmm_handling() == "warn"`?

pyiron/vasp/base.py
```diff
@@ -777,14 +771,14 @@ def write_magmoms(self):
...
             s.logger.debug("No magnetic moments")
 
-    def set_eddrmm_handling(self, status="not_converged"):
+    def set_eddrmm_handling(self, status="warn"):
         """
         Sets the way, how EDDRMM warning is handled.
 
         Args:
-            status (str): new status of EDDRMM handling (can be 'not_converged', 'ignore', or 'restart')
+            status (str): new status of EDDRMM handling (can be 'warning', 'ignore', or 'restart')
```
FeLoch: should it be `"warn"` instead of `"warning"`?

pyiron/vasp/base.py
liamhuber: Fair. Now it just says "EDDRMM warnings occured {} times, first in ionic step {}"

pyiron/vasp/base.py
liamhuber: Yes!

liamhuber:
> Nice work! You may also change the default here:
> 
> https://github.com/pyiron/pyiron/blob/ecf64369dde74305369146dcd2883a346277cf05/pyiron/vasp/base.py#L1706
>
>.

Thanks!

> Also backward compatibility should be conserved in here:
> 
> https://github.com/pyiron/pyiron/blob/ecf64369dde74305369146dcd2883a346277cf05/pyiron/vasp/base.py#L1753
> 
> .

I'm not sure what you mean backwards compatibility in the --- aha, yes, right, got it. Updated.

liamhuber:
@FeLoch No no, you were sufficiently clear, I just had a temporary mental blockage.

liamhuber:
@jan-janssen the Travis error does not seem related to my changes, could you verify please?

jan-janssen:
I restarted the build. You can do the same using the pyiron-runner account and login in on Travis.

liamhuber:
> I restarted the build. You can do the same using the pyiron-runner account and login in on Travis.

Cool, thanks! What exactly is the signal I'm looking for that this the action I ought to take? Is any 'suspicious failure' sufficient? I assume this is not a terribly expensive action.

jan-janssen:
Whenever it says timeout then usually it makes sense to simply restart the tests.

PR #546, opened 2020-03-10
resolve links in project path

pmrv:
pyiron rejects absolute project paths with symlinks in them, even if they point into the pyiron project folder.  Came up for me because of the /u -> /cmmc/u link on the cluster and I was doing some stuff on the command line.

### DIFF
--- a/pyiron/base/database/filetable.py
+++ b/pyiron/base/database/filetable.py
@@ -32,7 +32,7 @@ class FileTable(with_metaclass(Singleton)):
     def __init__(self, project):
         self._fileindex = None
         self._job_table = None
-        self._project = os.path.abspath(project)
+        self._project = os.path.realpath(project)
         self._columns = ['id', 'status', 'chemicalformula', 'job', 'subjob', 'projectpath', 'project', 'timestart',
                          'timestop', 'totalcputime', 'computer', 'hamilton', 'hamversion', 'parentid', 'masterid',
                          'username']


pmrv:
My idea was just to normalize the path of each project as soon as possible to always refer to its true path before comparing to the globally configured path.  If both the configured path at 
https://github.com/pyiron/pyiron/blob/master/pyiron/base/settings/generic.py#L561
and the path of each project go through `os.path.realpath()`, then symlinks and true paths should be recognized as the same.

In the change above I missed the call to `super` in [Project](https://github.com/pyiron/pyiron/blob/master/pyiron/base/project/generic.py#L124) and then only changed the other location in the code where the path is used.  Looking more closely I guess the correct location for the change would be in [ProjectPath](https://github.com/pyiron/pyiron/blob/14eb1a54004a08c7b318f8b9ddc3dbbb6a54bb31/pyiron/base/project/path.py#L228)?

jan-janssen:
No in https://github.com/pyiron/pyiron/blob/14eb1a54004a08c7b318f8b9ddc3dbbb6a54bb31/pyiron/base/project/path.py#L405 the Settings object is called, so we should try to move all path manipulations to one part. I am not sure if the Settings object is the right place though. What we need is a unique way to match a path to the current path to paths in the config file and this is how this part ended up in the Settings object.

jan-janssen:
Even more we just changed this part to be compatible with other clusters https://github.com/pyiron/pyiron/pull/416

pmrv:
Ah, was not aware of #416. Well, if it's as complicated as it is, no need to change everything around. I can just resolve it in my script.

Sorry, for the itchy trigger finger. ;)

PR #552, opened 2020-03-11
Vasp constrained dynamics

sudarsan-surendralal:
A separate module derived from `Vasp` is created to run constrained MD, slow growth MD, and metadynamics in VASP.


pyiron/base/generic/parameters.py
```diff
@@ -980,3 +980,11 @@ def _find_line(self, key_name):
                                                               self._dataset["Value"][i]))
             error_msg = "\n".join(error_msg)
             raise ValueError(error_msg)
+
+    def clear_all(self):
```
jan-janssen, line 984: Can not you use `load_default()` ?

sudarsan-surendralal: In this case yes. But what if `load_default()` does not give an empty dataset. In that case, wouldn't `clear_all` be useful?

jan-janssen: You are right.

jan-janssen:
@sudarsan-surendralal you can merge it now.

sudarsan-surendralal:
Cool! Let me quickly add a little testing!

PR #559, opened 2020-03-20
Spx remove pipes

samwaseda:
To make sure that compression can be performed properly


jan-janssen:
Looks good to me - did you test it?

samwaseda:
Yes, and it doesn't work XD

samwaseda:
This commit was actually not necessary, because when the files are compressed, the pipes are removed automatically.

PR #565, opened 2020-03-23
Changes from Yury

jan-janssen:


pyiron/dft/job/generic.py
```diff
@@ -273,3 +267,42 @@ def set_empty_states(self, n_empty_states=None):
...
+    Args:
+        cell:
+        kpoints_per_angstrom:
+    Returns:
+    """
+    omega = np.linalg.det(cell)
+    l1, l2, l3 = cell
+    g1 = 2 * np.pi / omega * np.cross(l2, l3)
+    g2 = 2 * np.pi / omega * np.cross(l3, l1)
+    g3 = 2 * np.pi / omega * np.cross(l1, l2)
```
samwaseda: Here's my problem: I do understand that 2pi has a physical origin, but in reality the k-mesh is solely determined empirically, i.e. people remember the numbers like 10, 12, 24 etc. I find it therefore confusing that there's a factor 2pi.

pyiron/dft/job/generic.py
```diff
@@ -273,3 +267,42 @@ def set_empty_states(self, n_empty_states=None):
...
+
+    kmesh = np.rint(
+        np.array([np.linalg.norm(g) for g in [g1, g2, g3]]) * kpoints_per_angstrom
+    )
+    if kmesh.min() <= 0:
+        warnings.warn("kpoint per angstrom too low")
+    return [int(k) for k in kmesh]
+
+def get_k_mesh_by_density(cell, kmesh_density_per_inverse_angstrom=1.0):
```
samwaseda: I'm not really sure if it's a good idea to offer both `get_k_mesh_by_cell` and `get_k_mesh_by_density`, because the difference is not very clear at first glance. I guess in the end it's easier for the people to remember that they have to put either a number or its inverse. Also I don't think it's rather unlikely that people need both of them.

pyiron/dft/job/generic.py
```diff
@@ -78,13 +78,12 @@ def exchange_correlation_functional(self, val):
...
-        latlens = [np.linalg.norm(lat) for lat in cell]
-        kmesh = np.rint(
-            np.array([2 * np.pi / ll for ll in latlens]) * kpoints_per_angstrom
-        )
-        if kmesh.min() <= 0:
-            warnings.warn("kpoint per angstrom too low")
-        return [int(k) for k in kmesh]
+        return get_k_mesh_by_cell(cell, kpoints_per_angstrom)
+
+    def get_k_mesh_by_density(self, cell=None, kmesh_density_per_inverse_angstrom=1):
```
samwaseda: In most of the cases, what we need is the k mesh of the current cell, so it's not really intuitive to me that I have to put cell before the k mesh density. I'd rather change the order and remove the mysterious `1` in the argument.

pyiron/dft/job/generic.py
yury-lysogorskiy: I use 2*pi factor to be consistent with my previous "intuition" from  Materials Design's MedeA  software, where they take into account factor  2*pi when kmesh was setup from the density.
So, the 0.5 is very rough setting, whereas 0.15 or 0.1 is tight "enough"

pyiron/dft/job/generic.py
yury-lysogorskiy: Agree, there is a bit of confusion.
I like @liamhuber  "kpoints_per_reciprocal_angstrom" notation.

pyiron/dft/job/generic.py
samwaseda: Ok at MPIE we usually use integers, but if there's another kind of background, I give in.

jan-janssen:
@yury-lysogorskiy, @samwaseda and @sudarsan-surendralal Do you plan to merge these changes? Or should we close the pull request for now?

pyiron/dft/job/generic.py
samwaseda: > In most of the cases, what we need is the k mesh of the current cell, so it's not really intuitive to me that I have to put cell before the k mesh density. I'd rather change the order and remove the mysterious `1` in the argument.

I'm mainly waiting for this one. Other than that, I'm fine with merging it

pyiron/vasp/vasp.py
```diff
@@ -54,3 +54,27 @@ def __init__(self, project, job_name):
...
+        """
+
+        Args:
+            structure:
+
+        Returns:
+
+        """
+        VaspInteractive.structure.fset(self,structure)
+        if structure is not None:
```
jan-janssen, line 78: This part should directly be implemented in the `VaspInteractive` class

Issue #540, opened 2020-03-09
Multi Word Keys in GenericParameters


pmrv:
Right now keys into GenericParameters cannot have white space in them and instead have to have underscores `'___'` instead, which is a bit inconvenient and unintuitive. 

Our consensus from Friday was to change `GenericParameters.__getitem__` to
special case keys with white space and replace it with
`GenericParameters.multi_word_separator = '___'` transparently.  Looking into
it there are three methods involved in this

1. [`__getitem__`](https://github.com/pyiron/pyiron/blob/master/pyiron/base/generic/parameters.py#L652)
2. [`get`](https://github.com/pyiron/pyiron/blob/master/pyiron/base/generic/parameters.py#L357)
3. [`_find_line`](https://github.com/pyiron/pyiron/blob/master/pyiron/base/generic/parameters.py#L962)

and the only reason special casing white space is necessary in the first place
is that `_find_line` splits keys on white space to make handling *values* with
white space a bit simpler.  I think the intended use case is something like
this (it's broken in `__getitem__` because of a small bug)
```python
>>> p = GenericParameters()
>>> p['fix'] = '2 3 foo bar'
>>> p['fix']
'2 3 foo bar'
>>> p.get('fix 2')
'3 foo bar'
>>> p.get('fix 2 3')
'foo bar'
>>> p.get('fix 3')
```
Something similar for `__setitem__` is not implemented.  If we replace white
space with `'___'` in `__getitem___` or `get`, then this is broken, because
`_find_line` will never see keys with white space.  OTOH, I don't think this
functionality is worth having to manually do the replacement and a superficial
grep on our internal notebooks (with `"input(.control)?\[['\"].*['\"]\]"`)
seems to show that is never used anyway, but I don't know about your personal
or external code.  Still I would rather then remove the code that enables this
behavior and just allow keys with white space in the underlying
`OrderedDict()`.  With this the use case e.g. for multiple fixes in lammps
would be
```python
>>> p = GenericParameters()
>>> p['fix ensemble'] = 'nvt'
>>> p['fix 2'] = 'foo bar'
>>> p
 Parameter      Value   Comment
1  foo ensemble nvt
2  foo 2        foo bar
>>> p.get_string_lst()
['fix ensemble nvt\n', 'fix 2 foo bar\n']
```
The code in `get_string_lst` to write the input files would handle this
already, if it weren't the case now that `_find_list` is (more or less) broken
on keys with white space.  We can also continue to special case keys with
[...]

PR #564, opened 2020-03-23
Generic parameters no multi word

jan-janssen:

PR #567, opened 2020-03-24
add methods to get charges/ids for pair_style full

pmrv:
- updated LammpsStructure to use the new methods when writing structures
  with pair_style full
- this removes the need for special casing multi word keys in
  GenericParameters

See #540 

PS: I managed to get myself confused between my fork and this repo. Sorry if you saw this popping up multiple times.


jan-janssen:
@pmrv I have to admit, I lost track of this issue here. As far as I understand it is an updated / more cleaned up version of https://github.com/pyiron/pyiron/pull/564 - so maybe it is reasonable to merge the current master branch into this pull request once before merging the changes. 

@sudarsan-surendralal Did you test this with one of your Water systems? I guess the water MD example should be sufficient as an integration test https://github.com/pyiron/pyiron/blob/master/notebooks/water_MD.ipynb

pmrv:
So did I honestly.  I think the idea was that with this the multi key word stuff would no longer be needed, but now I'm not sure anymore.  Will update to current master anyway and investigate.

jan-janssen:
@sudarsan-surendralal Can you test if these changes work with your water examples? Then we can finally merge this pull request.

pmrv:
I just checked the water notebook.  There are no errors and visually it seems ok.

pyiron/lammps/structure.py
```diff
@@ -402,10 +402,10 @@ def structure_full(self):
...
         # extract electric charges from potential file
         q_dict = {}
-        for el in self._structure.get_species_symbols():
-            q_dict[el] = float(self.potential.get("set group {} charge".format(el)))
-        species_translate_list = [self.potential.get("group {} type".format(el.Abbreviation))
-                                  for el in self.structure.species]
+        species_translate_list = []
+        for el in self.structure.get_species_symbols():
+            q_dict[el] = self.potential.get_charge(el)
+            species_translate_list.append(self.potential.get_element_id(el))
```
sudarsan-surendralal, line 408: For these lines, you need to change them to

```
        species_translate_list = []
        for el in self.structure.species:
            q_dict[el.Abbreviation] = self.potential.get_charge(el.Abbreviation)
            species_translate_list.append(self.potential.get_element_id(el.Abbreviation))
```
This is because, `self.structure.get_species_symbols()` and `self.structure.species` do not return the species in the same order. `species_translate_list` should follow the order of `self.structure.species`

PR #568, opened 2020-03-24
ase fix

jan-janssen:


jan-janssen:
@samwaseda Fixing the bug with the old mixing example

PR #570, opened 2020-03-24
file_name is not taken over by the child job anymore and path is not …

samwaseda:
…encapsulated

### DIFF
--- a/pyiron/base/master/generic.py
+++ b/pyiron/base/master/generic.py
@@ -531,10 +531,7 @@ class GenericMaster(GenericJob):
             parent_job:
             child_job:
         """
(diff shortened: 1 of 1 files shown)


samwaseda:
ok I see that for some reason `job_table` does not get updated anymore when the job is submitted.

samwaseda:
Which line

samwaseda:
Anyway I don't understand what `child_job.project_hdf5.file_name = parent_job.project_hdf5.file_name` is for

jan-janssen:
If you do not want this functionality, maybe it is easier to simply overload the function in your class rather than changing it in this class, as this class is used by many other classes.

samwaseda:
No, I close this one

PR #572, opened 2020-03-30
Rewrite line mode for band structure calculations

FeLoch:
Using the tool [seekpath](https://github.com/giovannipizzi/seekpath), high symmetry points and paths are created automatically.

The paths are now stored as list of tuples including start and end points, thus discontinuous paths are possible to calculate. E.g. `[("G", "H"), ("H", "N"), ("G", "P")]`.

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1196,7 +1196,7 @@ class Atoms(object):
         pbc = self.pbc
         cell = sp_dict["primitive_lattice"]
 
-        struc_new = Atoms(elements=element_list, positions=positions, pbc=pbc, cell=cell)
+        struc_new = Atoms(elements=element_list, scaled_positions=positions, pbc=pbc, cell=cell)
 
         struc_new._set_high_symmetry_points(sp_dict["point_coords"])
         struc_new._set_high_symmetry_path({"standard": sp_dict["path"]})
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1199,7 +1199,7 @@ class Atoms(object):
         struc_new = Atoms(elements=element_list, scaled_positions=positions, pbc=pbc, cell=cell)
 
         struc_new._set_high_symmetry_points(sp_dict["point_coords"])
-        struc_new._set_high_symmetry_path({"standard": sp_dict["path"]})
+        struc_new._set_high_symmetry_path({"full": sp_dict["path"]})
 
         return struc_new
 
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -13,6 +13,7 @@ import warnings
 from ase.geometry import cellpar_to_cell, complete_cell, get_distances
 from matplotlib.colors import rgb2hex
 from scipy.interpolate import interp1d
+import seekpath
 
 from pyiron.atomistics.structure.atom import Atom
 from pyiron.atomistics.structure.sparse_list import SparseArray, SparseList
@@ -94,7 +95,6 @@ class Atoms(object):
         elements=None,
         dimension=None,
         species=None,
-        high_symmetry_points=None,
         **qwargs
     ):
         if symbols is not None:
@@ -229,8 +229,7 @@ class Atoms(object):
                 self.pbc = pbc
         self.set_initial_magnetic_moments(magmoms)
         self._high_symmetry_points = None
-        if high_symmetry_points is not None:
-            self.set_high_symmetry_points(high_symmetry_points)
+        self._high_symmetry_path = None
 
 
     @property
@@ -320,7 +319,7 @@ class Atoms(object):
         """
         return self._high_symmetry_points
 
-    def set_high_symmetry_points(self, new_high_symmetry_points):
+    def _set_high_symmetry_points(self, new_high_symmetry_points):
         """
         Sets new high symmetry points dictionary.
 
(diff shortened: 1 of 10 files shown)


pyiron/atomistics/structure/atoms.py
```diff
@@ -331,6 +330,51 @@ def set_high_symmetry_points(self, new_high_symmetry_points):
...
+        Adds a new path to the dictionary of pathes for band structure calculations.
+
+        Args:
+            path (dict): dictionary of lists of tuples with start and end point.
+                E.G. {"my_path": [('Gamma', 'X'), ('X', 'Y')]}
+        """
+        if self.get_high_symmetry_path() is None:
+            raise AssertionError("Construct high symmetry path first. Use self.band_structure_calculations().")
+        else:
+            self._high_symmetry_path.update(path)
```
samwaseda: It would be helpful if the function also checks the content of the dictionary (especially because "dictionary of lists of tuples is quite complex), e.g.

```
for value in path.values():
    for index in range(len(value)-1):
        if value[index][1] != value[index+1][0]:
            raise ValueError('Path not connecting at {} and {}'.format(' '.join(value[index]), ' '.join(value[index+1]))
```

pyiron/atomistics/structure/atoms.py
```diff
@@ -1103,6 +1154,55 @@ def center_coordinates_in_unit_cell(self, origin=0, eps=1e-4):
             )
         return self
 
+    def band_structure_calculation(self,
```
samwaseda: I guess it's not easy to understand `band_structure_calculation` is a function that activates a band structure calculation. Maybe something like `set_band_structure_calculation` or `activate_band_structure_calculation`. To quote Liam: functions should start with a verb.

pyiron/atomistics/structure/atoms.py
```diff
@@ -331,6 +330,51 @@ def set_high_symmetry_points(self, new_high_symmetry_points):
...
+    def add_high_symmetry_path(self, path):
+        """
+        Adds a new path to the dictionary of pathes for band structure calculations.
+
+        Args:
+            path (dict): dictionary of lists of tuples with start and end point.
+                E.G. {"my_path": [('Gamma', 'X'), ('X', 'Y')]}
+        """
+        if self.get_high_symmetry_path() is None:
+            raise AssertionError("Construct high symmetry path first. Use self.band_structure_calculations().")
```
samwaseda: `self.band_structure_calculation` and not `self.band_structure_calculations`

pyiron/atomistics/structure/atoms.py
```diff
@@ -331,6 +330,51 @@ def set_high_symmetry_points(self, new_high_symmetry_points):
...
 
+    def add_high_symmetry_points(self, new_points):
+        """
+        Adds new points to the dict of existing high symmetry points.
+
+        Args:
+            new_points (dict): Points to add
+        """
+        if self.get_high_symmetry_points() is None:
+            raise AssertionError("Construct high symmetry points first. Use self.band_structure_calculations().")
```
samwaseda: `self.band_structure_calculation` and not `self.band_structure_calculations`

pyiron/atomistics/structure/atoms.py
FeLoch: Nice idea, but the path do not need to be connected. A path `G->H->N, G->P` is also valid. I will ad a check, whether the given points of the path are also valid keys of the high symmetry points.

pyiron/atomistics/structure/atoms.py
FeLoch: Sure, I also thought about the name. `set` sounds like setting a variable. I would suggest `create_structure_for_line_mode` or `create_line_mode_structrue`.

samwaseda: I just realized that it's in `atoms.py`, but shouldn't it rather be inside `GenericDFTJob`? I mean in `pyiron.dft.job.generic`

FeLoch: I would keep it in `atoms.py`. Then the user can create a structure and check the symmetries and paths before creating a job.

pyiron/atomistics/structure/atoms.py
FeLoch: done

pyiron/atomistics/structure/atoms.py
FeLoch: done

pyiron/atomistics/structure/atoms.py
FeLoch: new function name

pyiron/atomistics/structure/atoms.py
FeLoch: done

pyiron/atomistics/structure/atoms.py
```diff
@@ -13,6 +13,7 @@
 from ase.geometry import cellpar_to_cell, complete_cell, get_distances
 from matplotlib.colors import rgb2hex
 from scipy.interpolate import interp1d
+import seekpath
```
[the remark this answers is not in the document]
FeLoch: Yes, me neither...

pyiron/atomistics/structure/atoms.py
```diff
@@ -1103,6 +1164,55 @@ def center_coordinates_in_unit_cell(self, origin=0, eps=1e-4):
...
 
+    def create_line_mode_structure(self,
+                                   with_time_reversal=True,
+                                   recipe='hpkot',
+                                   threshold=1e-07,
+                                   symprec=1e-05,
+                                   angle_tolerance=-1.0,
+                                   ):
+        """
+        Creates a new structure!
```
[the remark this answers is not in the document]
FeLoch: sure! I wrote the line, because in `self.repeat` is a simmilar docstring.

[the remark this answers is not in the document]
FeLoch: done

pyiron/vasp/base.py
```diff
@@ -2225,51 +2230,29 @@ def __init__(self, input_file_name=None, table_name="kpoints"):
...
-        """
-        Returns high symmetry points name of k-points trace (Line mode only)
-
-        Returns:
-            list: trace values
-        """
-        return self._trace
-
-    def set(self, method=None, size_of_mesh=None, shift=None, n_trace=None, trace_coord=None):
+    def set(self, method=None, size_of_mesh=None, shift=None, n_path=None, path=None):
```
sudarsan-surendralal: I think we need to rename this function in the `Kpoints` class since it overwrites the `set` function of the base class with different input parameters. This is why codeacy tests complain. We should probably call it `set_kpoints_file` or something. Sorry, I should have done this much earlier but I never got to it!

sudarsan-surendralal: I think we need to rename this function in the `Kpoints` class since it overwrites the `set` function of the base class with different input parameters. This is why codeacy tests complain. We should probably call it `set_kpoints_file` or something. Sorry, I should have done this much earlier but I never got to it!

FeLoch: Sure, it is done.

jan-janssen:
Maybe I missed it - @FeLoch / @samwaseda  Does this mean we are already able to calculate band structures with Sphinx? Then we should definitely add an example to: https://github.com/pyiron/pyiron/tree/master/notebooks
I briefly discussed this with Christoph, and he suggested to use `nSloppy` to add additional bands but do not converge them. The corresponding Sphinx input files are attached 
[input-band.sx.txt](https://github.com/pyiron/pyiron/files/4414067/input-band.sx.txt)
[input-rho.sx.txt](https://github.com/pyiron/pyiron/files/4414068/input-rho.sx.txt)

FeLoch:
@jan-janssen Yes we are. But i didn't know about `nSloppy`! Let me try it!

FeLoch:
@jan-janssen I added the missing parameter in `sphinx/base.py`. @samwaseda could you have a look?

FeLoch:
@jan-janssen Could you help me with the `travis `error? I don't get it...

jan-janssen:
> @jan-janssen Could you help me with the `travis `error? I don't get it...

It seems like you already fixed it.

jan-janssen:
The only request I have, is to add a notebook how to use this need feature with Sphinx. Then we can use this for demonstrations but that is also fine to be added later.

FeLoch:
Ok great! If you say, the `travis` error is acceptable, I will merge it.

jan-janssen:
@FeLoch Yes, you can. 

If you click on the travis icon and it fails on the docker level, then that is usually related to some kind of time out, in particular when the other travis push runs through.

FeLoch:
Ah ok! Didn't know that!

PR #573, opened 2020-04-03
restart function for band structures

FeLoch:
new function for creating restart job for band structure calculations.

### DIFF
--- a/pyiron/dft/job/generic.py
+++ b/pyiron/dft/job/generic.py
@@ -143,6 +143,11 @@ class GenericDFTJob(AtomisticGenericJob):
     #    self.set_kpoints(mesh=mesh, scheme='MP', center_shift=None, symmetry_reduction=True, manual_kpoints=None,
     #                     weights=None, reciprocal=True)
 
+    def restart_for_band_structure_calculations(self, job_name=None):
+        raise NotImplementedError(
+            "restart_for_band_structure_calculations is not implemented for this code."
+        )
+
     def _set_kpoints(
         self,
         mesh=None,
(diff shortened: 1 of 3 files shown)


jan-janssen:
I am a bit confused. I guess the idea was that both `restart_for_band_structure_calculations` should have exactly the same input parameters, just like `calc_md`. If that is the case, then we should also add a function template in https://github.com/pyiron/pyiron/blob/master/pyiron/dft/job/generic.py

FeLoch:
Should now be done!

jan-janssen:
Can you also add a function template to generic DFT ? 
https://github.com/pyiron/pyiron/blob/master/pyiron/dft/job/generic.py#L130

FeLoch:
> 
> 
> Can you also add a function template to generic DFT ?
> https://github.com/pyiron/pyiron/blob/master/pyiron/dft/job/generic.py#L130

done

pyiron/dft/job/generic.py
```diff
@@ -143,6 +143,11 @@ def set_mixing_parameters(
     #    self.set_kpoints(mesh=mesh, scheme='MP', center_shift=None, symmetry_reduction=True, manual_kpoints=None,
     #                     weights=None, reciprocal=True)
 
+    def restart_for_band_structure_calculations(self, job_name=None):
```
jan-janssen, line 146: Can you copy the Docstring as well?

FeLoch: done

jan-janssen:
Looks good to me.

PR #578, opened 2020-04-07
Spx interactive parser

samwaseda:
This was needed since `interactive_close` is not called when the job does not finish within the cluster time, in order to transfer tags inside `output/interactive` to `output/generic`

### DIFF
--- a/pyiron/sphinx/interactive.py
+++ b/pyiron/sphinx/interactive.py
@@ -205,7 +205,7 @@ class SphinxInteractive(SphinxBase, GenericInteractive):
     def _interactive_to_output(self):
         with self.project_hdf5.open("output") as h5:
             if "interactive" in h5.list_groups():
-                for key in ["positions", "cells"]:
+                for key in ["positions", "cells", "indices", "cells"]:
                     h5["generic/" + key] = h5["interactive/" + key]
                 with self.project_hdf5.open("input") as hdf5_input:
                     with hdf5_input.open("generic") as hdf5_generic:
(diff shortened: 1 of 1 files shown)


pyiron/sphinx/interactive.py
```diff
@@ -202,6 +202,28 @@ def interactive_initialize_interface(self):
...
+                        with hdf5_generic.open("dft") as hdf5_dft:
+                            if (
+                                "atom_spin_constraints"
+                                in h5["interactive"].list_nodes()
+                            ):
+                                hdf5_dft["atom_spin_constraints"] = h5[
+                                    "interactive/atom_spin_constraints"
+                                ]
+
+    def collect_output(self, force_update=False):
```
samwaseda, line 223: @jan-janssen I was not sure if I'm allowed to overwrite `collect_output` this way. If `status.interactive` is set, will this function be called?

samwaseda: I mean, there's `collect_output` in `base.py` as well, but if it's interactive, I would like to call `_output_interactive_to_generic` as well (and therefore `collect_output` in `interactive.py`), but I don't know whether the distinction between `base.py` and `interactive.py` is done only by `status.interactive` or there's something more than that.

jan-janssen: @samwaseda Does it work? Or is this function also called for non interactive jobs? That would be my expectation.

jan-janssen: Maybe it should be: 

```
def collect_output(self, force_update=False):
    super(SphinxInteractive, self).collect_output(force_update=force_update)
    if self.server.run_mode.interactive == True:
        self._output_interactive_to_generic()
```

PR #579, opened 2020-04-08
Rewrite restart_from_... functions

FeLoch:
Just cleaning up a little bit. Since both functions `restart_from_wave_and_charge` and `restart_from_charge_density` used different code to get the correct `ICHARG` tag, I wrote a private function `_get_icharg_value` used for both.

### DIFF
--- a/pyiron/vasp/base.py
+++ b/pyiron/vasp/base.py
@@ -1507,7 +1507,7 @@ class VaspBase(GenericDFTJob):
         job_name=None,
         job_type=None,
         icharg=None,
-        self_consistent_calc=False,
+        self_consistent_calc=None,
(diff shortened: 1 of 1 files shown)


pyiron/vasp/base.py
```diff
@@ -1420,6 +1420,33 @@ def restart_for_band_structure_calculations(self, job_name=None):
             self_consistent_calc=None
         )
 
+    def _get_icharg_value(self, icharg=None, self_consistent_calc=None):
```
sudarsan-surendralal, line 1423: I like this function but I think it can be made public.

FeLoch: okay 😄

Issue #586, opened 2020-04-10
Incorrect displacements when cell shape varies


sudarsan-surendralal:
When calculating the displacements between the snapshots in a job, the current algorithm calculates the relative positions using the initial structure. 

https://github.com/pyiron/pyiron/blob/3bd1912c4417c637e8fc8f12058b60d03d4565e9/pyiron/atomistics/job/atomistic.py#L881-L883

But if the cell shape changes (eg. NpT simulations) during the simulation, the calculated displacements would be wrong.

PR #587, opened 2020-04-10
Fix displacements

sudarsan-surendralal:
To fix the issue mentioned here https://github.com/pyiron/pyiron/issues/586


samwaseda:
So, this was something I thought about when I first implemented the function, but then there was a practical question: How do you define a metric when the box size changes. Here's the thought experiment: Imagine a box of length 10 and there are an atom at position 0 and an atom at position 9 (let's say it's 1D) at the beginning. Now at the second step the box length changes to 11 and the atoms are at 0 and 10. How much did the atoms move?

sudarsan-surendralal:
> 
> 
> So, this was something I thought about when I first implemented the function, but then there was a practical question: How do you define a metric when the box size changes. Here's the thought experiment: Imagine a box of length 10 and there are an atom at position 0 and an atom at position 9 (let's say it's 1D) at the beginning. Now at the second step the box length changes to 11 and the atoms are at 0 and 10. How much did the atoms move?

The displacement of the second atom is 1 unit regardless of how the box changes. However, let's say atom 2 moves forward by 3 units and the cell value changes to 11. Due to PBC, the position value of this atom is wrapped to 1. However, if we use the original cell size of 10 units to "unwrap" the coordinates, the displacement across the cell boundary is wrongly calculated to be 2 units rather than 3. This is what I think.

samwaseda:
> The displacement of the second atom is 1 unit regardless of how the box changes.

I disagree. If there are only two atoms in the box in this case, the distance between the atoms does not change. We can also put it this way (assuming the box length changes from 10 to 11):

Scenario 1:
1st atom goes from 0 to 0
2nd atom goes from 9 to 10

Scenario 2:
1st atom goes from 0 to 0
2nd atom goes 1 to 1

In both cases the distance between the two atoms does not change, but in the first scenario there's a displacement for the second atom.

samwaseda:
> However, let's say atom 2 moves forward by 3 units and the cell value changes to 11. Due to PBC, the position value of this atom is wrapped to 1. However, if we use the original cell size of 10 units to "unwrap" the coordinates, the displacement across the cell boundary is wrongly calculated to be 2 units rather than 3. This is what I think.

Yes, what you're saying is correct. For this reason, originally I thought about forbidding the usage of this function altogether when the cell size changes. But then I thought maybe it's better to leave it there, because at least the user will have some idea about the displacements.

pyiron/atomistics/job/atomistic.py
```diff
@@ -876,18 +876,37 @@ def displacements(self):
...
         displacement -= np.append(
-            self._job.structure.get_scaled_positions(), displacement
-        ).reshape(len(self.positions) + 1, len(self._job.structure), 3)[:-1]
+            structure.get_scaled_positions(), displacement
+        ).reshape(len(positions) + 1, len(structure), 3)[:-1]
         displacement -= np.rint(displacement)
-        displacement = np.tensordot(
-            displacement, self._job.structure.cell, axes=([2, 0])
-        )
+        displacement = np.array([np.tensordot(pos, cell, axes=([1, 1])) for pos, cell in zip(displacement, cells)])
```
samwaseda, line 909: Isn't this the same as `np.einsum('ni,nji->nj', displacements, cells)`?

sudarsan-surendralal:
> 
> 
> > The displacement of the second atom is 1 unit regardless of how the box changes.
> 
> I disagree. If there are only two atoms in the box in this case, the distance between the atoms does not change. We can also put it this way (assuming the box length changes from 10 to 11):
> 
> Scenario 1:
> 1st atom goes from 0 to 0
> 2nd atom goes from 9 to 10
> 
> Scenario 2:
> 1st atom goes from 0 to 0
> 2nd atom goes 1 to 1
> 
> In both cases the distance between the two atoms does not change, but in the first scenario there's a displacement for the second atom.

But isn't the displacements in this pyiron function referring to the change in position of an atom wrt. the previous snapshot and not wrt. the other atoms?

samwaseda:
> But isn't the displacements in this pyiron function referring to the change in position of an atom wrt. the previous snapshot and not wrt. the other atoms?

Yes, and what's the change in position of an atom wrt. the previous snapshot if the box size changes? The one that you're claiming is just one possibility to calculate it and I'm saying it does not represent physics.

pyiron/atomistics/job/atomistic.py
sudarsan-surendralal: Yes. I've made the change

samwaseda:
> > But isn't the displacements in this pyiron function referring to the change in position of an atom wrt. the previous snapshot and not wrt. the other atoms?
> 
> Yes, and what's the change in position of an atom wrt. the previous snapshot if the box size changes? The one that you're claiming is just one possibility to calculate it and I'm saying it does not necessarily represent physics.

I guess this very much depends on whether the atoms are remapped when the box length changes. I assume they are remapped and if so, the displacements created through the remapping is not physically meaningful.

sudarsan-surendralal:
> 
> 
> > > But isn't the displacements in this pyiron function referring to the change in position of an atom wrt. the previous snapshot and not wrt. the other atoms?
> > 
> > 
> > Yes, and what's the change in position of an atom wrt. the previous snapshot if the box size changes? The one that you're claiming is just one possibility to calculate it and I'm saying it does not necessarily represent physics.

Why wouldn't this represent physics? The direct coordinates are consistent either way.  And we are working with positions that have been remapped at each snapshot based on the cell shape (which is what VASP and LAMMPS does). Am I missing something here?

pyiron/atomistics/job/atomistic.py
```diff
@@ -876,18 +876,37 @@ def displacements(self):
...
+            structure (pyiron.atomistics.structure.atoms.Atoms): The initial structure
+            positions (numpy.ndarray): List of positions in cartesian coordinates (N_steps x N_atoms x 3)
+            cells (numpy.cells): List of cells (N_steps x 3 x 3)
+
+        Returns:
+            numpy.ndarray: Displacements (N_steps x N_atoms x 3)
+
+        """
+        displacement = np.array(
+            [np.tensordot(pos, np.linalg.inv(cell), axes=([1, 1])) for pos, cell in zip(positions, cells)])
```
samwaseda, line 904: This is probably the same: `np.einsum('nki,nji->nkj', positions, cells)`

samwaseda:
Ok then let me give you another example: Let's say I have two atoms at 8.5 and 9.5 and my box length is 10 at the beginning. At the second step, the atoms are still at the same positions, but the box shrank to 9, which means the second atom is remapped to 0.5. What are their displacements? With your algorithm it'll be 0 for the first atom and 0.09 for the second atom, but is it really the physically meaningful value?

jan-janssen:
@sudarsan-surendralal and @samwaseda If I understand Sam's comment correctly the current implementation does not work. Therefore my suggestion would be to close this pull request.

sudarsan-surendralal:
> @sudarsan-surendralal and @samwaseda If I understand Sam's comment correctly the current implementation does not work. Therefore my suggestion would be to close this pull request.

Both the function we have now and the new changes are probably wrong IMO. There has to be a warning or a better way to do this. Can we keep this open?

samwaseda:
> > @sudarsan-surendralal and @samwaseda If I understand Sam's comment correctly the current implementation does not work. Therefore my suggestion would be to close this pull request.
> 
> Both the function we have now and the new changes are probably wrong IMO. There has to be a warning or a better way to do this. Can we keep this open?

I agree with @sudarsan-surendralal. I guess we just have to agree on one definition and make pyiron issue a warning when the cell size changes.

sudarsan-surendralal:
> 
> 
> > Both the function we have now and the new changes are probably wrong IMO. There has to be a warning or a better way to do this. Can we keep this open?
> 
> I agree with @sudarsan-surendralal. I guess we just have to agree on one definition and make pyiron issue a warning when the cell size changes.

I too think we should raise a warning. Shall we also bring this up during the pyiron meeting? Maybe we'll hear more opinions about this?

jan-janssen:
@sudarsan-surendralal Yes, I guess it is a great idea to discuss all open issues and pull requests in the next pyiron meeting https://github.com/pyiron/pyiron/issues/639 - most likely we can also use this in general for future pyiron meetings, so everybody who wants to present something can simply open up an issue.

samwaseda:
I'm okay with discussing it at the pyiron meeting, but I'm also okay if @sudarsan-surendralal  simply sets a definition and implements it as he likes, because I have the feeling that it's gonna be a bit time consuming to explain it or difficult to understand for many people what's the matter and in addition to this I expect that the conclusion will be the same.

sudarsan-surendralal:
Added a warning for varying cells! @samwaseda, it looks like there is no clear idea on how to handle this so this should be fine right?

PR #589, opened 2020-04-14
Reload atoms from HDF5 file

jan-janssen:
The code is still very much focused on the Atoms object but in general the idea is to allow any object to be reloaded from the HDF5 file, with the `to_object()` function:

```
from pyiron import Project
from pyiron.base.generic.hdfio import ProjectHDFio
pr = Project('test')
structure = pr.create_structure('Fe', 'bcc', 2.78)
hdf5_file = ProjectHDFio(project=pr, file_name='store.h5')
structure.to_hdf(hdf5_file)
hdf5_file['structure'].to_object()
```

Previously this was only possible for pyiron jobs.

### DIFF
--- a/pyiron/base/generic/hdfio.py
+++ b/pyiron/base/generic/hdfio.py
@@ -1285,7 +1285,13 @@ class ProjectHDFio(FileHDFio):
         new_obj = self.create_object(obj_type, **qwargs)
         if obj_type != str(type(new_obj)):  # Backwards compatibility
             self["TYPE"] = str(type(new_obj))
-        new_obj.from_hdf()
+        if obj_type != "<class 'pyiron.atomistics.structure.atoms.Atoms'>":
+            new_obj.from_hdf()
+        else:
+            new_obj.from_hdf(
+                hdf=self.open(".."),
+                group_name=self.h5_path.split('/')[-1]
+            )
         return new_obj
 
     def get_job_id(self, job_specifier):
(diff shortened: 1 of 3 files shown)


pyiron/base/generic/hdfio.py
```diff
@@ -1285,7 +1285,13 @@ def to_object(self, object_type=None, **qwargs):
         new_obj = self.create_object(obj_type, **qwargs)
         if obj_type != str(type(new_obj)):  # Backwards compatibility
             self["TYPE"] = str(type(new_obj))
-        new_obj.from_hdf()
+        if obj_type != "<class 'pyiron.atomistics.structure.atoms.Atoms'>":
```
sudarsan-surendralal, line 1288: If I understand correctly, for all objects except `Atoms`, we are loading it from the directory where `TYPE` is written right? And for `Atoms` we go one step back. I really like the way we do it for the `Atoms` class because it seems more intuitive. Should we make this consistent for all classes or are there some restrictions?

jan-janssen: Yes, in the mid-term we should make it more consistent. For now it is more like a quick fix to simplify loading atomistic structures.

samwaseda:
Is this still the standard way of storing structures in hdf?

jan-janssen:
By default the structure is stored automatically. So I am a bit confused about your question.

jan-janssen:
If you want to get a database ID for a structure you can use the structure container instead:
```
structure_container = pr.create_job(pr.job_type.StructureContainer, "structure_container")
structure_container.structure = structure
```

PR #590, opened 2020-04-15
Fix auto complete 

jan-janssen:


jan-janssen:
To set the correct behaviour we should use in IPython environment::

```  
%config Completer.use_jedi = False
```

or add to IPython config (``<HOME>\.ipython\profile_default\ipython_config.py``)::
     
```
c.Completer.use_jedi = False
```

jan-janssen:
Exactly - I tested it locally.

jan-janssen:
@sudarsan-surendralal The white space issue is fixed as well.

PR #592, opened 2020-04-16
New SPHInX input parsing scheme

ashtonmv:
As discussed, it would be powerful to control sphinx input groups more or less like dictionaries. My approach was to create a `class Group` inside of sphinx/base.py, and then instantiate each input group as a `Group`. Below is an example of how it can be used. I had to change several of the tests in test_base to reflect these changes. I hope I haven't removed anything crucial for reasons I couldn't foresee.

```python
import os
from pyiron import Project
```
```python
os.chdir("your/pyiron/directory")
pr = Project("sphinx_test")
job = pr.create_job(pr.job_type.Sphinx, "test")
```
when the job is created, all seven possible SPHInX groups (structure, species, basis, hamilton, guess, spin, and main) are instantiated but completely empty, so the user can build them from scratch or use something like `calc_minimize()` to load defaults, as below.
```python
# Just some possibilities
job.structure = pr.create_structure("Fe", "bcc", 2.78)
job.structure.add_tag(selective_dynamics=[True, True, True])
job.structure.selective_dynamics[1] = [False, False, True]
job.structure.set_initial_magnetic_moments([-4, 4])
```
```python
# Loads all default groups internally, as long
# as self.structure has been set
job.calc_minimize()
```
Where the actual parsing happens:
```python
# Modify structure group
job.input.structure.set_group("symmetry", {})

# Modify guess group
job.input.guess.rho.charged = {"charge": 2, "z": 20}
job.input.guess.rho.charged.charge = 3

# Modify hamilton group
job.input.hamilton.set_flag("dipoleCorrection")
job.input.hamilton.set("nExcessElectrons", -2)
job.input.hamilton.nExcessElectrons = -4
```
Dot notation works recursively and everything can be controlled via normal dict[key] = value setting.
```python
job.write_input()  # or run(), etc.
```


pyiron/sphinx/base.py
```diff
@@ -479,9 +704,7 @@ def from_hdf(self, hdf=None, group_name=None):
     def from_directory(self, directory):
         try:
             if not self.status.finished:
-                subprocess.call(
-                    "module load sphinx && sx2aims", cwd=directory, shell=True
-                )
+                os.system("module load sphinx && sx2aims")
```
jan-janssen, line 707: The module load part is specific to our cluster - I know it was in there before, but it should not be. The way we call executables is using the scripts in https://github.com/pyiron/pyiron-resources/tree/master/sphinx/bin - so maybe we could have a sx2aims script in there and then have the module load command in there if necessary.

ashtonmv: Good point- in this case I think we can eliminate that whole code block. As far as I can tell, the conversion to FHIaims was just to be able to read in the structure via ase_to_pyiron. Sam already had a function to read sphinx files directly

jan-janssen: Maybe we should move this discussion to a separate pull request. I have to admit I was not even aware of this part.

ashtonmv: Sure no problem- @samwaseda might have further thoughts on this. It's not really related to input parsing and I also wouldn't have noticed it if it hadn't broken the checks.

samwaseda:
I just executed these lines and failed...

```
spx = pr.create_job('Sphinx', 'spx')
spx.structure = pr.create_structure('Fe', 'bcc', 2.83)
spx.structure.set_initial_magnetic_moments([2, 2])
spx.calc_static()
spx.run()
```

I also tried the example that @ashtonmv provided in his first comment, but in both cases it says it didn't enter the main loop.

samwaseda:
I guess one problem is that in `class Input` we need `self._bool_dict = {True: "true", False: "false"}` to convert `True` to `true` and `False` to `false`

samwaseda:
Another problem is apparently `EmptyStates = auto` in `userparameters.sx` is translated correctly, but not `nEmptyStates = auto` in `PAWHamiltonian.sx`

ashtonmv:
> I also tried the example that @ashtonmv provided in his first comment, but in both cases it says it didn't enter the main loop.

Thanks Sam, it looks like it was for exactly the reasons you mentioned. I think they are fixed in the latest commit, along with a few bugs wrt spinConstraints.sx and main.sx. At least the basic Fe calc_static() and calc_minimize() are now running and exiting normally for me.

samwaseda:
Here I'd like to make a quite drastic suggestion. I'd be very much interested to know what @jan-janssen @sudarsan-surendralal and @FeLoch think.

As far as I can see, what @ashtonmv implemented here is exactly what we wanted to have. After having seen this, however, I realized that we didn't really think about how to *know* in the first place which groups are actually available. With this, I mean especially the ones that are anyway pre-defined and not the esoteric ones. Of course, tab completion will allow us to see those ones as well, but since the class is derived from `dict`, there are already quite a few attributes and also we should anticipate further extensions to this class in the future, which makes it not realistic to fully rely on autocomplete. To solve this problem, an idea I had was to get rid of `GenericParameters` altogether in `Input` and instead make `job.input` return a list of groups available. I guess getting rid of `GenericParameters` should not be a problem since if I understand it well, what should be compatible among DFT codes is a set of set-functions (e.g. `set_kpoints`). Furthermore, I see that we have the problem in the current implementation that some variables appear twice, as we saw for `nEmptyStates` and `EmptyStates`, but if `GenericParameters` is not there anymore, this problem will be solved as well.

So let me know what you think.

pyiron/sphinx/base.py
```diff
@@ -1515,6 +1553,94 @@ def write_spins_constraints(self, file_name="spins.in", cwd=None, spins_str=None
...
+
+    def to_sx_str(self, content="__self__", level=0):
+        line = ""
+        if content == "__self__":
+            content = self
+
+        for k, v in content.items():
+            k = k.split("___")[0]
+
+            if k == "eCut":
```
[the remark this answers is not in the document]
ashtonmv: Do you mean writing e.g. `EnCut = 340 / 13.606` in userparameters.sx instead? I'm happy with that but didn't want to mess with the `Input.write_file()` method invoked to write userparameters.

pyiron/sphinx/base.py
```diff
@@ -1515,6 +1553,94 @@ def write_spins_constraints(self, file_name="spins.in", cwd=None, spins_str=None
...
+        super(Group, self).__setitem__(key, value)
+
+    __getattr__ = dict.get
+    # __setattr__ = dict.__setitem__
+    __delattr__ = dict.__delitem__
+
+    def set(self, name, content):
+        self[name] = content
+
+    def set_group(self, name, group):
```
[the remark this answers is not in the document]
ashtonmv: Yes this is to add a sub-group to the group, e.g. self.input.structure.set_group("symmetry", {})
Where would you like to see `group=Group()`? Any `dict`-like object that gets added to the group is automatically converted to a Group instance via `__setitem__` and `__setattr__`

pyiron/sphinx/base.py
```diff
@@ -713,128 +984,228 @@ def _set_kpoints(
...
+                f.write("}\n")
+            for species in self.input.structure["species"]:
+                f.write("species {\n")
+                elt = species["element"]
+                f.write(f"\telement = {elt};\n")
+                for atom in species["atom"]:
+                    f.write("\tatom {\n")
+                    f.write(Group(atom).to_sx_str(level=2))
+                    f.write("\t}\n")
+                f.write("}\n")
```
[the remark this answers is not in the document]
ashtonmv: Oh nice catch; yeah they do. I _really_ like that `odict_to_spx_str` function!

Since they're replaceable with `write_group` I removed `write_species` and `write_structure` in my latest commit

pyiron/sphinx/base.py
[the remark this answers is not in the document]
samwaseda: What I meant was to offer a possibility like `job.input.structure.set_group("symmetry")` without further arguments.

pyiron/sphinx/base.py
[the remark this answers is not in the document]
samwaseda: No I mean instead of

```
1069         self.input.basis = Group({
1070             "eCut": "EnCut",
1071             "kPoint": Group({
1072                 "coords": "KpointCoords",
1073                 "weight": 1,
1074                 "relative": True
1075             }),
1076             "folding": "KpointFolding",
1077         })
```

You can put

```
1069         self.input.basis = Group({
1070             "eCut": "EnCut/13.606",
1071             "kPoint": Group({
1072                 "coords": "KpointCoords",
1073                 "weight": 1,
1074                 "relative": True
1075             }),
1076             "folding": "KpointFolding",
1077         })
```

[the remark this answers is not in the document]
samwaseda: But maybe this is a question of taste. I prefer not to have exceptions in `to_sx_str` and I also think it's rather rare that the value is changed via `job.input.basis`. Besides I guess if the user sees `EnCut/13.606` it is clear that there must be a change of unit. On the other hand some people might find it confusing to see different units for `EnCut` and `eCut`.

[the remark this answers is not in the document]
ashtonmv: I agree; I removed the exception from `to_sx_str` and built the conversion into the default `job.input.basis` (line 1070)

pyiron/sphinx/base.py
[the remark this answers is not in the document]
ashtonmv: I think that's a good idea. The behavior is now:

```python
1577    def set_group(self, name, content=None):
1578        if content is None:
1579            content = Group()
1580        self.set(name, content)
```

jan-janssen:
> Here I'd like to make a quite drastic suggestion. I'd be very much interested to know what @jan-janssen @sudarsan-surendralal and @FeLoch think.

I guess that is a longer discussion and I would like to bring it up in the next pyiron meeting. 

The motivation behind generic parameters is to have one format how to access the input of a specific code. Originally this was realized by providing a representation of the input in generic parameters - that is why all the parameters are converted to strings. This is in contrast to other frameworks like ASE. For example if you compare the pseudo potential parameters for cp2k https://github.com/jan-janssen/ase-examples/blob/master/20cp2k.ipynb and quantum espresso https://github.com/jan-janssen/ase-examples/blob/master/25quantumespresso.ipynb they are different and have nothing to do with the original input files of the codes. The pyiron idea is once you know how to access the input files and you know the input for the underlying code you can simply reuse the input scripts that you might find online. 

Historically generic parameters were based on the key value design of input files which is used in VASP, I know that is is not very suitable and it is also not very suitable for LAMMPS. So we have to rewrite it at some point. 

All I want is a way to preview the input files that are written before pyiron actually writes the files and the ability to parse existing input files to reuse them. Finally it would be nice to modify individual lines of the input files before they are written. 

This been said, there is no need to use the existing Generic Parameters class for this. Most likely it makes more sense to develop a new approach, like the one you have with ordered dictionaries and merge it with the Generic Parameters class later on.

freyso:
> I'm wondering if it's not easier to make `Input` a class derived from `Group` and write a gigantic input file `input.sx` which contains all the information which was previously written in different files.

I am very much in favour of this. It would make it easier to debug problems from a sphinx perspective. I understood that the previous version was triggered by my initial interaction with Jan, and that it had the advantage that one could easily rewrite one part of the input file. With the new way of modifying SPHInX input, this splitting is no longer needed.

samwaseda:
> 
> I guess that is a longer discussion and I would like to bring it up in the next pyiron meeting.

I'm fine with the pyiron meeting, but it might also make sense to set up another meeting with the SPHInX users.

> 
> The motivation behind generic parameters is to have one format how to access the input of a specific code. Originally this was realized by providing a representation of the input in generic parameters - that is why all the parameters are converted to strings. This is in contrast to other frameworks like ASE. For example if you compare the pseudo potential parameters for cp2k https://github.com/jan-janssen/ase-examples/blob/master/20cp2k.ipynb and quantum espresso https://github.com/jan-janssen/ase-examples/blob/master/25quantumespresso.ipynb they are different and have nothing to do with the original input files of the codes. The pyiron idea is once you know how to access the input files and you know the input for the underlying code you can simply reuse the input scripts that you might find online.
> 
> Historically generic parameters were based on the key value design of input files which is used in VASP, I know that is is not very suitable and it is also not very suitable for LAMMPS. So we have to rewrite it at some point.
> 
> All I want is a way to preview the input files that are written before pyiron actually writes the files and the ability to parse existing input files to reuse them. Finally it would be nice to modify individual lines of the input files before they are written.
> 
> This been said, there is no need to use the existing Generic Parameters class for this. Most likely it makes more sense to develop a new approach, like the one you have with ordered dictionaries and merge it with the Generic Parameters class later on.

Maybe we can first introduce something like `job.input.sphinx`, whose content will be what is going to be written in `input.sx` (or other files if necessary). Then we can switch to functions, e.g. `job.input.EnCut`, which will look up `job.input.sphinx.basis.eCut`. This way, we can keep a GenericParameters-like behaviour. Maybe we can go even further than this and create a `setter` (and `getter`), which parses the key and returns the value as it did in the previous implementation.

pyiron/sphinx/base.py
```diff
@@ -1004,117 +1375,71 @@ def compress(self, files_to_compress=None):
...
-                                                    spin_constraint_enabled=spin_constraint_enabled,
-                                                    job_name=job_name)
+        if potformat == 'JTH':
+            potentials = SphinxJTHPotentialFile(xc=xc)
+            find_potential_file = find_potential_file_jth
+            pot_path_dict = {"PBE": "jth-gga-pbe"}
+        elif potformat == 'VASP':
+            potentials = VaspPotentialFile(xc=xc)
+            find_potential_file = find_potential_file_vasp
+            pot_path_dict = {"PBE": "paw-gga-pbe", "LDA": "paw-lda"}
```
freyso, line 1406: See above: potential format and potential set may not be the same thing.

pyiron/sphinx/base.py
```diff
@@ -1515,6 +1522,93 @@ def write_spins_constraints(self, file_name="spins.in", cwd=None, spins_str=None
...
+        del self[name]
+
+    def remove_group(self, group):
+        self.remove(group)
+
+    def remove_flag(self, flag):
+        self.remove(flag)
+
+    def remove_parameter(self, parameter):
+        self.remove(parameter)
```
freyso, line 1567: why do you provide so many remove functions when they all do the same? Wouldn't 
```remove(whatever)```
not be much simpler?

File: pyiron/sphinx/base.py [hunk 7]
```diff
@@ -1515,6 +1522,93 @@ def write_spins_constraints(self, file_name="spins.in", cwd=None, spins_str=None
...
+    def set_group(self, name, content=None):
+        if content is None:
+            content = Group()
+        self.set(name, content)
+
+    def set_flag(self, flag, val=True):
+        self.set(flag, val)
+
+    def set_parameter(self, parameter, val):
+        self.set(parameter, val)
```
freyso, line 1555: wouldn't a simple
```set(whatever, content)```
not be simpler and then use the content to do the right thing?
```
set ('structure', Group ()) # set a group
set ('dipoleCorrection', True) # set a flag
set ('nEmptyStates', 23) # set a parameter
```
Largely self-explaining, no need to think about the right function to use.
One could still argue whether flags should have the ```True``` content, or if `None` content corresponds to a flag set to true...

File: pyiron/sphinx/base.py [hunk 8]
```diff
@@ -1790,7 +1905,8 @@ def get_partial_log(file_content, start_line, end_line):
                 float(line.replace("=", " ").replace(",", " ").split()[1])
                 * HARTREE_TO_EV
                 for line in log_main
-                if line.startswith("eTot(") and not line.startswith("eTot(Val)")
+                if line.startswith("eTot(") and not line.startswith(
+                    "eTot(Val)")
             ]
             energy_zero = 0.5 * (np.array(energy_free) + np.array(energy_int))
```
freyso, line 1911: This is not the best way to get these numbers. Better extract them from energy.dat.
For instance, eTot(Val) is the Kohn-Sham energy using the latest (mixed?) density, the Harris-Foulkes energy is more accurate for scfDiag. The energy_zero extrapolation depends on the smearing-scheme order (Methfessel-Paxton or order 1 Fermi-Dirac are different).

tests/sphinx/test_base.py
```diff
@@ -253,27 +268,25 @@ def test_set_kpoints(self):
...
+            ]),
+            'to___1': odict([
+                ('coords', '[0.0, 0.0, 0.5]'), ('nPoints', 20), ('label', '"N"')
+            ]),
+            'to___1___1': odict([
+                ('coords', '[0.25, 0.25, 0.25]'), ('nPoints', 0), ('label', '"P"')
+            ]),
+            'to___2': odict([
+                ('coords', '[0.5, -0.5, 0.5]'), ('nPoints', 20), ('label', '"H"')
+            ]),
```
freyso, line 288: This is ugly.
This could be nicer:
```
'to[]' = [odict([('coords', '[0.5, -0.5, 0.5]'), ('nPoints', 20), ('label', '"H"')]),
          odict([('coords', '[0.0, 0.0, 0.5]'), ('nPoints', 20), ('label', '"N"')]),
          odict([('coords', '[0.25, 0.25, 0.25]'), ('nPoints', 0), ('label', '"P"')]),
          odict([('coords', '[0.5, -0.5, 0.5]'), ('nPoints', 20), ('label', '"H"')])]
```
Technically, it might even be easy to omit the ```[]``` marker in the identifier name ```to[]```; but it might be easier for the user to identify this group that appears multiple times.

It also might be convenient to have a auxiliary function to construct such a list of odicts from a pandas table (using column name for the parameter name).

pyiron/sphinx/base.py
samwaseda: `set ('structure', Group ())` won't be possible since `Group()` will be defined only inside this class and therefore not available on the notebook. But then I guess also it's a matter of taste whether `set_group('structure')` is better or `set_group('structure', Group())`, which for me is also an argument for `set('dipoleCorrection', True)` or `set_flag('dipoleCorrection')`.

pyiron/sphinx/base.py
samwaseda: This is now what @ashtonmv did. I opened [a new issue here](https://github.com/pyiron/pyiron/issues/605)

pyiron/sphinx/base.py
freyso: what about
```
set ('structure{}')
set ('dipoleCorrection')
set ('nEmptyStates', 23)
````
? 
But I don't insist on this.

Comment (samwaseda) [reply pyiron/sphinx/base.py:1555]: I like the way @ashtonmv did it, but I don't insist either.

File: pyiron/sphinx/base.py [hunk 10]
```diff
@@ -42,36 +42,43 @@
...
     """
-    Class to setup and run Sphinx simulations which is a derivative of pyiron_atomistics.job.generic.GenericJob.
-    The functions in these modules are written in such the function names and attributes are very generic
-    (get_structure(), molecular_dynamics(), version) but the functions are written to handle Sphinx specific input and
-    output.
+    Class to setup and run Sphinx simulations which is a derivative
+    of pyiron_atomistics.job.generic.GenericJob. The functions in
+    these modules are written in such the function names and attributes
+    are very generic (get_structure(), molecular_dynamics(), version) but
+    the functions are written to handle Sphinx specific input and output.
```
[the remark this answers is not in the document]
ashtonmv: I improved this and a few other docstrings in 154e30f

> putting a new top-level group

Right now this is not really possible, but I agree it should eventually be enabled. If we agree to convert Input itself to a Group object, this will be pretty easy to add.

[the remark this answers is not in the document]
freyso: Thanks. Let's discuss "Input is a Group" in tomorrow's meeting. I regard the docu issue fixed, otherwise.

pyiron/sphinx/base.py
```diff
@@ -127,115 +140,308 @@ def exchange_correlation_functional(self):
...
+
+    def set_species_group(self, check_overlap=True, potformat='JTH'):
+        """
+        Build the species Group object based on self.structure
+
+        Args:
+            check_overlap (bool): Whether to check overlap
+                (see set_check_overlap)
+            potformat (str): type of pseudopotentials that will be
+                read. Options are JTH or VASP.
```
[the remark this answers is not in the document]
ashtonmv: I can easily change "JTH" --> "AtomPAW". But the underlying question about distinguishing psp's goes a bit over my paygrade. My understanding is that one has to add key/values to `pot_path_dict`, and then put the potentials under pyiron/resources/vasp/potentials/.

@samwaseda @jan-janssen is that right? Should we allow `pot_path_dict` to be controlled as an arg in `copy_potentials` like is already done in `find_potential_file`?

e.g.
```python
    def copy_potentials(self, potformat="JTH", xc=None, cwd=None, pot_path_dict=None):
        if pot_path_dict is None:
            pot_path_dict = {}
        if potformat == 'JTH':
            potentials = SphinxJTHPotentialFile(xc=xc)
            find_potential_file = find_potential_file_jth
            if "PBE" not in pot_path_dict:
                pot_path_dict["PBE"] = "jth-gga-pbe"
        elif potformat == 'VASP':
            potentials = VaspPotentialFile(xc=xc)
            find_potential_file = find_potential_file_vasp
            if "PBE" not in pot_path_dict:
                pot_path_dict["PBE"] = "paw-gga-pbe"
            if "LDA" not in pot_path_dict:
                pot_path_dict["LDA"] = "paw-lda"
        else:
            raise ValueError('Only JTH and VASP potentials are supported!')
```

pyiron/sphinx/base.py
```diff
@@ -338,8 +559,8 @@ def calc_md(
 
     def restart_for_band_structure_calculations(self, job_name=None):
         """
-        Restart a new job created from an existing Vasp calculation by reading the charge density
-        for band structures.
+        Restart a new job created from an existing Vasp calculation
```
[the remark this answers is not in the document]
ashtonmv: Omitted as of 3f50ba0

pyiron/sphinx/base.py
```diff
@@ -360,12 +581,15 @@ def restart_from_charge_density(
             band_structure_calc=False
     ):
         """
-        Restart a new job created from an existing Vasp calculation by reading the charge density.
+        Restart a new job created from an existing Vasp calculation
```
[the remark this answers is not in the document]
ashtonmv: VASP --> Sphinx as of 3f50ba0

pyiron/sphinx/base.py
```diff
@@ -386,11 +610,13 @@ def restart_from_wave_functions(
             job_type=None,
     ):
         """
-        Restart a new job created from an existing Vasp calculation by reading the wave functions.
+        Restart a new job created from an existing Vasp calculation
```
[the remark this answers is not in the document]
ashtonmv: VASP --> Sphinx as of 3f50ba0

pyiron/sphinx/base.py
```diff
@@ -513,25 +727,28 @@ def set_check_overlap(self, check_overlap=True):
...
             raise ValueError("check_overlap has to be a boolean")
         if self.executable.version != "2.5.1" and not check_overlap:
             vers_num = [
-                int(vv) for vv in self.executable.version.split("_")[0].split(".")
+                int(vv)
+                for vv in self.executable.version.split("_")[0].split(".")
             ]
             if (
                 vers_num[0] < 2
                 or vers_num[1] < 5
```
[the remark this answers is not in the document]
ashtonmv: how about this:

First, a new auxiliary method for SphinxBase:
```python
def get_version_float(self):
    version_str = self.executable.version.split("_")[0]
    version_float = float(
        version_str.split(".")[0]
    )
    if len(version_str.split(".")) > 1:
        version_float += float(
            "0." + "".join(version_str.split(".")[1:])
            )
    return version_float
```

and call this in `set_check_overlap`:
```python
def set_check_overlap(self, check_overlap=True):
    if not isinstance(check_overlap, bool):
        raise ValueError("check_overlap has to be a boolean")
    if self.get_version_float() < 2.51 and not check_overlap:
        warnings.warn(
            "SPHInX executable version has to be 2.5.1 or above "
            + "in order for the overlap to be considered. "
            + "Change it via job.executable.version"
        )
    self.input["CheckOverlap"] = check_overlap
```

also later in `set_main_group`:
```python
                if self.get_version_float() > 2.5:
                    self.input.main["evalForces"] = odict(
                        [("file", '"relaxHist.sx"')]
                        )
```

This is the behavior as of 510c057

pyiron/sphinx/base.py
```diff
@@ -587,10 +813,12 @@ def set_mixing_parameters(
...
     def set_occupancy_smearing(self, smearing=None, width=None):
         """
-        Set how the finite temperature smearing is applied in determining partial occupancies
+        Set how the finite temperature smearing is applied in
+        determining partial occupancies
 
         Args:
-            smearing (str): Type of smearing (only fermi si implemented anything else will be ignored)
+            smearing (str): Type of smearing (only fermi is
+                            implemented anything else will be ignored)
```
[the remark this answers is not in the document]
ashtonmv: I removed "(only fermi is supported in SPHInX)" from the error message below this, but I think we should keep the above message since only fermi is supported at the pyiron level right now. Adding MP-smearing, etc. should probably be the subject of another pull request. For now I think this is a good example of something that can be controlled by "expert" users via Groups e.g. `job.input.hamilton.MethfesselPaxton = 2`

pyiron/sphinx/base.py
```diff
@@ -662,37 +899,66 @@ def _set_kpoints(
...
         if not isinstance(symmetry_reduction, bool):
             raise ValueError("symmetry_reduction has to be a boolean")
         if manual_kpoints is not None:
             raise ValueError("manual_kpoints is not implemented in SPHInX yet")
         if weights is not None:
-            raise ValueError("manual weights are not implmented in SPHInX yet")
+            raise ValueError(
+                "manual weights are not implmented in SPHInX yet"
+                )
+
```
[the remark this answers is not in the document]
ashtonmv: Yeah I guess it's just the Pyiron wrapper, so I changed the error messages to say "not implemented in Pyiron for SPHInX".

Implementing this would be another good pull request for later.

pyiron/sphinx/base.py
```diff
@@ -713,128 +981,200 @@ def _set_kpoints(
...
+                    str(self.structure.get_high_symmetry_points()[path[1][1]])
+                    ),
+                    ("nPoints", n_path),
+                    ("label", '"' + path[1][1].replace("'", "p") + '"'),
+                ])
+            self.input.basis["kPoints"] = Group(kpoints)
         else:
-            raise ValueError("only Monkhorst-Pack mesh and Line mode are implemented in SPHInX")
+            raise ValueError("only Monkhorst-Pack mesh and Line mode\
+                are implemented in SPHInX")
```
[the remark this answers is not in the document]
ashtonmv: Same as above; replace "SPHInX" --> "Pyiron for Sphinx"

pyiron/sphinx/base.py
```diff
@@ -713,128 +981,200 @@ def _set_kpoints(
...
+            f.write("include <parameters.sx>;\n")
+            f.write("include <userparameters.sx>;\n")
+            f.write("pawPot { include <pawPot.sx>; ")
+            if enable_kjxc:
+                f.write("kjxc; ")
+            f.write("}\n")
+            f.write("structure { include <structure.sx>; }\n")
+            f.write("basis { include <basis.sx>; }\n")
+            f.write("PAWHamiltonian { include <PAWHamiltonian.sx>; }\n")
+            f.write("initialGuess { include <initialGuess.sx>; }\n")
```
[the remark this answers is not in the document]
ashtonmv: No I also wanted to do this but thought maybe @samwaseda had a specific reason to parse everything to separate files (?) and didn't want to mess with that too much. But a single file would make even more sense if we decide to represent the Input object itself as a group. Not to mention kinder to our file quotas...

pyiron/sphinx/base.py
ashtonmv: My suggestion is to send `pot_path_dict` up to become an attribute of the InputWriter object itself, so that users can modify it if they want to play with various iterations of e.g. VASP psp's. Its default is an empty dict(), which then gets populated as above.

So to change it,
```python
job = pr.create_job("Sphinx", "example")
job.input_writer.pot_path_dict = {"PBE": "paw-gga-pbe_2018"}
```
Then for that job all PBE calculations should pull potentials from pyiron/resources/vasp/potentials/paw-gga-pbe_2018. I implemented this as of f00eb1ed

pyiron/sphinx/base.py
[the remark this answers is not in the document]
freyso: This implementation restricts versioning to single-digit minor versions / patchlevels, but I don't see that this is a major problem for SPHInX. I'll be happy to name the 2.9 successor as 3.0...

pyiron/sphinx/base.py
ashtonmv: > set ('structure', Group ()) # set a group
> set ('dipoleCorrection', True) # set a flag
> set ('nEmptyStates', 23) # set a parameter

These are all already possible with the `set` function. (Except, as Sam pointed out, Group won't usually be imported into a user's notebook, but the same goal is achieved by `set('structure', {})`, since all dictionaries are internally converted to Groups within `__setitem__()`)

I personally like having the specifically-named methods along with `set`, so that if I'm just stabbing in the dark with autocomplete it's more clear what those methods would accomplish. `set` is programmatically pretty ambiguous, and if I saw it as an autocomplete option I would probably naively expect it to return a "set" of the Group's unique items. It's also nice to have defaults for `set_group` & `set_flag`, so I can just write
```python
job.input.structure.set_group("symmetry")
job.input.hamilton.set_flag("dipoleCorrection")
```
But I'm willing to remove them for purity's sake if I'm outnumbered!

freyso: ok, I give in. Let's drop my suggestions for the setter routines (in particular since I realized only later that Sam and you had a previous discussion on this, and probably
```
set ('structure',{})
```
will also work, no?

pyiron/sphinx/base.py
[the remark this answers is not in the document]
freyso: ok. I regard this fixed.

pyiron/sphinx/base.py
ashtonmv: Yes, `remove(whatever)` was also still possible. I just wanted symmetry between all the `set()` and `remove()` functions, probably because I am just OCD :)

But as of 38bbfaaa I got rid of everything but the `remove()` function.

samwaseda:
I will mention this at the meeting tomorrow, but I guess it's a bit problematic that

```
job = pr.create_job('Sphinx', 'spx')
job.structure = pr.create_structure('Fe', 'bcc', 2.8)
job.calc_static()
job.structure.positions[0,0] = 0.01
job.run()
```

does not reflect the position change because it's written after `calc_static()`.

pyiron/sphinx/base.py
ashtonmv: yep, that works

pyiron/sphinx/base.py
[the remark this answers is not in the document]
freyso: new error message fixes my concern, but good idea to put this as an open issue separate from this pull request. Let's discuss tomorrow with Felix - it might overlap with the generic k-point path stuff...

ashtonmv:
> I will mention this at the meeting tomorrow, but I guess it's a bit problematic that
> 
> ```
> job = pr.create_job('Sphinx', 'spx')
> job.structure = pr.create_structure('Fe', 'bcc', 2.8)
> job.calc_static()
> job.structure.positions[0,0] = 0.01
> job.run()
> ```
> 
> does not reflect the position change because it's written after `calc_static()`.

That's a good use case I hadn't thought of. The easy fix is of course to swap lines 3 + 4, but what you have should also be handled appropriately.

In 0e339c2c, I added a check inside of `write_input` to compare job.structure's cell, positions, and species to those that are in the job.input.structure and job.input.species groups at the last second before the files get written. If they aren't equal, then job.input.structure and job.input.species are updated to reflect job.structure. So technically any changes made via job.structure will always overwrite changes made via job.input.structure, but actually I think this is essentially the desired behavior anyway.

pyiron/sphinx/base.py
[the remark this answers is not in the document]
jan-janssen: Let me quickly describe the old implementation: I added a flag: `VaspPot=false`
https://github.com/pyiron/pyiron/blob/master/tests/sphinx/test_base.py#L224 in the input parameters which is set to `true` if the corresponding resource folder exists. This is in addition to the `Xcorr=PBE` flag which specifies the kind of potential from a given distribution vasp or JTH ? 

I guess we should define clear rules for defaults: 

- How are the default potentials selected? 
- Which settings overwrite? For example when both code specific and generic commands like `set_encut` are used.

tests/sphinx/test_base.py
jan-janssen: We started to use https://github.com/psf/black - an automated code formatter for Python, which caused some of the line breaks. In addition I would like to emphasise that we follow the python coding conventions - PEP8 - as much as possible to make it easy for Python developers to understand and modify the code later on.

pyiron/sphinx/base.py
```diff
@@ -95,7 +126,16 @@ def id_spx_to_pyi(self):
...
     def plane_wave_cutoff(self):
-        return self.input["EnCut"]
+        if "eCut" in self.input.basis:
+            if "EnCut" in self.input.basis["eCut"]:
+                return self.input["EnCut"]
+            else:
+                warnings.warn("WARNING: cutoff was modified "
+                + "directly in job.input.basis; the following "
+                + "value is in Ry)")
+                return self.input.basis["eCut"]
```
jan-janssen, line 136: The pyiron units should always be SI-Units. https://pyiron.github.io/source/faq.html#what-units-does-pyiron-use

pyiron/sphinx/base.py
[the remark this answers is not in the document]
freyso: We agreed in today's discussion to postpone this.

pyiron/sphinx/base.py
freyso: postponed.

tests/sphinx/test_base.py
freyso: Sam is too clever for me - the list-of-dicts is already implemented. But we agreed that the triple-underscore notation also should be kept for the time being, so we probably leave the test as is (maybe add a comment how the list notation would look like, and that the present version tests the underscore version?)

pyiron/sphinx/base.py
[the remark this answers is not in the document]
freyso: postponed.

pyiron/sphinx/base.py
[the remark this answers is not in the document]
samwaseda: Did we? I had the impression that we agreed that we'd write everything in one file in the end.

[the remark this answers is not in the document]
FeLoch: I had also the impression that we postponed it...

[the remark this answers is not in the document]
freyso: oh yes, Sam, you are right. What Michael suggested to postpone was to change the entire input into one Group(). But I agree that even with separate Group() objects at the pyiron level, the output could (and should) still be a single file.
I reopened the conversation.

freyso:
I copy some email conversation back into the github discussion:
Citing from @samwaseda 's email
>Yeah it was a super fruitful discussion. Thanks! Just one thing:

>> - order of writing/consistency issues, e.g. updates in positions
>> Mostly just a problem for structures, so the structure group (and the species group?) should be left blank until the last minute, or until the user intentionally calls set_structure_group().

>I understood we don’t initialize any of the values, or we set them to None. And then when run() is called, pyiron does something like:
```
EnCut = whatever_EnCut_for_spx_should_be
if ‘eCut’ is in self.input.basis.keys() and self.input.basis.eCut != EnCut:
	warn.warnings(‘eCut was set directly. For a safer behaviour, set it via job.set_encut()’)
else:
	self.input.basis.eCut = EnCut
```
> No?

Delaying **all** the "generic-parameter" insertion until the final write (with checks) was not so obvious to me, but I think it is a good idea and yields a very good user experience. However, it should still be possible to fill in the Generic Parameters at any time with an optional "force overwrite" parameter.

pyiron/sphinx/base.py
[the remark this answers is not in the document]
ashtonmv: > I had the impression that we agreed that we'd write everything in one file in the end.
But still not the structure, right?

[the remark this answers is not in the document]
freyso: I tend to agree on this, but don't have a strong opinion. On the plus side: structure can produce many lines difficult to parse by human eye, and "mistakes" in the structure are typically not correlated with mistakes elsewhere in the input file, so a simultaneous view of structure + other stuff is typically not needed.
I think this is something that you can rightfully decide on your own, but maybe add a short comment on your thoughts on this within the code to make sure that the next person does not change this again without considering your points.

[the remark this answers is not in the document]
samwaseda: > But still not the structure, right?

I'm perfectly fine with or without structure. I'll gladly accept your choice, even if you put one of them without comments, because as far as I can see, it won't be a problem even if someone changes it again later (i.e. I don't think there'd be a question of backward compatibility).

pyiron/sphinx/base.py
```diff
@@ -1161,11 +1134,24 @@ def write_group(self, group, file_name):
         with open(file_name, "w") as f:
             f.write(group.to_sx_str())
 
-    def write_input(self):
+    def write_input(self, file_name="input.sx"):
         """
         Generate all the required input files for the Sphinx job.
 
+        Creates:
+        userparameters.sx: user-defined variables in job.input
```
freyso, line 1142: Would it hurt to write the user parameters directly into input.sx? Many important settings are in here, that are important in different places in the input file later on, so having a chance to just scroll up to see what they are would be helpful for me.

samwaseda: It's maybe my personal thing, but I almost **never** open SPHInX input files. So whether things are written in different files or not hardly affects how I work.

freyso: > It's maybe my personal thing, but I almost **never** open SPHInX input files.

Is that why I get an email when SPHInX doesn't work for you ;-) ? - I don't insist on this, but having the content of userparameters.sx inside the input.sx (or maybe even putting settings right where they belong) would render **my** life easier when I am asked why SPHInX doesn't do what the pyiron users expected it to do.

ashtonmv: I like this suggestion, but is there a good reason to write these as variables to the input file at all anymore? If everything is in one file, it seems more elegant to me just to write their values straight to the places they are used

pyiron/sphinx/base.py
```diff
@@ -42,45 +42,76 @@
...
+        self.original_species = None
+        self.original_cell = None
         if self.check_vasp_potentials():
             self.input["VaspPot"] = True  # use VASP potentials if available
         self._kpoints_odict = None
         self._generic_input["restart_for_band_structure"] = False
         self._generic_input["path_name"] = None
         self._generic_input["n_path"] = None
 
+    def get_version_float(self):
```
jan-janssen, line 104: Maybe this function can be moved to the executable class in future.

pyiron/sphinx/base.py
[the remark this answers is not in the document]
jan-janssen: @ashtonmv After merging this pull request, can you open an issue with all the things we postponed? I guess these smaller changes might also be interesting exercises for master students to become familiar with the code, so we should start collecting them somewhere.

pyiron/sphinx/base.py
```diff
@@ -634,26 +897,32 @@ def set_empty_states(self, n_empty_states=None):
...
         else:
             if n_empty_states < 0:
-                raise ValueError("Number of empty states must be greater than 0")
+                raise ValueError(
+                    "Number of empty states must be greater than 0"
+                    )
             self.input["EmptyStates"] = n_empty_states
 
-    def _set_kpoints(
+    def set_kpoints(
```
jan-janssen, line 925: @sudarsan-surendralal Can you comment on this? I though that the codes should implement the `_set_kpoints` function while `set_kpoints` is only defined in `GenericDFT`.

pyiron/sphinx/base.py
```diff
@@ -1515,6 +1566,107 @@ def write_spins_constraints(self, file_name="spins.in", cwd=None, spins_str=None
...
+    def set_flag(self, flag, val=True):
+        self.set(flag, val)
+
+    def set_parameter(self, parameter, val):
+        self.set(parameter, val)
+
+    def remove(self, name):
+        del self[name]
+
+    def to_sx_str(self, content="__self__", level=0):
```
jan-janssen, line 1627: `to_sx_str` - this might be a function name the Sphinx developers remember, but to make this feature more generic and hopefully be included in other parts of pyiron we should think about a more generic name.

pyiron/sphinx/base.py
```diff
@@ -1515,6 +1566,107 @@ def write_spins_constraints(self, file_name="spins.in", cwd=None, spins_str=None
             s.logger.debug("No magnetic moments")
 
 
+class Group(dict):
```
jan-janssen, line 1569: While this class is currently Sphinx specific it might make sense to move it to https://github.com/pyiron/pyiron/blob/master/pyiron/base/generic/parameters.py to make it available for other codes as well. In addition it might be reasonable to think about a slightly more specific name of this class like InputGroup . And I noticed that quite a bit of the functionality implemented here is already available in https://github.com/cdgriffith/Box so it might even be possible to derive this class from there.

jan-janssen:
@ashtonmv Please also check that these changes are consistent with the sphinx example notebook https://github.com/pyiron/pyiron/blob/master/notebooks/bandstructure.ipynb or update the notebook accordingly. I guess there is nothing worse than an example which is not working.

pyiron/sphinx/base.py
samwaseda: I agree, but I can't remember what was the argument.

jan-janssen: I guess the main reason is that there are checks implemented on the generic level which should be available in every code https://github.com/pyiron/pyiron/blob/master/pyiron/dft/job/generic.py#L206 
Another and maybe more fancy way of implementing this would be using a decorator an example could be found here https://realpython.com/primer-on-python-decorators/#simple-decorators then we could have an `@set_kpoints` decorator which modifies the `set_kpoints` function in the specific code accordingly. Still this should not be part of this pull request.

pyiron/sphinx/base.py
samwaseda: > Is that why I get an email when SPHInX doesn't work for you ;-) ?

Yeah and if it doesn't work, then it usually doesn't work anyway, whether I look at the input files or not :)

> I like this suggestion, but is there a good reason to write these as variables to the input file at all anymore? If everything is in one file, it seems more elegant to me just to write their values straight to the places they are used

That's true. I guess as long as `to_sx_str` is available, I don't think we need `user parameters.sx`.

pyiron/sphinx/base.py
samwaseda: Maybe we can write something like `get_C_syntax`, `get_xml_syntax` and `get_python_syntax` in the hdf5-class (or maybe also `print(syntax=XXX)`). For example for

```
job['input/basis/eCut'] = 0.5
```

`job['input'].get_XXX_syntax()` (or `job['input'].print(syntax=XXX)`) returns

```
basis {
    eCut = 0.5;
}
```

```
<basis>
    eCut = 0.5 # I'm not sure if it should be <eCut> 0.5 </eCut>
</basis>
```

and

```
basis:
    eCut = 0.5
```

None of them actually follows the real syntax, but this could make it easier for the user as well to see all the contents of a certain group they're interested in.

freyso: but isn't the output specific to sx-format? I'd imagine if you reuse this (or whatever @pmrv develops) for other output formats, you'd introduce another to_whatever_str function.

pyiron/sphinx/base.py
freyso: I suggest to postpone this generalization to a new issue when you have the first example (LAMMPS by @pmrv  ?)

pyiron/sphinx/base.py
ashtonmv: > Maybe we can write something like get_C_syntax, get_xml_syntax and get_python_syntax in the hdf5-class (or maybe also print(syntax=XXX))

I like `print(syntax)` a lot. Then we just need if's and elif's for each syntax we support.

pyiron/sphinx/base.py
sudarsan-surendralal: Did we decide this? Because for the VASP module, both `set_kpoints` and `_set_kpoints` are implemented at the base class. But I agree, we should discuss this in a separate pull request.

pyiron/sphinx/base.py
samwaseda: > but isn't the output specific to sx-format? I'd imagine if you reuse this (or whatever @pmrv develops) for other output formats, you'd introduce another to_whatever_str function.

I guess it'll be independent of @pmrv, because the data has to be stored in hdf5 and we should be able to restore all the input files from there. It's indeed probably an sx-specific format, but for other codes I hope either the user can do small modifications themselves (e.g. via `replace`), or maybe we can add a few arguments which account for the modifications required.

But one way or other, I don't think it's a bad idea to have a `print` function that returns the full content of an hdf5-file (or one of its subgroups), because I sometimes find it frustrating to have to go through each tag to find the data I look for.

pmrv: Agree with @freyso, this is a sx-specific format and should be labelled as such.  I would think for now `str(job.input)` is generic enough in that it works without the user knowing what format to expect.  If we want to add multiple output formats also that feels like a separate project to me. @samwaseda  is right though in that this would ideally more or less independent of my changes.

pyiron/sphinx/base.py
ashtonmv: created #611

pyiron/sphinx/base.py
ashtonmv: @jan-janssen I also saw the Box object. `Group` already has the main features we need, but there are indeed some nice extras that might be nice to have with Box. My only hesitation was adding it to the installation requirements. But if you're okay with it, I'm all for it.

I also think we should call them `InputGroup`s, and I'm happy to leave this to @pmrv's generalization.

pyiron/sphinx/base.py
ashtonmv: Okay I reverted it to `_set_kpoints` as of 7b05294c, and created #612

ashtonmv:
> Delaying all the "generic-parameter" insertion until the final write (with checks) was not so obvious to me, but I think it is a good idea and yields a very good user experience. However, it should still be possible to fill in the Generic Parameters at any time with an optional "force overwrite" parameter.

FWIW, I am personally against waiting until the final `write_input` to populate all groups; as @jan-janssen pointed out, it should always be obvious exactly what output files will look like before a user calls `job.run()`. I'll survive if I'm overruled, but I think it's very confusing if `print(job.input.basis)` -- or someday `print(job.input)`-- returns `None` one line before `job.run()`. But I am in favor of checking every group at the last minute, to warn users about possible discrepancies between the group values and the job's generic parameters.

freyso:
Am 23.04.20 um 11:57 schrieb Michael Ashton:
>
>     Delaying all the "generic-parameter" insertion until the final
>     write (with checks) was not so obvious to me, but I think it is a
>     good idea and yields a very good user experience. However, it
>     should still be possible to fill in the Generic Parameters at any
>     time with an optional "force overwrite" parameter.
>
> FWIW, I am personally against waiting until the final |write_input| to 
> populate all groups; as @jan-janssen <https://github.com/jan-janssen> 
> pointed out, it should always be obvious exactly what output files 
> will look like before a user calls |job.run()|. I'll survive if I'm 
> overruled, but I think it's very confusing if |print(job.input.basis)| 
> -- or someday |print(job.input)|-- returns |None| one line before 
> |job.run()|. But I am in favor of checking every group at the last 
> minute, to warn users about possible discrepancies between the group 
> values and the job's generic parameters.
>
Michael, I think there is no obvious best solution - so go ahead with 
whatever you think works best. But I suggest to implement this via the 
"fill in the generic parameters to the various groups and complain about 
significant discrepancies" function, maybe with an optional argument of 
the group to be touched. This could be called twice (or more times) - 
once early on, and once before the final write - and maybe whenever you 
try to print a particular (top-level) group. Having that function will 
make it easy to adapt the behavior with little effort later on.


-------------------------------------------------
Max-Planck-Institut für Eisenforschung GmbH
Max-Planck-Straße 1
D-40237 Düsseldorf
 
Handelsregister B 2533 
Amtsgericht Düsseldorf
 
Geschäftsführung
Prof. Dr. Gerhard Dehm
Prof. Dr. Jörg Neugebauer
Prof. Dr. Dierk Raabe
Dr. Kai de Weldige
 
Ust.-Id.-Nr.: DE 11 93 58 514 
Steuernummer: 105 5891 1000


Please consider that invitations and e-mails of our institute are 
only valid if they end with …@mpie.de. 
If you are not sure of the validity please contact rco@mpie.de

Bitte beachten Sie, dass Einladungen zu Veranstaltungen und E-Mails
aus unserem Haus nur mit der Endung …@mpie.de gültig sind. 
In Zweifelsfällen wenden Sie sich bitte an rco@mpie.de
-------------------------------------------------

ashtonmv:
@jan-janssen @samwaseda can I merge this? I'm not sure who is allowed to merge to master.

The coverage check seems like it's fluctuating randomly at this point, and essentially everything new is covered.

jan-janssen:
Yes, approved pull requests can be merged any time. So you can do it on your own.

samwaseda:
> @jan-janssen @samwaseda can I merge this? I'm not sure who is allowed to merge to master.
> 
> The coverage check seems like it's fluctuating randomly at this point, and essentially everything new is covered.

I guess it's the right moment to put a milestone here.

PR #597, opened 2020-04-17
Fix compatibility to pylammpsmpi

jan-janssen:
@raynol-dsouza This should fix your issue


raynol-dsouza:
Cool! I will report back once I run with the latest fix.

PR #599, opened 2020-04-17
scaled_positions updated

samwaseda:
The warnings (which were there anyway) will be shown from now on


samwaseda:
I was wondering whether to give a guarantee limit by date or by version. What do you think? @jan-janssen @sudarsan-surendralal @liamhuber

jan-janssen:
From a software development perspective, I would say we should keep it for `0.2.X` and we can drop it in `0.3.X` . When we release `0.3.X` is a bit of a discussion. Version `0.2.X` included the interactive jobs, so we should think about what is the core feature of `0.3.X` :  https://pyiron.github.io/source/about.html

samwaseda:
> In my opinion, I think we should just throw up an error every time scaled_positions is called as a setter or getter because this can seriously mess up your calculations. A warning is something that can easily be missed if your workflow is submitted to the queuing system and all the warnings collect in a log file that nobody really notices. Thoughts?

I thought the strategy was to make a time limit for deprecated functionalities where only a warning is issued. But I also agree with @sudarsan-surendralal that in this case it's not just a deprecation, but partly also an error, so I'm also okay if I'm overruled.

PR #603, opened 2020-04-18
Surface pbc

sudarsan-surendralal:
Adding options to set the pbc when calling structures


jan-janssen:
I was just thinking about the create_structure and create_surface functions to be moved to the structure class. And then import them in the project as well as in pyiron __init__ this would allow users to import these functions without the need of creating a project.

jan-janssen:
Can you do the same for `create_structure()` and `create_ase_bulk()` ? And add all three functions in `pyiron/__init__.py`.

sudarsan-surendralal:
I made the changes but, I'm getting the following error when I test locally (not here in Travis CI)

`AttributeError: module 'pyiron.atomistics.structure' has no attribute 'atoms'`

Also `create_structure` simply calls `CrystalStructure`

jan-janssen:
That is because you import `pyironase` before the `Atoms` class is defined and in `pyironase` The `atoms` class is imported from `atoms` causing an infinite loop. Most likely we have to move these functions to a separate file.

sudarsan-surendralal:
@jan-janssen let me know if you need more changes

jan-janssen:
I do not like the import in the functions - that is not PEP8. My suggestion would be to move the functions to a separate module rather than the Atoms class.

sudarsan-surendralal:
> 
> 
> Looks good to me - Can you present it in next weeks pyiron meeting? Just a quick notebook should be fine.

Sure!

Issue #606, opened 2020-04-22
write method does not work


srmnitc:
If a structure is created in pyiron using `create_ase_bulk`, and try to write it to a file using `structure.write`, the following error is generated-
```
AttributeError: 'Atoms' object has no attribute 'get_cell_lengths_and_angles'
```
However if method `write_poscar` is imported from `pyiron.vasp.structure`, it works.

jan-janssen:
`from pyiron import pyiron_to_ase`

https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/structure/atoms.py#L4287

srmnitc:
Thanks! But I think the write function should be handled somehow. Either raise a warning or convert it internally. It is a little bit confusing when the API documentation lists the method, and it does not work.

jan-janssen:
@sudarsan-surendralal or @samwaseda  Can one of you take a look at this?

PR #607, opened 2020-04-22
Add ASE write function

jan-janssen:
Closes https://github.com/pyiron/pyiron/issues/606


pyiron/atomistics/structure/atoms.py
```diff
@@ -3106,6 +3106,14 @@ def convert_formula(elements):
         return el_list
 
     # ASE compatibility
+    def write(self, filename, format=None, **kwargs):
+        """Write atoms object to a file.
```
sudarsan-surendralal, line 3110: Better to write

```
def write(self, filename, format=None, **kwargs):
    """
    Write atoms object to a recognized file format using ase parsers.

```

PR #608, opened 2020-04-22
Pyiron tables

sudarsan-surendralal:
A new feature to filter the data using the project table itself rather than the job

```
pt_obj = PyironTable(pr)
pt_obj.db_filter_function = db_filter_function
pt_obj.filter_function = filter_function
pt_obj.add.get_incar
pt_obj.convert_to_object = False
pt_obj.create_table()
```

### DIFF
--- a/pyiron/table/datamining.py
+++ b/pyiron/table/datamining.py
@@ -525,7 +525,8 @@ class PyironTable(object):
 
     def _get_filtered_job_ids_from_project(self, recursive=True):
         project_table = self._project.job_table(recursive=recursive)
-        return project_table[self.db_filter_function(project_table)]["id"].tolist()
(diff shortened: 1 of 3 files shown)


jan-janssen:
Maybe we should try to implement some unit tests, in particular for the reloading of functions from the HDF5 file.

sudarsan-surendralal:
> 
> 
> Maybe we should try to implement some unit tests, in particular for the reloading of functions from the HDF5 file.

Done!

sudarsan-surendralal:
> 
> 
> Maybe we should also add a notebook to demonstrate the functionality of the pyiron table class. But apart from this the code looks fine to me.

Sure! Will update the examples soon!

PR #615, opened 2020-04-24
Update continuous integration environment 

jan-janssen:
Based on the discussion with @ashtonmv


jan-janssen:
After merging this pull request there are only two places where new dependencies have to be added, once in the `setup.py` for pip and once in the `environment.yml` for conda. 
https://github.com/pyiron/pyiron/blob/cienv/setup.py
https://github.com/pyiron/pyiron/blob/cienv/.ci_support/environment.yml

Two files are necessary as certain packages like `pytables` have different names in conda `pytables` and `tables` on pip.

PR #617, opened 2020-04-24
Updating example notebooks

sudarsan-surendralal:
Updating the examples on the main pyiron repository and adding a new data minig notebook based on pyiron tables: notebooks/data_mining.ipynb


amit1rrr:
@sudarsan-surendralal @jan-janssen Sorry to intrude but I noticed this PR contains Jupyter notebook files that are hard to review on GitHub due to the JSON diffs.

You might want to checkout [ReviewNB](https://www.reviewnb.com/), code review tool for notebooks. Provides visual diffs & commenting for Jupyter notebooks on GitHub. Completely free for open source repositories.

Disclaimer: I built ReviewNB

diff_and_commenting

jan-janssen:
@amit1rrr Thanks, I installed it. @sudarsan-surendralal The comparison is now available at: https://app.reviewnb.com/pyiron/pyiron/pull/617/files/

sudarsan-surendralal:
> 
> 
> I really like the structure class notebook, but I am a bit confused about the `data_mining` notebook. Calculating something inside the reduce part of map-reduce sounds to me like an anti-pattern and while it is technically possible I have the feeling it is very confusing. The workflow should be to do the calculation (map) and then collect the data (reduce). If specific properties require additional calculations then this is as secondary step.

I'm not sure how this violates the map-reduce pattern since we are acting on already filtered data. Doesn't the ability to have workflows in the reduce part make `PyIronTables` much more powerful?

jan-janssen:
The issue is the error handling, because currently there is no way for pyiron to track the relationship of the jobs - no master id is set and the number of cores is not monitored. So if you plan on developing this feature in the future I am fine with it, but I am strongly against publishing those kind of hacks as official examples.

PR #620, opened 2020-04-27
Revert "New SPHInX input parsing scheme"

samwaseda:
Reverts pyiron/pyiron#592

In `structure.sx`, `cell` does not have an equal sign and `noWavesStorage` is written in `main` (which should appear only in `initialGuess`).


jan-janssen:
@samwaseda Can you open a separate pull request for these changes?

PR #622, opened 2020-04-28
Fix sphinx

samwaseda:
This one still doesn't fully work, but you can start reviewing while I fix the rest of the problems. Right now the problem is mainly when it's used as an interactive job with a wrapper, e.g.:

```
job = pr.create_job(pr.job_type.Sphinx, 'spx_Al_interactive')
job.structure = pr.create_structure('Al', 'fcc', a_Al)
job.structure.positions[0,0] += 0.05
job.server.run_mode.interactive = True
job.calc_static()
minim = pr.create_job(pr.job_type.SxExtOptInteractive, 'sxextopt_Al')
minim.ref_job = job
minim.run()
```

Otherwise I guess it's functioning.


samwaseda:
Ok now I got the problem (which must have been there even before this fix but somehow I must have overlooked it). There's a `self.output` in `base.py`, which is derived from `GenericOutput`, which does what it should when it's a static job. When it's interactive though, I guess `self.output` has to come from `GenericInteractiveOutput`, which now is overwritten by the `self.output` in `base.py`. Here's my question to @jan-janssen: Do I understand correctly that regardless of interactive or static, `self.output` should always be derived from `GenericInteractiveOutput`?

jan-janssen:
Yes, because it falls back to the default behaviour https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/job/interactive.py#L408

samwaseda:
> Yes, because it falls back to the default behaviour https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/job/interactive.py#L408
> 
> On the other hand if your class is derived from GenericInteractive this should already be set: https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/job/interactive.py

ok thanks. I just submitted a functioning version. Now I'm gonna change the unit tests quickly and then we should probably merge this as quickly as possible

samwaseda:
FYI I don't stick to this version - the only one thing that's important to me was it works...

Issue #605, opened 2020-04-20
How to extract energy


samwaseda:
https://github.com/pyiron/pyiron/blob/0271ee40d825b58671f78b1ba3537a13241d31da/pyiron/sphinx/base.py#L1793

This is not the best way to get these numbers. Better extract them from energy.dat.
For instance, eTot(Val) is the Kohn-Sham energy using the latest (mixed?) density, the Harris-Foulkes energy is more accurate for scfDiag. The energy_zero extrapolation depends on the smearing-scheme order (Methfessel-Paxton or order 1 Fermi-Dirac are different).

_Originally posted by @freyso in https://github.com/pyiron/pyiron/pull/592_

samwaseda:
Implemented [here](https://github.com/pyiron/pyiron/issues/605)

PR #623, opened 2020-04-30
order of parsing changed

samwaseda:
Reflecting [this issue](https://github.com/pyiron/pyiron/issues/605)


samwaseda:
Since SPHInX did not return all energy values in earlier versions (2.3 and below), we have to keep the original parsing scheme as well. The energy values parsed will be overwritten if they are found in `energy.dat`

PR #625, opened 2020-04-30
Sphinx 1d profiles

ashtonmv:
We want to demonstrate @freyso's dipole correction in a notebook as supporting material for a paper, so I added a helper method `get_1D_profiles` the `sphinx.Output` class to read binary (netcdf) SPHInX output files:
- rho.sxb (charge density)
- vElStat-eV.sxb (cell potential)

I think/hope it will be a generally useful feature for other SPHInX users as well. The method basically does what sxmeshline does as a command-line tool: returns an averaged 1D profile of the volumetric data along a specified axis, intended to be convenient for plotting. The 1D averages are calculated and stored to the output parser when `collect` is called, and the actual plotting can be done by the `SphinxOutput` object later.

Here's how it works now, based on how I saw @samwaseda implemented `check_band_occupancy`:

```python
# calc_minimize a small Al 111 surface with 10 Å vacuum
# and an electric field on either side, and then

job.output.plot_density_profile(axis=2)
```

```python
job.output.plot_potential_profile(axis=2)
```

Did I add this feature correctly within the standard paradigm of Pyiron outputs? Any other feedback on how it could be improved? This time I promise I tested what I wrote with real jobs...


pyiron/sphinx/base.py
```diff
@@ -2232,6 +2233,75 @@ def collect_relaxed_hist(self, file_name="relaxHist.sx", cwd=None):
...
+                data. Defaults to 2 (z-axis).
+
+        Returns:
+            mesh, values (numpy.ndarrays): 1D grid points in Angstroms
+                (mesh) and the respective data (values) as parallel arrays
+        """
+
+        with netcdf_file(file_name, mmap=False) as f:
+            dim = [int(d) for d in f.variables["dim"]]
+            data = np.array(f.variables["mesh"][:]).reshape(dim)
```
samwaseda, line 2258: This line fails when it's magnetic

sudarsan-surendralal:
If I may, we have a `VolumetricData` class for storing and handling of volumetric data:
https://github.com/pyiron/pyiron/blob/d1f3fff9ecd9f2021649c96670ae1ec129a4935e/pyiron/atomistics/volumetric/generic.py#L21 

Vasp, for example, parses the CHGCAR/LOCPOT files and stores it as a VolumetricData object in `job["output/charge_density"]` and `job["output/electrostatic_potential"]` respectively. All methods such as planar averaging are already implemented in this class. See this example: https://pyiron.github.io/source/notebooks/hcp_workfunction.html. Maybe it's a good idea to store the data from SPHINX the same way?

samwaseda:
> If I may, we have a `VolumetricData` class for storing and handling of volumetric data:
> 
> https://github.com/pyiron/pyiron/blob/d1f3fff9ecd9f2021649c96670ae1ec129a4935e/pyiron/atomistics/volumetric/generic.py#L21
> 
> Vasp, for example, parses the CHGCAR/LOCPOT files and stores it as a VolumetricData object in `job["output/charge_density"]` and `job["output/electrostatic_potential"]` respectively. All methods such as planar averaging are already implemented in this class. See this example: https://pyiron.github.io/source/notebooks/hcp_workfunction.html. Maybe it's a good idea to store the data from SPHINX the same way?

Sounds good to me, but shouldn't it rather be in `output/dft`?

sudarsan-surendralal:
> Sounds good to me, but shouldn't it rather be in `output/dft`?

We could have volumetric data that can be generated from non-DFT calculations as well right (3D densities, scalar fields etc.). That's why we put it there.

ashtonmv:
Okay it sounds like this will take longer than I thought, and I don't want to delay Christoph's paper any longer. For now I think I'll just leave this out and include it as an external function in our notebook

sudarsan-surendralal:
Okay. But if you'd like, I could help to implement this. Since you already have the parsers, it's quite easy to make it equivalent to VASP. Let me know if it's okay

PR #630, opened 2020-05-04
calc_minimize changed to ricQN

samwaseda:
According to [sxrepo](https://sxrepo.mpie.de), `ricQN` is faster than `linQN` as a minimisation scheme. As far as I know `linQN` was there only for historical reasons so I don't think there's someone who strongly opposes the idea of switching it to `ricQN`.

### DIFF
--- a/pyiron/sphinx/base.py
+++ b/pyiron/sphinx/base.py
@@ -394,20 +394,20 @@ class SphinxBase(GenericDFTJob):
                 )
             )
         if self.input["Istep"] is not None:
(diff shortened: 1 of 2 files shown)


jan-janssen:
This currently breaks the tests

```
ERROR: test_calc_minimize (sphinx.test_base.TestSphinx)
  File "/home/vsts/work/1/s/tests/sphinx/test_base.py", line 311, in test_calc_minimize
    self.assertEqual(self.sphinx.input.sphinx.main['linQN']['maxSteps'], '50')
KeyError: 'linQN'

```

Issue #632, opened 2020-05-05
Sphinx Bandstructure example fails


jan-janssen:
Tested with the current pyiron master branch on mybinder. Most likely this is related to https://github.com/pyiron/pyiron/pull/592 

```
    file ./input.sx at line 33
    Validation error in group 'basis.kPoints.from'!
>>> 'coords' has wrong number of elements.
```

It is the same error @sudarsan-surendralal reported in https://github.com/pyiron/pyiron-resources/issues/13 

The kpoints are written as: 

```
	kPoints {
		relative;
		from {
			coords = 0.0;
			coords = 0.0;
			coords = 0.0;
			label = "GAMMA";
		}
		to {
			coords = 0.5;
			coords = -0.5;
			coords = 0.5;
			nPoints = 100;
			label = "H";
		}
		to {
			coords = 0.0;
			coords = 0.0;
			coords = 0.5;
			nPoints = 100;
			label = "N";
		}
		to {
			coords = 0.0;
			coords = 0.0;
			coords = 0.0;
			nPoints = 100;
			label = "GAMMA";
		}
		to {
			coords = 0.25;
			coords = 0.25;
			coords = 0.25;
			nPoints = 100;
			label = "P";
		}
		to {
			coords = 0.5;
			coords = -0.5;
			coords = 0.5;
			nPoints = 100;
			label = "H";
		}
		to {
			coords = 0.25;
			coords = 0.25;
			coords = 0.25;
			nPoints = 0;
			label = "P";
		}
		to {
			coords = 0.0;
			coords = 0.0;
			coords = 0.5;
			nPoints = 100;
			label = "N";
		}
```

From my understanding it should be:

```
	kPoints {
		relative;
		from {
			coords = [0.0, 0.0, 0.0];
			label = "GAMMA";
		}
		to {
			coords = [0.5, -0.5, 0.5];
			nPoints = 100;
			label = "H";
		}
		to {
			coords = [0.0, 0.0, 0.5];
			nPoints = 100;
			label = "N";
		}
		to {
			coords = [0.0, 0.0, 0.0];
			nPoints = 100;
			label = "GAMMA";
		}
		to {
			coords = [0.25, 0.25, 0.25];
			nPoints = 100;
			label = "P";
		}
		to {
			coords = [0.5, -0.5, 0.5];
			nPoints = 100;
			label = "H";
		}
		to {
			coords = [0.25, 0.25, 0.25];
			nPoints = 0;
			label = "P";
		}
		to {
[...]

PR #634, opened 2020-05-05
applied numpy array for all coordinates in k points

samwaseda:
Fix [this problem](https://github.com/pyiron/pyiron/issues/632)


samwaseda:
Right now I see that there's a server error and I cannot test it, but I guess this solves the problem

jan-janssen:
We can test it online [Binder](https://mybinder.org/v2/gh/pyiron/pyiron/spx_fix_bandstructure)

jan-janssen:
These changes break even the first part of the bandstructure calculation 

```
ERROR: test_set_kpoints (sphinx.test_base.TestSphinx)
  File "/Users/runner/runners/2.166.4/work/1/s/tests/sphinx/test_base.py", line 263, in test_set_kpoints
    self.assertEqual(kpoints_group, self.sphinx_band_structure.input.sphinx.basis.kPoints)
  File "/usr/local/miniconda/envs/pyiron/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/local/miniconda/envs/pyiron/lib/python3.8/unittest/case.py", line 902, in _baseAssertEqual
    if not first == second:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

```

jan-janssen:
Does the band structure example work with these changes? unfortunately mybinder is currently is also done.

samwaseda:
For the coordinate error it seems to be fine but somehow SPHInX fails to allocate memory

jan-janssen:
It works with 
```
conda install sphinxdft=2.6.1=h6ced99e_5
```

Still then I get the following error: 
```
+-----------------------------------------------------------------------------
| Path folding for 2D and 1D projected band structures
| WARNING: no Monkhorst-Pack folding
| for normal bandstructures, set folding=[1,1,1];
+-----------------------------------------------------------------------------
Symmetry reduction ...
| Time reversal symmetry used for reducing the k-point set ...

SxAccelerate: memory allocation failed.
```
which is most likely the memory issue you mentioned.

jan-janssen:
The issue is:

```
basis {
	eCut = 33.07438995808949;
	folding = [8, 8, 8];
	saveMemory;
	kPoints {
```

it should be: 

```
basis {
	eCut = 33.07438995808949;
	saveMemory;
	kPoints {
```

Removing the line: 

```
folding = [8, 8, 8];
```

samwaseda:
This one should work, and as long as the user knows more or less what's happening behind the scenes it should be fine, but I think we have to think about the input structure for SPHInX at some point one more time (since the GenericParameters class and SPHInX input are not intrinsically connected).

PR #644, opened 2020-05-06
Docs - use versioneer to get version number 

jan-janssen:


jan-janssen:
This does not seem to work.

PR #645, opened 2020-05-06
new create_hkl_surface function

ashtonmv:
I like how easy it is to build simple surfaces using a string like "fcc111", but as far as I can tell, there is no way to use `pyiron.atomistics.structure.generator.create_surface` to make a slab whose miller indices aren't hard-coded into pyiron. I noticed that `create_surface` accepts "ase_surf" as a surface type, but that doesn't seem to work since ase_surf takes fundamentally different args than e.g. fcc111.

The new function `create_hkl_surface` basically wraps ase.build.surface, so it can handle a more flexible set of args, including arbitrary miller indices.

```python
from pyiron.atomistics.structure.generator import create_hkl_surface, create_structure

# based on an element's implicit unit cell
s1 = create_hkl_surface("Al", [10, 8, 7], layers=20, vacuum=10)

# based on an explicit unit cell
Mo_bulk = create_structure("Mo", "bcc", 3.0)
s2 = create_hkl_surface(Mo_bulk, [10, 8, 1], layers=40, vacuum=10, center=True)
```

I'm not sure if this also needs to be accessible by Project. Right now it's not, just to make the diff as small as possible.

### DIFF
--- a/pyiron/atomistics/structure/generator.py
+++ b/pyiron/atomistics/structure/generator.py
@@ -114,11 +114,11 @@ def create_surface(
 
 def create_hkl_surface(lattice, hkl, layers, vacuum=1.0, center=False):
     """
-    Create a surface with an arbitrary surface normal (hkl).
+    Use ase.build.surface to build a surface with surface normal (hkl).
 
     Args:
-        lattice: Atoms bulk instance or str, e.g. "Fe", from
-            which to build the surface
+        lattice (pyiron.atomistics.structure.atoms.Atoms/str): bulk Atoms
+            instance or str, e.g. "Fe", from which to build the surface
         hkl (list): miller indices of surface to be created
         layers (int): # of atomic layers in the surface
         vacuum (float): vacuum spacing
@@ -128,6 +128,8 @@ def create_hkl_surface(lattice, hkl, layers, vacuum=1.0, center=False):
     Returns:
         pyiron.atomistics.structure.atoms.Atoms instance: Required surface
     """
+    # https://gitlab.com/ase/ase/blob/master/ase/lattice/surface.py
+    s.publication_add(publication_ase())
 
     surface = ase_surf(lattice, hkl, layers)
     z_max = np.max(surface.positions[:, 2])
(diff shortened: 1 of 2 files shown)


pyiron/atomistics/structure/generator.py
```diff
@@ -112,6 +112,30 @@ def create_surface(
...
+        hkl (list): miller indices of surface to be created
+        layers (int): # of atomic layers in the surface
+        vacuum (float): vacuum spacing
+        center (bool): shift all positions to center the surface
+            in the cell
+
+    Returns:
+        pyiron.atomistics.structure.atoms.Atoms instance: Required surface
+    """
+
```
sudarsan-surendralal, line 131: We need to add the line `s.publication_add(publication_ase())` because we're still using ASE surface creators.

pyiron/atomistics/structure/generator.py
```diff
@@ -112,6 +112,30 @@ def create_surface(
         return None
 
 
+def create_hkl_surface(lattice, hkl, layers, vacuum=1.0, center=False):
+    """
+    Create a surface with an arbitrary surface normal (hkl).
+
+    Args:
+        lattice: Atoms bulk instance or str, e.g. "Fe", from
```
sudarsan-surendralal, line 120: Small change in Docstrings: 

`lattice (pyiron.atomistics.structure.atoms.Atoms/str): Atoms bulk instance or str, e.g. "Fe", from`

tests/atomistics/structure/test_atoms.py
```diff
@@ -1336,10 +1336,13 @@ def test__setitem__(self):
         self.assertEqual(struct.get_chemical_formula(), 'Mg4')
 
     def test_static_functions(self):
-        self.assertIsInstance(create_ase_bulk("Al"), Atoms)
+        Al_bulk = create_ase_bulk("Al")
+        self.assertIsInstance(Al_bulk, Atoms)
         surface = create_surface("Al", "fcc111", size=(4, 4, 4), vacuum=10)
         self.assertTrue(all(surface.pbc))
         self.assertIsInstance(surface, Atoms)
+        hkl_surface = create_hkl_surface(Al_bulk, [10,8,7], layers=20, vacuum=10)
```
sudarsan-surendralal, line 1344: Another case with `center=True` would be nice (just a nitpick).

ashtonmv:
Thanks Sudarsan for the quick review. I just put all your changes in, including the new test.

PR #646, opened 2020-05-07
New contributing guidelines

### LINKED ISSUE #643: git installation instructions
Would it make sense to add something like this to the README below the conda installation instructions, to help people who want to develop their own pyiron features? Right now I'm afraid it's somewhat difficult for an outsider to know where pyiron's dependencies are (at least for us old-fashioned requirements.txt people). Unless there's a good reason to keep it there, we might also consider moving environment.yml to the main folder instead of .ci_support.

**git (developmental) install**
Alternatively, the developmental version can be installed from this repository:

    git clone https://github.com/pyiron/pyiron.git

Then install all dependencies (we recommend managing these with conda):

    conda install -f pyiron/.ci_support/environment.yml

ashtonmv:
*This is intended as a draft, which hopefully others can also contribute to.*

Closes #643, but now that pyiron is public on github and has ~15 contributors, it might be a good time to expand the contributing guidelines. Feel free to remove/replace 100% of what I wrote, since you guys are the core maintainers.

I noticed that @jan-janssen already had some stuff related to this in docs/source/developers.rst, so that's what I've built on. Nearly everything I changed is in CONTRIBUTING.rst, which developers.rst imports from for the online documentation.

To start the discussion, I have a few points/questions:

1. What's the best way for people to ask us questions (email, issues page, slack channel, etc.)? If someone wants to "join the organization", what should they do? Do we even want that?
2. Is it worth the effort to enforce issue/pull request templates [like these](https://github.com/atom/atom/blob/master/PULL_REQUEST_TEMPLATE.md)? That could just be despotic while we're still small, but as the repository grows it might become very helpful.
3. The sections "The structure of pyiron" and "The philosophy of pyiron" should probably be filled in by someone who knows more about the code than I do. IMO, the more graphically we can show pyiron's code structure the better, like the uml diagram in build_docs_on_master.sh but simpler. There's a lot of multi-level inheritance in pyiron and a naive new contributor (read: myself) will have difficulty following these around at first.


CONTRIBUTING.rst
```diff
@@ -0,0 +1,313 @@
...
+.. _What should I know before I get started?:
+What should I know before I get started?
+========================================
+
+The structure of pyiron
+-----------------------
+
+The principles of pyiron
+------------------------
+
```
sudarsan-surendralal, line 63: I think the structure & principles of pyiron is something that will change drastically in the future based on the discussions we had earlier. For the time being, we could save these lines as a comment and work on it once the new structure is finalized.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,313 @@
...
+To get the developmental (git) version of pyiron,
+
+```
+git clone https://github.com/pyiron/pyiron.git
+```
+If you don't already have the dependencies installed in your environment,
+```
+cd pyiron
+pip install .
```
sudarsan-surendralal, line 128: Wouldn't it be more consistent to install all the packages through `conda-forge` first and then overwrite the python path to use the your local branch of pyiron?

CONTRIBUTING.rst
```diff
@@ -0,0 +1,313 @@
...
+
+Git commit messages
+-------------------
+
+* Use the present tense ("Add feature" not "Added feature")
+* Use the imperative mood ("Move cursor to..." not "Moves cursor to...")
+* Limit the first line to 72 characters or less
+* Reference issues and pull requests liberally after the first line
+* When only changing documentation, include [ci skip] in the commit title
+* Consider starting the commit message with an applicable emoji:
```
sudarsan-surendralal, line 166: For pycharm users, you can install the [emoji plugin](https://plugins.jetbrains.com/plugin/9174-emoji-support-plugin) to add emojis for your commit messages: 👨‍🎨 🐎 🍰

CONTRIBUTING.rst
```diff
@@ -0,0 +1,313 @@
...
+🔥 :fire: when removing code or files
+
+💚 :green_heart: when fixing the CI build
+
+✅ :white_check_mark: when adding tests
+
+Python styleguide
+-----------------
+
+Please follow `PEP8 conventions`_ for all python code added to pyiron. Pull
```
sudarsan-surendralal, line 184: Maybe IDEs like pycharm could be recommended

CONTRIBUTING.rst
ashtonmv: > This looks great! Should we also have a markdown version `CONTRIBUTING.md` on github?

I'm happy to also make a markdown version, but it seems the rst file also renders automatically online. Everything else was in rst for the sphinx docs, so that's why I left it that way.

> I think the structure & principles of pyiron is something that will change drastically in the future based on the discussions we had earlier. For the time being, we could save these lines as a comment and work on it once the new structure is finalized.

sounds good 👍

CONTRIBUTING.rst
ashtonmv: Maybe we should add a section like "Tips and Tricks" where we include this kind of recommendation

CONTRIBUTING.rst
ashtonmv: Yeah doing it with conda does seem more elegant, but I admit I am sometimes confused by the interaction between conda and $PYTHONPATH.

Maybe conda develop is the way to go? e.g.
```bash
conda create --name pyiron_dev
conda activate pyiron_dev
conda install -c conda-forge conda-build pyiron
conda remove --force pyiron  # leaves dependencies untouched
conda develop /path/to/git/installed/pyiron  # adds git folder to site-packages/conda.pth file
```

ashtonmv:
My impression from yesterday was that we can merge this, with more changes likely to come someday. Is that right?

Issue #665, opened 2020-05-12
calc_minimize within LAMMPS is not working


lfzhu-phys:
It complains as following:
```
-----------------------------------------------------------------------------------
/u/system/SLES12/soft/pyiron/dev/pyiron_mpie/pyiron/pyiron/lammps/base.py in calc_minimize(self, e_tol, f_tol, max_iter, pressure, n_print, style)
    597             n_print=n_print,
    598             style=style,
--> 599             rotation_matrix=self._prism.R
    600         )
    601     calc_minimize.__doc__ = LammpsControl.calc_minimize.__doc__

AttributeError: 'NoneType' object has no attribute 'R'
--------------------------------------------------------------------------------------
```
Any suggestion for this? Many thanks in advance.

jan-janssen:
Please first assign a structure before calling `calc_minimize()`: 

This works: 

```
job = pr.create_job(pr.job_type.Lammps, 'lmp')
job.structure = pr.create_structure('Fe', 'bcc', 2.78)
job.calc_minimize()
```

This gives the error message above: 

```
job = pr.create_job(pr.job_type.Lammps, 'lmp')
job.calc_minimize()
```

Still I guess it would be nice to raise a warning that no structure is set.

lfzhu-phys:
Thanks, Jan. I am using machine learning potential with our automated interface method.
The error occurs even when the structure is assigned. Is that mean pr.job_type.LammpsMlip not working?

jan-janssen:
I have not worked with the Mlip potentials for quite some time but they were working for me. @pmrv recently looked into the Mlip potentials, maybe he can comment on this. But in general the error message you posted above simply says no structure is set, that is the only reason why `self._prism` can be `None`. So it seems the error is somewhere else.

pmrv:
Iirc the error was MlipLammps deriving from LammpsBase not from LammpsInteractive and that messed with the `self._prism`.  I have a fix for this on my fork, will open a PR after ADIS. ;)

lfzhu-phys:
Thanks, Jan. I will check it again and keep you posted about this issue.

lfzhu-phys:
> Iirc the error was MlipLammps deriving from LammpsBase not from LammpsInteractive and that messed with the `self._prism`. I have a fix for this on my fork, will open a PR after ADIS. ;)

That's great. Thanks a lot:)

lfzhu-phys:
Jan has solved this issue. It's working now:)

Thanks a lot:)

PR #666, opened 2020-05-12
Fix error message when calc_* is set before structure

jan-janssen:
closes https://github.com/pyiron/pyiron/issues/665

### DIFF
--- a/pyiron/lammps/base.py
+++ b/pyiron/lammps/base.py
@@ -580,7 +580,7 @@ class LammpsBase(AtomisticGenericJob):
             n_print=100,
             style='cg'
     ):
-        rotation_matrix = self._validate_structure(pressure=pressure)
+        rotation_matrix = self._get_rotation_matrix(pressure=pressure)
         # Docstring set programmatically -- Ensure that changes to signature or defaults stay consistent!
         super(LammpsBase, self).calc_minimize(
             e_tol=e_tol,
@@ -630,7 +630,7 @@ class LammpsBase(AtomisticGenericJob):
             warnings.warn(
                 "calc_md() is not implemented for the non modal interactive mode use calc_static()!"
             )
-        rotation_matrix = self._validate_structure(pressure=pressure)
+        rotation_matrix = self._get_rotation_matrix(pressure=pressure)
         super(LammpsBase, self).calc_md(
             temperature=temperature,
             pressure=pressure,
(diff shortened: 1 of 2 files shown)


jan-janssen:
I am open for suggestions, otherwise I merge the changes now, because we should anyway define more clear rules when to raise warnings and when to throw errors. As by default the user does not see the warnings.

pyiron/lammps/base.py
```diff
@@ -659,7 +659,7 @@ def calc_md(
             delta_temp=delta_temp,
             delta_press=delta_press,
             job_name=self.job_name,
-            rotation_matrix=self._prism.R
+            rotation_matrix=rotation_matrix
```
liamhuber, line 662: nit: pep8 spaces around =

pyiron/lammps/base.py
```diff
@@ -1146,7 +1146,8 @@ def _set_selective_dynamics(self):
                             "velocity___constraintz"
                         ] = "set NULL NULL 0.0"
 
-    def _ensure_requested_cell_deformation_allowed(self, pressure):
+    @staticmethod
+    def _ensure_requested_cell_deformation_allowed(structure, pressure, prism=None):
```
liamhuber, line 1150: I don't find this name makes sense for something that returns a structure. (I think I might be at fault here, I know I had my fingers on this corner of the code at some point!) Do you think `_modify_structure_to_allow_requested_deformation` is better? I mean they're both really verbose, but if we're going to be verbose anyhow we might as well be precise :p

pyiron/lammps/base.py
```diff
@@ -1160,18 +1161,48 @@ def _ensure_requested_cell_deformation_allowed(self, pressure):
...
+                        return skew_structure
                 except AttributeError:
                     warnings.warn(
                         "WARNING: Setting a calculation type which uses pressure before setting the structure risks " +
                         "constraining your cell shape evolution if non-diagonal pressures are used but the structure " +
                         "is not triclinic from the start of the calculation."
                     )
+        return structure
+
+    def _validate_structure(self, pressure):
```
liamhuber, line 1181: This was actually the guy that got me thinking about names. I find it difficult to digest that `_validate_structure` returns `rotation_matrix`.

pyiron/lammps/base.py
jan-janssen: It is a function call, so: 
`PEP 8: E251 unexpected spaces around keyword / parameter equals`

PR #667, opened 2020-05-12
pylammpsmpi clean up

jan-janssen:
Remove files which are now included in pylammpsmpi


jan-janssen:
Should have been done in https://github.com/pyiron/pyiron/pull/591

Issue #443, opened 2019-11-07
tags not updated in get_structure


samwaseda:
I noticed that the spins are not updated when I get the final structures via `get_structure`, which is not difficult to understand since the tags were historically used for initial structures, but I still find it quite confusing that I see the initial magnetic moments in the final structure. Shall we apply individual updates (i.e. `spin` is updated while `selective_dynamics` isn't etc.)?

https://github.com/pyiron/pyiron/blob/29441c4cfd78c889e781801e2043994fdfc2c8a6/pyiron/atomistics/job/atomistic.py#L630

jan-janssen:
We should also consider keeping the consistency with ASE, the have: 
`get_initial_magnetic_moments` and `get_magnetic_moments`
https://gitlab.com/ase/ase/blob/master/ase/atoms.py#L606

samwaseda:
Okay but then shall we implement `get_magnetic_moments` in `pyiron.atomistics.job.atomistic`? Or should `dft.job.generic` rather be extended?

jan-janssen:
I just noticed that I missed this issue - my suggestion would be to implement it in `dft.job.generic` for now. Still I am not sure if this issue is still active.

jan-janssen:
@samwaseda Was this fixed by https://github.com/pyiron/pyiron/pull/670 ? If it is fixed, can you close the issue?

PR #670, opened 2020-05-13
Update spin tag in structure

FeLoch:
If you use `ham.get_structure(iteration_step)`, the spin tag will be updated. Also a function `get_magnetic_moments()` is created, but for now it is located at `dft/job/generic.py`.

### DIFF
--- a/pyiron/dft/job/generic.py
+++ b/pyiron/dft/job/generic.py
@@ -116,15 +116,15 @@ class GenericDFTJob(AtomisticGenericJob):
 
         Returns:
             pyiron.atomistics.structure.atoms.Atoms: The required structure
-
-
         """
         snapshot = super(GenericDFTJob, self).get_structure(
             iteration_step=iteration_step, wrap_atoms=wrap_atoms
         )
-        spins = self.get("output/generic/dft/atom_spins")
-        if spins is not None:
-            snapshot.set_initial_magnetic_moments(spins[iteration_step])
+        try:
+            if len(self.get_magnetic_moments().shape) > 0:
+                snapshot.set_initial_magnetic_moments(self.get_magnetic_moments()[iteration_step])
+        except NotImplementedError:
+            pass
         return snapshot
 
     def set_mixing_parameters(
--- a/pyiron/dft/job/generic.py
+++ b/pyiron/dft/job/generic.py
@@ -158,6 +158,17 @@ class GenericDFTJob(AtomisticGenericJob):
             "restart_for_band_structure_calculations is not implemented for this code."
         )
 
+    def get_magnetic_moments(self):
+        """
+        Gives the magnetic moments of a calculation
+
+        Returns:
+            numpy.ndarray: array of final magmetic moments
+        """
+        raise NotImplementedError(
+            "'get_magnetic_moments' is not implemented for this code."
+        )
+
     def _set_kpoints(
         self,
         mesh=None,
--- a/pyiron/dft/job/generic.py
+++ b/pyiron/dft/job/generic.py
@@ -121,8 +121,8 @@ class GenericDFTJob(AtomisticGenericJob):
             iteration_step=iteration_step, wrap_atoms=wrap_atoms
         )
         try:
-            if len(self.get_magnetic_moments().shape) > 0:
-                snapshot.set_initial_magnetic_moments(self.get_magnetic_moments()[iteration_step])
+            if len(self.get_magnetic_moments()) > 0:
+                snapshot.set_initial_magnetic_moments(self.get_magnetic_moments(iteration_step=iteration_step))
         except NotImplementedError:
             pass
         return snapshot
@@ -158,9 +158,12 @@ class GenericDFTJob(AtomisticGenericJob):
             "restart_for_band_structure_calculations is not implemented for this code."
         )
 
-    def get_magnetic_moments(self):
+    def get_magnetic_moments(self, iteration_step=-1):
         """
         Gives the magnetic moments of a calculation for each iteration step.
+
+        Args:
+            iteration_step (int): Step for which the structure is requested
 
         Returns:
             numpy.ndarray: array of final magmetic moments
--- a/pyiron/dft/job/generic.py
+++ b/pyiron/dft/job/generic.py
@@ -160,7 +160,7 @@ class GenericDFTJob(AtomisticGenericJob):
 
     def get_magnetic_moments(self):
         """
-        Gives the magnetic moments of a calculation
+        Gives the magnetic moments of a calculation for each iteration step.
 
         Returns:
             numpy.ndarray: array of final magmetic moments
--- a/pyiron/dft/job/generic.py
+++ b/pyiron/dft/job/generic.py
@@ -120,11 +120,9 @@ class GenericDFTJob(AtomisticGenericJob):
         snapshot = super(GenericDFTJob, self).get_structure(
             iteration_step=iteration_step, wrap_atoms=wrap_atoms
         )
-        try:
-            if self.get_magnetic_moments() is not None:
-                snapshot.set_initial_magnetic_moments(self.get_magnetic_moments(iteration_step=iteration_step))
-        except NotImplementedError:
-            pass
+        spins = self.get_magnetic_moments(iteration_step=iteration_step)
+        if spins is not None:
+            snapshot.set_initial_magnetic_moments(spins)
         return snapshot
 
     def set_mixing_parameters(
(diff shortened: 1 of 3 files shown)


FeLoch:
I am not sure if this pull will close #443. @samwaseda could you take a look?

pyiron/vasp/base.py
```diff
@@ -1341,6 +1341,15 @@ def get_nelect(self):
         else:
             return self["output/generic/dft/n_elect"]
 
+    def get_magnetic_moments(self):
+        """
+        Gives the magnetic moments of a calculation
```
sudarsan-surendralal, line 1346: Maybe we need to mention they are given for each snapshot.

FeLoch: Fair enough. I changed the docstrings.

FeLoch:
> 
> 
> This seems great but I'll be nice if we have some testing for this.

Yes you're right. Maybe we can discuss it, because I have no clue how testing in pyiron works for finished calculations.

pyiron/vasp/interactive.py
```diff
@@ -56,8 +56,11 @@ def interactive_enforce_structure_reset(self, reset):
             "interactive_enforce_structure_reset() is not implemented!"
         )
 
-    def get_structure(self, iteration_step=-1):
-        return GenericInteractive.get_structure(self, iteration_step=iteration_step)
+    def get_structure(self, iteration_step=-1, wrap_atoms=True):
+        structure = GenericInteractive.get_structure(self, iteration_step=iteration_step, wrap_atoms=wrap_atoms)
+        if len(self.get_magnetic_moments()) > 0:
```
samwaseda, line 61: Shouldn't it be like this?

```
try:
    if len(self.get_magnetic_moments()) > 0:
except NotImplementedError:
    pass
```

FeLoch: Should it? Since the function is located at `vasp/interactive.py`, there won't be a `NotImplementedError`, will there?

samwaseda: Oh it's inside VASP, then it's even worse, because that would mean `get_structure` behaves differently depending on the DFT code. It should be inside generic DFT.

pyiron/vasp/base.py
```diff
@@ -1341,6 +1341,15 @@ def get_nelect(self):
...
             return self["output/generic/dft/n_elect"]
 
+    def get_magnetic_moments(self):
+        """
+        Gives the magnetic moments of a calculation for each iteration step.
+
+        Returns:
+            numpy.ndarray: array of final magmetic moments
+        """
+        return self['output/generic/dft/final_magmoms']
```
samwaseda, line 1351: To me it makes more sense that it behaves in the same way as `get_structure`, i.e. it has `snapshot` inside the argument which is set to -1.

pyiron/vasp/interactive.py
FeLoch: Right! the question is where to put it. At `dft/job/generic.py` or `atomistic/job/atomistic.py`?

FeLoch: In `dft/job/generic.py` is something similar implemented already. I will change this.
https://github.com/pyiron/pyiron/blob/8625d1c003fb848dd4a8bf05514c3b40506537c7/pyiron/dft/job/generic.py#L125

pyiron/vasp/base.py
FeLoch: Fair enough! I also thought about it before.

pyiron/vasp/interactive.py
FeLoch: > 
> 
> Oh it's inside VASP, then it's even worse, because that would mean `get_structure` behaves differently depending on the DFT code. It should be inside generic DFT.

I quess we should discuss this in detail. Since the jobs use `get_structure()` of the `VaspInteractive` class, the one in `dft/job/generic.py` will never be called!

FeLoch: I added now the changes @jan-janssen suggested at the meeting.

pyiron/vasp/base.py
```diff
@@ -1341,14 +1341,17 @@ def get_nelect(self):
...
         Gives the magnetic moments of a calculation for each iteration step.
 
+        Args:
+            iteration_step (int): Step for which the structure is requested
+
         Returns:
             numpy.ndarray: array of final magmetic moments
         """
-        return self['output/generic/dft/final_magmoms']
+        return self['output/generic/dft/final_magmoms'][iteration_step]
```
jan-janssen:
```
magmoms = self['output/generic/dft/final_magmoms']
if magmoms is not None: 
    return magmoms[iteration_step]
else: 
    return None
```

pyiron/dft/job/generic.py
```diff
@@ -116,15 +116,15 @@ def get_structure(self, iteration_step=-1, wrap_atoms=True):
...
-
         """
         snapshot = super(GenericDFTJob, self).get_structure(
             iteration_step=iteration_step, wrap_atoms=wrap_atoms
         )
-        spins = self.get("output/generic/dft/atom_spins")
-        if spins is not None:
-            snapshot.set_initial_magnetic_moments(spins[iteration_step])
+        try:
+            if len(self.get_magnetic_moments()) > 0:
```
jan-janssen:
```
magmoms = self.get_magnetic_moments(iteration_step=iteration_step)
if magmoms is not None: 
    snapshot.set_initial_magnetic_moments(magmoms)
```

pyiron/vasp/base.py
FeLoch: is already done 😄 Look at the latest commit.

pyiron/dft/job/generic.py
FeLoch: is already done 😄 Look at the latest commit.

jan-janssen: I am not a fan of `try except` is there another way how we can handle this?

pyiron/vasp/base.py
jan-janssen: But you read the data twice from the HDF5 file, while I suggest to read it once, store it in a variable and then use this variable twice.

FeLoch: That's right, your way is a little bit faster, my way consumes less memory. I don't think that any of them is improving speed or memory consumption.

jan-janssen: No, file access is always terrible, in particular when using a shared file system on an HPC cluster. While `self['output/generic/dft/final_magmoms']` looks simple, it opens the file, reads the data and closes the file again, that is not worth saving the memory.

pyiron/dft/job/generic.py
FeLoch: What I would suggest is to implement `get_magnetic_moments` in `dft/job/generic.py` as follows:
```
def get_magnetic_moments(iteration_step=-1):
      spins = self.get("output/generic/dft/atom_spins")
      if spins is not None:
          return spins[iteration_step]
      else:
          return None
```
These changes should work for Sphinx, shouldn't they? Than we do not need the  `try except`.

pyiron/vasp/base.py
FeLoch: Fair enough.

pyiron/dft/job/generic.py
jan-janssen: I like this suggestion because it is in agreement with https://pyiron.github.io/source/faq.html#which-output-quantities-are-stored-in-pyiron

FeLoch: > I like this suggestion because it is in agreement with https://pyiron.github.io/source/faq.html#which-output-quantities-are-stored-in-pyiron
BTW: There the quantity is given as `atomS_spins`, not `atom_spins`.

pyiron/vasp/base.py
```diff
@@ -1351,8 +1351,9 @@ def get_magnetic_moments(self, iteration_step=-1):
         Returns:
             numpy.ndarray/None: array of final magmetic moments or None if no magnetic moment is given
         """
-        if len(self['output/generic/dft/final_magmoms']) > 0:
-            return self['output/generic/dft/final_magmoms'][iteration_step]
+        spins = self["output/generic/dft/final_magmoms"]
+        if len(spins) > 0:
```
jan-janssen, line 1355:
```
TypeError: object of type 'NoneType' has no len()
```
So use: 
```
if is not spins and len(spins) > 0:
```

FeLoch: Are you sure? The tests work for me. As far as I know for non magnetic calculations it returns an empty numpy array.

jan-janssen: If it is not included in the HDF5 file it returns None.

FeLoch: Ah I got it! You mean for a not yet collected job! Sure!

PR #679, opened 2020-05-15
Create job template

jan-janssen:


jan-janssen:
Defining a new class in pyiron is now as simple as: 

```
class ToyJob(TemplateJob):
    def __init__(self, project, job_name):
        super(ToyJob, self).__init__(project, job_name) 
        self.input['input_energy'] = 100
        self._python_only_job = True
    
    # This function is executed 
    def run_static(self):
        with self.project_hdf5.open("output/generic") as h5out: 
             h5out["energy_tot"] = self.input["input_energy"]
        self.status.finished = True
```

PR #682, opened 2020-05-17
fix pyiron table

jan-janssen:


pyiron/table/datamining.py
```diff
@@ -802,6 +802,7 @@ def always_true_pandas(job_table):
         pandas.Series: A series of True values
 
     """
+    import pandas
```
sudarsan-surendralal, line 805: Why are you reimporting pandas?

jan-janssen: Exactly - when reloading the function from the HDF5 file all import statements have to be included in the same function. This broke my calculations.

sudarsan-surendralal: In that case, it's better to import `pandas.Series`  as `Series` in the function. This avoids the unpythonic reimporting of the `pandas` module.

pyiron/table/datamining.py
```diff
@@ -802,7 +802,8 @@ def always_true_pandas(job_table):
         pandas.Series: A series of True values
 
     """
-    return pandas.Series([True] * len(job_table), index=job_table.index)
+    import pandas.Series as Series
```
sudarsan-surendralal, line 805: Sorry! It's supposed to be `from pandas import Series`

PR #690, opened 2020-05-20
Sphinx volumetric data

ashtonmv:
I did my best to write the SphinxVolumetricData class based on the generic VolumetricData. I couldn't see if/how the generic VolumetricData class handles spin polarized densities (is it supposed to be through "diff_data"?), so I overrode the `get_average_along_axis` method. If there's a better way to treat spin polarization in the current setup, I'll change to that. If not, we should probably add spin polarization features to the generic class.

```python
pr = Project("VolumetricData")
job = pr.create_job(pr.job_type.Sphinx, 'spx_Fe_ferro')
job.structure = pr.create_structure('Fe', 'bcc', 2.8)
job.structure.set_initial_magnetic_moments([2, 2])
job.calc_static()
job.run()
```
```python
rho = job.get_charge_density()
vel = job.get_electrostatic_potential()
# returns tuple of len 2 if magnetic, or single list if not
spin_up, spin_down = rho.get_average_along_axis(2)
potential = vel.get_average_along_axis(2)
```
```python
x = np.linspace(0, job.structure.cell[0][0], len(spin_up))
fig, ax = plt.subplots(1, 2, figsize=(8, 4))
ax[0].plot(x, spin_up)
ax[0].plot(x, spin_down)
ax[1].plot(x, potential)
```


pyiron/sphinx/base.py
```diff
@@ -125,7 +126,7 @@ def plane_wave_cutoff(self):
         if "eCut" in self.input.sphinx.basis.keys():
             return self.input.sphinx.basis["eCut"] * RYDBERG_TO_EV
         else:
-            return self.input["EnCut"]
+            return self.input.sphinx["EnCut"]
```
samwaseda, line 129: Are you sure that this is the right change?

samwaseda:
I guess I'll leave it to @sudarsan-surendralal for the review.

pyiron/sphinx/base.py
ashtonmv: No, I guess I just got lost updating between input/ input.sphinx, etc. calls.

pyiron/sphinx/base.py
```diff
@@ -1399,13 +1432,13 @@ def validate_ready_to_run(self):
...
 
         if np.any([len(all_groups[group]) == 0 for group in all_groups]):
-            # warnings.warn("The following input groups have not been loaded:\n"
-            #     + ", ".join([
-            #         g for g in all_groups if len(all_groups[g]) == 0
-            #         ])
-            #     + "\n"
-            #     + "They are set to default values."
-            # )
+            warnings.warn("The following input groups have not been loaded:\n"
```
jan-janssen, line 1435: This reverts - https://github.com/pyiron/pyiron/pull/684 - from my perspective a default job should not raise a warning. In particular when we present pyiron to new users warnings always look like something is wrong/ unstable.

ashtonmv: Yeah sorry I didn't mean to uncomment it. Somehow I missed some commits locally when I git pulled.

sudarsan-surendralal:
This looks good. As I review this, a couple of things have to be clarified so that pyiron vasp and sphinx behave the same way. I implemented this a long time ago for VASP and for now, I think I'm the only one using the volumetric data features for vasp. So if we need to make changes in VASP it's completely fine.

## Where the volumetric data is stored

In the current implementation for SPHINX, the volumetric data is stored in `job["output/generic"]` whereas for VASP it's stored in `job["output"]`. Where do you think it's better? Both are fine IMO.

## Parsing the charge density and electrostatic potential files

In VASP, these files are not written/parsed by default and a flag is necessary to write them. In the above example, I see that it is written by default in sphinx. I think for most calculations they are not necessary and therefore do not need to be parased (can save storage). For VASP, we have attributes `write_charge_density` and `write_electrostatic_potential` to do this. Is it possible to do the same thing in SPHINX? 

## Spin densities

For the spin-resolved charge densities, `total_data` refers to the spin up + spin down charge density and the difference spin up - spin down charge density is in `diff_data`.

## Separate pull request for changes in VaspVolumetricData?

If we have to make changes in VASP, should be also do it in this pull request or separately?

pyiron/sphinx/base.py
```diff
@@ -1242,6 +1243,38 @@ def write_input(self):
...
             f.write(self.input.sphinx.to_sphinx(indent=0))
 
+    def get_charge_density(self):
+        """
+        Gets the charge density from the hdf5 file. This value is normalized by the volume
+
+        Returns:
+            atomistics.volumetric.generic.VolumetricData instance
+        """
+        if not self.status.finished:
```
sudarsan-surendralal, line 1253: In some cases, the status could be `warning` or `not_converged` and still have the charge density stored. So I think we should have something like, `if self.status in ["finished", "warning", "not_converged"]`. This has to be done for VASP too.

pyiron/sphinx/base.py
```diff
@@ -1242,6 +1243,38 @@ def write_input(self):
             f.write("include <parameters.sx>;\n\n")
             f.write(self.input.sphinx.to_sphinx(indent=0))
 
+    def get_charge_density(self):
```
sudarsan-surendralal, line 1246: If the way we store this is compatible with VASP, we could even shift this to `GenericDFTJob` and raise a `NotImplementedError` for other codes.

pyiron/sphinx/base.py
```diff
@@ -1242,6 +1243,38 @@ def write_input(self):
...
+        """
+        Gets the charge density from the hdf5 file. This value is normalized by the volume
+
+        Returns:
+            atomistics.volumetric.generic.VolumetricData instance
```
sudarsan-surendralal, line 1251: `pyiron.atomistics.volumetric.generic.VolumetricData instance`

pyiron/sphinx/volumetric_data.py
```diff
@@ -0,0 +1,229 @@
...
+__email__ = "surendralal@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
+
+
+BOHR_TO_ANGSTROM = (
+    scipy.constants.physical_constants["Bohr radius"][0] /
+    scipy.constants.angstrom
+)
+
```
sudarsan-surendralal, line 28: An extra blank line for pep8 necessary

pyiron/sphinx/volumetric_data.py
```diff
@@ -0,0 +1,229 @@
...
+                    np.average(np.average(self._total_data[0], axis=0), 0),
+                    np.average(np.average(self._total_data[1], axis=0), 0)
+                )
+        else:
+            if ind == 0:
+                return np.average(np.average(self._total_data, axis=1), 1)
+            elif ind == 1:
+                return np.average(np.average(self._total_data, axis=0), 1)
+            else:
+                return np.average(np.average(self._total_data, axis=0), 0)
```
sudarsan-surendralal, line 229: Extra blank line for pep8 necessary

ashtonmv:
> ## Where the volumetric data is stored
> In the current implementation for SPHINX, the volumetric data is stored in `job["output/generic"]` whereas for VASP it's stored in `job["output"]`. Where do you think it's better? Both are fine IMO.

I noticed that- It seems like pretty much everything in the sphinx hdf file is under output/generic, so that's why I ultimately went that way. I don't really know enough about the difference to have an opinion one way or another. 

> ## Parsing the charge density and electrostatic potential files
> In VASP, these files are not written/parsed by default and a flag is necessary to write them. In the above example, I see that it is written by default in sphinx. I think for most calculations they are not necessary and therefore do not need to be parased (can save storage). For VASP, we have attributes `write_charge_density` and `write_electrostatic_potential` to do this. Is it possible to do the same thing in SPHINX?

Yeah, that's a good idea. We could set `job.input.sphinx.main.scfDiag.noRhoStorage` as the default for scfDiag input groups; this would mean no charge density written by default. I couldn't find a similar flag for the potential in the manual, but I can ask @freyso if there's a way to prevent writing it. Maybe this would fit better in another pull request, since it regards the inputs?

> ## Spin densities
> For the spin-resolved charge densities, `total_data` refers to the spin up + spin down charge density and the difference spin up - spin down charge density is in `diff_data`.

Oh okay. I can change to that. I agree it's best if the two data structures look as identical as possible.

> ## Separate pull request for changes in VaspVolumetricData?
> If we have to make changes in VASP, should be also do it in this pull request or separately?

So far I don't see anything that has to change in the VASP version.

pyiron/sphinx/base.py
```diff
@@ -1242,6 +1243,42 @@ def write_input(self):
...
+
+        Returns:
+            pyiron.atomistics.volumetric.generic.VolumetricData
+        """
+        if self.status.finished not in [
+            "finished", "warning", "not_converged"
+        ]:
+            return
+        else:
+            with self.project_hdf5.open("output/generic") as ho:
```
sudarsan-surendralal, line 1258: To be compatible with VASP maybe we should store this at `output` rather than `output/generic`

ashtonmv: Okay I moved it in the latest commits. Thanks again for pointing me to the VolumetricData class, it adds a lot of nice features.

PR #692, opened 2020-05-25
Update requirements for readthedocs

jan-janssen:
http://pyiron.readthedocs.io


jan-janssen:
Fixed failing readthedocs builds

jan-janssen:
Follow up to https://github.com/pyiron/pyiron/pull/615 - to reduce the dependencies which need to be updated when additional dependences are added.

Issue #703, opened 2020-05-30
Protocol jobs not running on the queue


raynol-dsouza:
Since last week, all protocol jobs submitted to the queuing system return this error immediately after creating the sub jobs within the protocol:

```
Traceback (most recent call last):
  File "/u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/u/rads/pyiron_repos/pyiron/pyiron/base/job/wrappercmd.py", line 52, in <module>
    command_line(sys.argv[1:])
  File "/u/rads/pyiron_repos/pyiron/pyiron/base/job/wrappercmd.py", line 46, in command_line
    submit_on_remote=submit_on_remote
  File "/u/rads/pyiron_repos/pyiron/pyiron/base/job/wrapper.py", line 147, in job_wrapper_function
    job.run()
  File "/u/rads/pyiron_repos/pyiron/pyiron/base/job/wrapper.py", line 113, in run
    self.job.run_static()
  File "/u/rads/pyiron_repos/pyiron_contrib/pyiron_contrib/protocol/generic.py", line 642, in run_static
    self.execute()
  File "/u/rads/pyiron_repos/pyiron_contrib/pyiron_contrib/protocol/generic.py", line 636, in execute
    super(Protocol, self).execute()
  File "/u/rads/pyiron_repos/pyiron_contrib/pyiron_contrib/protocol/generic.py", line 437, in execute
    self.update_and_archive(self.get_output())
  File "/u/rads/pyiron_repos/pyiron_contrib/pyiron_contrib/protocol/compound/neb.py", line 163, in get_output
    'energy_pot': ~gp.calc_static.output.energy_pot[-1],
  File "/u/rads/pyiron_repos/pyiron_contrib/pyiron_contrib/protocol/utils/pointer.py", line 279, in __invert__
    return self.resolve()
  File "/u/rads/pyiron_repos/pyiron_contrib/pyiron_contrib/protocol/utils/pointer.py", line 282, in resolve
    return self._resolve_function(self._resolve_path())
  File "/u/rads/pyiron_repos/pyiron_contrib/pyiron_contrib/protocol/utils/pointer.py", line 237, in _resolve_path
    result = getattr(result, crumb_name)
[...]

PR #706, opened 2020-05-31
update JOB_CLASS_DICT automatically via metaclass

pmrv:
Since we talked about this a while ago and it came up again in #703 (though not as the cause).  I spent some time digging in the [python docs](https://docs.python.org/3.8/reference/datamodel.html#object.__init_subclass__) about subclassing.

With this `JOB_CLASS_DICT` is automatically updated when `JobCore` is subclassed independently from where, so we don't have to manually maintain it and without having to search for subclasses at runtime (as previously).  A downside is that all job types actually need to be imported on importing pyiron (hence the new imports in `pyiron.__init__`) to show up in the auto-completion. But since we want to modularize pyiron in the midterm, it might be ok.   Metaclasses are a bit scary, but hope the comments make it clear how it works.
 
I'm not sure how it interacts with loading custom jobs, but my understanding was that `JOB_CLASS_DICT` exists for the auto-completion only anyways.

### DIFF
--- a/pyiron/__init__.py
+++ b/pyiron/__init__.py
@@ -16,6 +16,42 @@ del get_versions
 # jedi fix
 fix_ipython_autocomplete()
 
+# populate namespace with default job classes, this also populates the job
+# class dict in pyiron.base.job.jobtype
+from pyiron.atomistics.structure.atoms import Atoms
+from pyiron.testing.randomatomistic import AtomisticExampleJob
+from pyiron.dft.master.convergence_encut_parallel import ConvEncutParallel
+from pyiron.dft.master.convergence_encut_serial import ConvEncutSerial
+from pyiron.atomistics.master.convergence_volume import ConvergenceVolume
+from pyiron.dft.master.convergence_kpoint_parallel import ConvKpointParallel
+from pyiron.testing.randomatomistic import ExampleJob
+from pyiron.base.master.flexible import FlexibleMaster
+from pyiron.gaussian.gaussian import Gaussian
+from pyiron.gpaw.gpaw import GpawJob
+from pyiron.thermodynamics.hessian import HessianJob
+from pyiron.lammps.lammps import Lammps
+from pyiron.atomistics.master.parallel import MapMaster
+from pyiron.atomistics.master.murnaghan import Murnaghan
+from pyiron.dft.master.murnaghan_dft import MurnaghanDFT
+from pyiron.atomistics.master.phonopy import PhonopyJob
+from pyiron.quickff.quickff import QuickFF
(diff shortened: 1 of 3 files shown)


jan-janssen:
If I understand it correctly you load all classes before they are used, while the previous code only loads the classes when they are used. In addition classes like ParallelMaster are derived from GenericJob but are never used by the end user. Given these constraints I decided to specify the classes available for the user manually.

pmrv:
Yes, it is necessary to import them now, before they show up.  But once we start splitting pyiron up, the user would have to import the sub-packages that they need anyways.

Fair point that this makes all subclasses show up, an alternative might to use a class decorator explicitly on all job types we want to offer for auto-completion
```python
def autocomplete(cls):
  from pyiron.job.jobtype import JOB_CLASS_DICT
  JOB_CLASS_DICT[cls.__name__] = cls.__module__

@autocomplete
class UserFacingJob(JobCore):
  pass
```

I'm not married to this PR though, I just saw `__init_subclass__` and thought it might be a good fit.

jan-janssen:
Can you time both variants? From my feeling using static paths should be faster when importing pyiron, because we only import one class `from pyiron import Project` instead of 30+.

pmrv:
```bash
time python -c "import pyiron"
```
takes 1.3 seconds on the master branch and 1.7 seconds on the meta branch.   The approach with the decorators is also 1.7 seconds.

jan-janssen:
I looked at the memory usage: 
```
from guppy import hpy
h=hpy()
h.heap()
```
and afterwards:
```
import pyiron
h.heap()
```

First print for just python 29MB:
```
Partition of a set of 235100 objects. Total size = 28881936 bytes.
```

Current pyiron master branch 99MB:
``` 
Partition of a set of 713563 objects. Total size = 99259704 bytes.
```

This branch 128MB:
```
Partition of a set of 953148 objects. Total size = 128047875 bytes.
```

While I like meta programming, like the introduction of meta classes here, I am not sure if increasing out memory footprint by roughly 30% is worth it.

pmrv:
I agree.  The additional memory consumption is only from having all job types immediately imported though, so maybe once we have a more modular pyiron this can be revisited.

jan-janssen:
Yes, I agree. On a deeper level it is elegancy vs. efficiency and I do not have a general solution for this one. On the one hand ugly optimised code is hard to maintain, on the other hand elegant code is usually a bit slower. Therefore I would suggest to open this pull request again, so we can discuss this general principle in the next pyiron meeting and maybe come up with some kind of rule what level of tradeoff is reasonable.

PR #709, opened 2020-06-04
forces missing in the list of interactive output tags

samwaseda:
Small bugfix


jan-janssen:
Can you add a unit test or at least a notebook to test interactive jobs inside sphinx? Like https://gitlab.mpcdf.mpg.de/pyiron/pyiron.pages.mpcdf.de/-/blob/master/tests/sphinx/sphinx_check_all.ipynb to https://github.com/pyiron/pyiron/tree/master/notebooks

PR #712, opened 2020-06-04
Doctests

jan-janssen:


jan-janssen:
Previously when the building of the docs failed the tests did not fail, now this is no longer the case. So when the building of the docs fail also the tests fail.

PR #715, opened 2020-06-08
job status bugfix

FeLoch:
See the attached issue.

### DIFF
--- a/pyiron/base/job/jobstatus.py
+++ b/pyiron/base/job/jobstatus.py
@@ -185,7 +185,7 @@ class JobStatus(object):
         from the database.
         """
         if self.database and self.job_id \
-                and any([self._status_dict[i] for i in ["running", "collect", "submitted", "initialized"]]):
+                and any([self._status_dict[i] for i in ["initialized", "submitted", "running", "collect", "suspended", "refresh"]]):
             try:
                 status = self.database.get_item_by_id(self.job_id)["status"]
             except IndexError:
--- a/pyiron/base/job/jobstatus.py
+++ b/pyiron/base/job/jobstatus.py
@@ -227,7 +227,7 @@ class JobStatus(object):
 
     def _get_status_from_dict(self):
         return [key for key, val in self._status_dict.items() if val][0]
-            
+
     def __repr__(self):
         """
         Human readable representation of the job status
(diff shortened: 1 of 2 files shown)


jan-janssen:
Please add `refresh_status()` before https://github.com/pyiron/pyiron/blob/master/pyiron/base/job/jobstatus.py#L157 rather than changing the job class.

FeLoch:
is done!

jan-janssen:
I would combine both lines: 

```
if [key for key, val in self._status_dict.items() if val][0] in ["running", "collect", "submitted"]:
             if self.database and self.job_id:
```

to: 

```
if self.database and self.job_id and any([self._status_dict[i] for i in ["running", "collect", "submitted"]]):
```

jan-janssen:
Can you add `"created"` it seems like that is required for non modal jobs which causes the tests to fail:
```
if self.database and self.job_id and any([self._status_dict[i] for i in ["running", "collect", "submitted", "created"]]): 
```

pyiron/base/job/jobstatus.py
```diff
@@ -183,7 +184,7 @@ def refresh_status(self):
         Refresh the job status - check if the database and job_id are set and if this is the case load the job status
         from the database.
         """
-        if self.database and self.job_id:
+        if self.database and self.job_id and any([self._status_dict[i] for i in ["running", "collect", "submitted"]]):
```
jan-janssen, line 187:
```suggestion
if self.database and self.job_id and any([self._status_dict[i] for i in ["running", "collect", "submitted", "created"]]):
```

FeLoch:
I guess, it would be better to just exclude the really "finished" jobs.

`for i not in ["finished", "aborted", "warning", "not_converged"]`

jan-janssen:
> I guess, it would be better to just exclude the really "finished" jobs.
> 
> `for i not in ["finished", "aborted", "warning", "not_converged"]`

But it is a for loop ...

FeLoch:
Oh shit, I should take a break ...

pyiron/base/job/jobstatus.py
FeLoch: nice!

samwaseda:
It would be great if you could remind me of the following question at the pyiron meeting if I forget it:

Is it feasible to update the `timestop` tag in the job table if the status is to be updated? I think it would good enough if it's set to the time point when the refreshing is performed.

PR #719, opened 2020-06-09
get_k_mesh_by_cell changed

samwaseda:
I had the feeling that it makes little sense to set a default k-point density, so I removed the default value and changed the argument positions accordingly. In addition to this, I corrected the argument name.

### DIFF
--- a/pyiron/dft/job/generic.py
+++ b/pyiron/dft/job/generic.py
@@ -207,7 +207,7 @@ class GenericDFTJob(AtomisticGenericJob):
         Function to setup the k-points
 
         Args:
-            mesh (list): Size of the mesh (ignored if scheme is not set to 'MP' or kpoints_per_angstrom is set)
+            mesh (list): Size of the mesh (ignored if scheme is not set to 'MP' or kpoints_per_reciprocal_angstrom is set)
             scheme (str): Type of k-point generation scheme (MP/GP(gamma point)/Manual/Line)
             center_shift (list): Shifts the center of the mesh from the gamma point by the given vector in relative coordinates
             symmetry_reduction (boolean): Tells if the symmetry reduction is to be applied to the k-points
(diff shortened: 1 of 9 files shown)


jan-janssen:
So this is a replacement for https://github.com/pyiron/pyiron/pull/565

pyiron/gpaw/gpaw.py
```diff
@@ -46,15 +46,15 @@ def plane_wave_cutoff(self):
     def plane_wave_cutoff(self, val):
         self.input["encut"] = val
 
-    def get_k_mesh_by_cell(self, cell=None, kpoints_per_angstrom=1):
+    def get_k_mesh_by_cell(self, kpoints_per_reciprocal_angstrom, cell=None):
```
jan-janssen, line 49: Why is this function defined here again? Should not GPAW be derived from the GenericDFT class to have access to this function? Duplicate code is always hard to maintain.

pyiron/dft/job/generic.py
```diff
@@ -75,15 +75,22 @@ def exchange_correlation_functional(self, val):
             "The exchange property is not implemented for this code."
         )
 
-    def get_k_mesh_by_cell(self, cell=None, kpoints_per_angstrom=1):
+    def get_k_mesh_by_cell(self, kpoints_per_reciprocal_angstrom, cell=None):
```
jan-janssen, line 78: Should this function be accessed by the user or should we use a private function instead?

pyiron/dft/job/generic.py
```diff
@@ -136,11 +143,6 @@ def set_mixing_parameters(
             "set_mixing_parameters is not implemented for this code."
         )
 
-    # def set_kmesh_density(self, kspace_per_in_ang=0.10):
```
jan-janssen, line 139: I though this is the function which is used by the users.

pyiron/dft/job/generic.py
```diff
@@ -197,7 +199,7 @@ def set_kpoints(
         manual_kpoints=None,
         weights=None,
         reciprocal=True,
-        kpoints_per_angstrom=None,
+        kpoints_per_reciprocal_angstrom=None,
```
jan-janssen, line 202: Please also update the DocString

pyiron/gpaw/gpaw.py
```diff
@@ -66,7 +66,7 @@ def set_kpoints(
         manual_kpoints=None,
         weights=None,
         reciprocal=True,
-        kpoints_per_angstrom=None,
+        kpoints_per_reciprocal_angstrom=None,
```
jan-janssen, line 69: Please also update the DocString

pyiron/sphinx/base.py
```diff
@@ -987,7 +987,7 @@ def _set_kpoints(
         manual_kpoints=None,
         weights=None,
         reciprocal=True,
-        kpoints_per_angstrom=None,
+        kpoints_per_reciprocal_angstrom=None,
```
jan-janssen, line 990: Please also update the DocString

pyiron/vasp/base.py
```diff
@@ -1025,7 +1025,7 @@ def set_kpoints(
         manual_kpoints=None,
         weights=None,
         reciprocal=True,
-        kpoints_per_angstrom=None,
+        kpoints_per_reciprocal_angstrom=None,
```
jan-janssen, line 1028: Please also update the DocString

jan-janssen:
Instead of copy and pasting the docstrings from one class to the next which is annoying when we modify them, we should follow the example of @liamhuber and use decorators for docstrings instead https://github.com/pyiron/pyiron/blob/master/pyiron/base/job/jobstatus.py#L42 then we only have to define the doctoring once and can reuse it in all the derived classes, as the Docstrings should be consistent for all codes I guess that is the way to go.

jan-janssen:
The changes were already merged with https://github.com/pyiron/pyiron/pull/721 - currently there are no changed files in this pull request, so I close it now.

PR #721, opened 2020-06-10
New Error class

samwaseda:
As we talked last time, I created this `Error` class which will hopefully allow us to detect errors more systematically. For this version, it only shows the error files in a human-readable format.

### DIFF
--- a/pyiron/base/job/generic.py
+++ b/pyiron/base/job/generic.py
@@ -1610,18 +1610,17 @@ class GenericError(object):
         self._job = job
 
     def print_message(self, string=''):
-        return self._print_error('error.msg', string=string)
+        return self._print_error(file_name='error.msg', string=string)
 
     def print_queue(self, string=''):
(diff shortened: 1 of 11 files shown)


jan-janssen:
The branch included commits from https://github.com/pyiron/pyiron/pull/719 to I removed those.

pyiron/base/job/generic.py
```diff
@@ -1604,6 +1605,25 @@ def _before_successor_calc(self, ham):
...
+    def print_message(self, string=''):
+        return self._print_error('error.msg', string=string)
+
+    def print_queue(self, string=''):
+        return self._print_error('error.out', string=string)
+
+    def _print_error(self, file_name, string='', print_yes=True):
+        if self._job[file_name] is None:
+            return False
+        else:
```
jan-janssen, line 1621: Alternatively:

```suggestion
        elif print_yes:
            print(string.join(self._job[file_name]))
        return True
```

Issue #612, opened 2020-04-23
set_kpoints for specific DFT modules


ashtonmv:
@sudarsan-surendralal Can you comment on this? I though that the codes should implement the `_set_kpoints` function while `set_kpoints` is only defined in `GenericDFT`.

_Originally posted by @jan-janssen in https://github.com/pyiron/pyiron/pull/592_

jan-janssen:
@ashtonmv sorry - I already created an issue for this https://github.com/pyiron/pyiron/issues/610 so we can close this one.

sudarsan-surendralal:
@jan-janssen, since this is a more specific issue with inheritance and not about decorators let's discuss this here.

The idea to have `set_kpoints` (or any function) defined in `GenericDFTJob` which calls `_set_kpoints` from the derived class seems good. However, if the user looks up the docstrings of the `set_kpoints` function, it returns the base class docstring. But what if you have additional options in some codes (eg. VASP  has the "Gamma" scheme while SPHINX doesn't) and you need it to be included in the docstrings as well? Wouldn't it be better to implement them in the derived class then?

jan-janssen:
@sudarsan-surendralal Do we need to change anything here? Or can we close the issue for now.

PR #726, opened 2020-06-10
VASP: Remove redundant set_kopints() function

jan-janssen:
The `set_kpoints()` function is defined once in the generic job class:
https://github.com/pyiron/pyiron/blob/master/pyiron/dft/job/generic.py#L191


sudarsan-surendralal:
As discussed in https://github.com/pyiron/pyiron/issues/612, isn't it better to have this in the derived class?

jan-janssen:
You mean when the code add specific functionality? But I thought that set_kpoints() is one of the generic functions which should be implemented in the same way for all codes, just like calc_md() is the same. Some codes support more features others less, but in general the functions should be the same, right? My fear would be that we otherwise loose transferability.

PR #727, opened 2020-06-11
External input

jan-janssen:
Access input of a scriptjob by:
```
from pyiron import ExternalInput
input_dict = ExternalInput()
```
this is the same as:
```
from pyiron import Notebook
input_dict = Notebook().get_custom_dict()
```
Besides the scriptjob input also JSON files as input are supported, just place a JSON file named "input.json" next to the notebook.

### DIFF
--- a/pyiron/base/job/external.py
+++ b/pyiron/base/job/external.py
@@ -1,0 +1,62 @@
+# coding: utf-8
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+
+from __future__ import print_function
+import json
+from pathlib2 import Path
+from pyiron.base.generic.hdfio import FileHDFio
+from pyiron.base.generic.parameters import GenericParameters
+
+
+"""
+Load input parameters for jupyter notebooks from external HDF5 or JSON file
+"""
+
+__author__ = "Osamu Waseda"
+__copyright__ = (
+    "Copyright 2020, Max-Planck-Institut für Eisenforschung GmbH - "
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Jan Janssen"
+__email__ = "janssen@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2019"
+
+
+class Notebook(object):
+    """
+    class for pyiron notebook objects
+    """
+
+    @staticmethod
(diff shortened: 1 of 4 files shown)


jan-janssen:
The support for JSON files was added in https://github.com/pyiron/pyiron/pull/717

samwaseda:
I don't really understand why these two different ways should coexist. Can you explain the advantage of the new method, and if it's convincing I'd rather put a deprecation warning for the old one.

jan-janssen:
I find it more intuitive. If you want to we can deprecate the old function.

samwaseda:
I find both of them not particularly intuitive, because they differ significantly from what we do in pyiron, i.e. we have to look it up anyway. What's for me intuitive would be something like

```
pr = Project('some_project')
input_dict = pr.load_input()
```

jan-janssen:
That sounds good to me as well, I update the PR accordingly.

jan-janssen:
@samwaseda Can you take another look?

pyiron/base/job/external.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @staticmethod
+    def store_custom_output_dict(output_dict):
+        folder = Path(".").cwd().parts[-1]
+        hdf_file = Path(".").cwd().parents[1] / folder
+        hdf_file = str(hdf_file) + ".h5"
+        hdf = FileHDFio(hdf_file)
+        hdf[folder].create_group("output")
+        for k, v in output_dict.items():
+            hdf[folder + "/output"][k] = v
```
samwaseda, line 62: Isn't it better to make the path the same as `input` or at least similar to it?

jan-janssen: I guess that would be another pull request, so far I only copied your code ;-)

samwaseda: I'm okay with having another pull request but apparently it was [your change](https://github.com/pyiron/pyiron/commit/45b9e31c51b95dcc13a7659ddcee59800ef3f2f8) 😎

samwaseda:
I guess without Docstring it's sort of impossible for the user to figure out how to use this function.

samwaseda:
And I guess it would be nice to give an error message or simply print a statement if the function is called when the input data is empty, because I'm pretty sure that one of the error sources will be that there's an error in communication between the notebooks/jobs and somehow the data gets lost.

jan-janssen:
@samwaseda I added both the DocString and the Error message. If you agree, I would like to merge this PR now.

PR #734, opened 2020-06-14
Try to fix SQLerror

jan-janssen:


jan-janssen:
Both tries did not work, therefore I am not sure what is causing the error.

PR #740, opened 2020-06-15
small bugfix

samwaseda:


samwaseda:
But frankly I'm not even sure if it's worth keeping `coarse_run`

jan-janssen:
Should not `self._coarse_run` be defined in `__init__.py`

samwaseda:
Since I usually don't like to have a lot of lines in `__init__`, I tend to use this structure, but I'm also fine with writing it in `__init__` in this case.

jan-janssen:
Python recommends to be explicit rather than implicit https://stackoverflow.com/questions/40750584/this-inspection-detects-instance-attribute-definition-outside-init-method

jan-janssen:
@samwaseda I guess you can merge this.

PR #743, opened 2020-06-16
Update spin constraints

FeLoch:
just a little bit updated + added some tests

### DIFF
--- a/pyiron/vasp/base.py
+++ b/pyiron/vasp/base.py
@@ -1619,12 +1619,15 @@ class VaspBase(GenericDFTJob):
 
         Args:
             rwigs_dict (dict): Dictionary of species and corresponding radii.
+                (structure has to be defined before)
         """
+        if not isinstance(rwigs_dict, dict):
+            raise AssertionError("'rwigs_dict' has to be a dict!")
         species_keys = self.structure.get_number_species_atoms().keys()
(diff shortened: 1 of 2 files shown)


pyiron/vasp/base.py
```diff
@@ -1619,12 +1619,17 @@ def set_rwigs(self, rwigs_dict):
...
             rwigs_dict (dict): Dictionary of species and corresponding radii.
+                (structure has to be defined before)
         """
+        if not isinstance(rwigs_dict, dict):
+            raise AssertionError("'rwigs_dict' has to be a dict!")
+        if not all([isinstance(val, (int, float)) for val in rwigs_dict.values()]):
+            raise ValueError("The values of 'rwigs_dict' has to be floats!")
         species_keys = self.structure.get_number_species_atoms().keys()
         rwigs_keys = rwigs_dict.keys()
         for i in species_keys:
```
sudarsan-surendralal, line 1630: Sorry for the annoying nitpick! But `i` usually indices a number. In this case it's better to write something like ` for sp in species_keys:`. This is an optional change!

jan-janssen: or just k instead of i ;-)

FeLoch: yes, you're totally right!

PR #701, opened 2020-05-29
new input data structure closing #674

pmrv:
there's a few open issues, that I'd like to discuss first
1. storing the list explicit vs. pickling; @jan-janssen gave good reasons today in the meeting not to pickle, my remaining question would then only be if we want to store version+type information in the HDF5 or just  the hierarchical data
2. I'm unsure about what happens (or should happen) on loading existing jobs; do we want to load old jobs with GenericParamaters, should it be automatically converted, etc.
3. I haven't added unit tests yet, only the original doctests, but I'll convert those soon-ish

I added a change keep all sphinx input groups in a single `Group()` as we discussed in the original issue, but that breaks a bunch of tests I haven't fixed yet.

Issue #738, opened 2020-06-15
How to save and launch job in different notebooks


samwaseda:
I have two notebooks which do (more or less) the following:

Notebook 1:
```
job = Project('SPX').create_job('Sphinx', 'spx')
... job creation ...
job.save()
```

Notebook 2:
```
job = Project('SPX').load('spx')
job.run()
```

When I run these notebooks, it shows `job exists already and therefore was not created!` in the second notebook and pyiron refuses to run the job. Did I do it wrong?

jan-janssen:
The save function, should not be used by the user - it is mainly an internal function. What do you try to achieve? If you want to execute the job in the second notebook you can use `job.run(run_again=True)`

samwaseda:
The second notebook is Steepest Descent. I'd like to have full flexibility in the first notebook, e.g. containing SSA, ART etc. and it would be nice if the second notebook can be submitted to the queue and simply does steepest descent.

jan-janssen:
Then you need a scrip job

jan-janssen:
I still struggle to understand what you try to achieve.

samwaseda:
Instead of creating a new pyiron class, I'm trying to create a notebook which does just one thing - steepest descent in this case, but then nothing else, so that this notebook can be considered as a class on its own. Ideally, it's like this:

Notebook 1
```
spx = Project('SPX').create_job('Sphinx', 'spx')
... job creation ...
spx.save()

scr = Project('SPX').create_job('ScriptJob', 'submit_spx')
scr.script_path = 'steepest_descent.ipynb'
scr.input['job_name'] = 'spx'
scr.run()
```

Notebook 2:
```
spx = Project(path_to_spx+'SPX').load(Project(...).get_custom_dict()['job_name'])
spx.run()
for force>0:
    ... steepest descent algorithm ...
```

jan-janssen:
Sphinx - `spx` - is an interactive job right?

jan-janssen:
Can you try `job.run(repair=True)` ?

samwaseda:
> Sphinx - `spx` - is an interactive job right?

Yes

> Can you try `job.run(repair=True)` ?

With `job.project_hdf5.create_working_directory()` in the first notebook it doesn't complain but the job doesn't seem to be starting (it just shows the execution star in the notebook cell and doesn't advance).

jan-janssen:
[...]

PR #749, opened 2020-06-19
Inputlist - 2nd try

jan-janssen:
I reverted the changes on the master branch https://github.com/pyiron/pyiron/pull/701 - and opened the pull request again


samwaseda:
Thanks Jan, this time I really really appreciated your rapid intervention.

jan-janssen:
I would like to first develop a test that fails with the same error and then start fixing it. I tired saving and reloading the job but I am still not able to reproduce the error so far - even before your fix. @samwaseda Can you post some more details about your specific sphinx calculation?

samwaseda:
```
  File "/u/samsstud/dev_sam/pyiron/pyiron/sphinx/base.py", line 798, in from_hdf
    self.input.from_hdf(self._hdf5)
  File "/u/samsstud/dev_sam/pyiron/pyiron/base/generic/inputlist.py", line 552, in from_hdf
    data = hdf_group["data"]
  File "/u/samsstud/dev_sam/pyiron/pyiron/base/generic/hdfio.py", line 803, in __getitem__
    raise ValueError("Unknown item: {}".format(item))
```

samwaseda:
I guess any job that you'd submit to the queueing system would have the same error since all of my jobs seem to have failed.

jan-janssen:
@samwaseda I tried the steps we discussed in https://github.com/pyiron/pyiron/issues/738 - mainly saving and reloading. For normal jobs it does not fail even with https://github.com/pyiron/pyiron/tree/inputlist_crash

```
from pyiron import Project
pr = Project('SPX_CHECK_ALL')
job = pr.create_job(pr.job_type.Sphinx, 'nonmodal2')
job.structure = pr.create_structure('Fe', 'bcc', a_Fe)
job.calc_static()
job._create_job_structure()
job_reload = pr.load(job.job_name)
job_reload.run()
```

samwaseda:
Now it looks like I cannot reload old jobs where `data` is not present

samwaseda:
But otherwise it looks like simulations run properly

jan-janssen:
@samwaseda Does this mean the jobs that crashed were created before updating pyiron? Then you updated and the queuing system started the jobs with the new version? 

**In summary**: Once we merge these changes we loose backwards compatibility to previous Sphinx calculations.

samwaseda:
I was trying to get the structure of an already finished calculation. There, `pr.load('old_job')` did not work. The case you described doesn't have to work, but `pr.load` should work for both new jobs and old jobs.

pmrv:
Not sure if I understand the issue.  If loading an old input structure from hdf5 is the problem, I suppose I could add some code to `Sphinx.from_hdf` that special cases if it finds `GenericParameters` in the hdf5 group.

samwaseda:
> Not sure if I understand the issue. If loading an old input structure from hdf5 is the problem, I suppose I could add some code to `Sphinx.from_hdf` that special cases if it finds `GenericParameters` in the hdf5 group.

Yeah please. I'm talking about all the old jobs that ran before this new input class was introduced.

pmrv:
Ah ok. Will do.

pmrv:
It's now possible to load old jobs, I just use the GenericParameter class to read it in and then convert manually.  I had to add a new version flag __hdf_version__ though, because the normal __version__ can also be the executable version.  I think it would be nicer if it would refer strictly to the version of the pyiron wrapper, but I guess that would be a lot of changes.

I'm not sure if `__hdf_version__` is the best name, alternatives I thought about are `__data_version__`, `__schema_version__` or just `__schema__`.

jan-janssen:
@pmrv I guess we should try to add a unit test which tests backward compatibility. 

We should create a new section in the `.travis` file just like we did for the notebook.

1. Install the old version `pip install pyiron==0.2.16`
2. run the python file `create_job.py`
3. update to the current master `pip install --force .`
4. run the python file `run_job.py`

The files should be something like this: 

For `create_job.py`:
```
from pyiron import Project
pr = Project("back")
structure = pr.create_ase_bulk("Al")
job = pr.create_job(pr.job_type.Sphinx, "old")
job.structure = structure 
job._create_structure()
```

For `create_job.py`:
```
from pyiron import Project
pr = Project("back")
job = pr.load("old")
job.run()
```

pmrv:
I agree, will think about corner cases first and add some test before the meeting tomorrow.

pmrv:
Ah, one potential issue still is that this version number is checked only in the sphinx job class (the leaf in the job class inheritance hierarchy), but the parent classes also modify the HDF5.  So that even if their version are properly incremented when they change their formats the sphinx class will not automatically now.  The manual way around would be to increment the versions of job classes whenever they or one of their dependencies change, but that seems error prone and easy to forget.  Don't have a better solution yet, though.

.travis.yml
```diff
@@ -42,6 +42,16 @@ matrix:
...
 
+    - name: backwards_compatibility
+      os: linux
+      language: python
+      env: MINICONDA="https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh" PYTHONVER="3.7"
+      install:
+        - export PATH="$HOME/miniconda/bin:$PATH"
+        - ./.ci_support/travis_setup_pyiron.sh ${MINICONDA} ${PYTHONVER}
+      script:
+        - ./..ci_support/travis_test_backwards.sh || travis_terminate 1;
```
jan-janssen, line 53:
```suggestion
        - ./.ci_support/travis_test_backwards.sh || travis_terminate 1;
```

.ci_support/travis_test_backwards.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# # Conda fix
+# eval "$(conda shell.bash hook)"
+# conda activate root
+# env
+
+pip install --force pyiron=0.2.16
```
jan-janssen, line 8:
```suggestion
pip install --force pyiron==0.2.16
```

.ci_support/travis_test_backwards.sh
```diff
@@ -0,0 +1,22 @@
...
+for t in tests/backwards/*save.py; do
+    python $t
+done
+
+pip install --force .
```
jan-janssen, line 13:
```suggestion
pip install --force --no-deps .
```

pmrv:
I force pushed to clean up the tiny typo fixes, hope it should work now

.travis.yml
```diff
@@ -42,6 +42,16 @@ matrix:
       script:
         - ./.ci_support/travis_build_notebooks.sh || travis_terminate 1;
 
+    - name: backwards_compatibility
+      os: linux
+      language: python
+      env: MINICONDA="https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh" PYTHONVER="3.7"
+      install:
+        - export PATH="$HOME/miniconda/bin:$PATH"
+        - ./.ci_support/travis_setup_pyiron.sh ${MINICONDA} ${PYTHONVER}
```
jan-janssen, line 51:
```suggestion
        - ./.ci_support/travis_setup_pyiron.sh ${MINICONDA} ${PYTHONVER}
        - ./.ci_support/travis_setup_notebooks.sh
```

.ci_support/travis_test_backwards.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# # Conda fix
+# eval "$(conda shell.bash hook)"
+# conda activate root
+# env
+
+pip install --force pyiron==0.2.16
```
jan-janssen, line 8:
```suggestion
pip install --force --ignore-installed pyiron==0.2.16
```

.ci_support/travis_test_backwards.sh
```diff
@@ -0,0 +1,22 @@
...
+for t in tests/backwards/*save.py; do
+    python $t
+done
+
+pip install --no-deps --force .
```
jan-janssen, line 13:
```suggestion
pip install --no-deps --force --ignore-installed .
```

.travis.yml
jan-janssen: This basically installs sphinx and other executables

.ci_support/travis_test_backwards.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# # Conda fix
+# eval "$(conda shell.bash hook)"
+# conda activate root
+# env
+
+pip install --force --ignore-installed --no-deps pyiron==0.2.16
+for t in tests/backwards/*save.py; do
```
jan-janssen, line 9:
```suggestion
for t in $(ls tests/backwards/*save.py); do
```
I do not see any print out so I have the feeling it might be another shell which travis is using.

.ci_support/travis_test_backwards.sh
```diff
@@ -0,0 +1,22 @@
...
+    python $t
+done
+
+pip install --force --ignore-installed --no-deps .
+i=0;
+for t in tests/backwards/*load.py; do
```
jan-janssen, line 15:
```suggestion
for t in $(ls tests/backwards/*load.py); do
```

.ci_support/travis_test_backwards.sh
pmrv: No, running it locally also doesn't produce any output except 
```
job exists already and therefore was not created!
```
just as on the travis pipeline.  (That's on loading, so there's a probably a mild bug in `Project.load`, but I didn't get to look deeply into it.)

jan-janssen: There should at least be a print out that the job was save in the database. And sphinx can not be executed without installing it and the installation is currently missing - as highlighted above. To test it you can also add print commands.

pmrv: The sphinx installation is part of `travis_setup_notebooks.sh`, right? That's in c35b226.  I've checked that it is written to hdf and the (sqlite) database on my machine after running the save script, but you're right it should give a message.  That seems to be an unrelated bug then.

jan-janssen: If you add a simple print message to the script we should see that they are currently not executed. The message that you get:
```
job exists already and therefore was not created!
```
Means the job exists locally already - maybe from a previous test.

jan-janssen: Ok I was wrong, the prints are shown nicely. Everything is fine from my perspective.

jan-janssen:
I can not approve my own pull request but from my perspective this can be merged now.

jan-janssen:
Wait - now you tried me - I meant adding the print statements to the python scripts rather then the shell script, to make sure the python scripts are executed.

jan-janssen:
Perfect - thanks again for your patience

tests/base/generic/test_inputlist.py
```diff
@@ -0,0 +1,346 @@
...
+                "nested str converted to InputList")
+
+    def test_get_item(self):
+        self.assertEqual(self.pl[0], {"foo": "bar"},
+                "index with integer does not give correct element")
+        self.assertEqual(self.pl[0]["foo"], "bar",
+                "index with string does not give correct element")
+        with self.assertRaises(IndexError,
+                               msg = "no IndexError on out of bounds index"):
+            self.pl[15]
```
jan-janssen, line 103:
```suggestion
            self.assertEqual(self.pl[15], 42)
```
Does this work? Then codacy would not complain

tests/base/generic/test_inputlist.py
```diff
@@ -0,0 +1,346 @@
...
+        with self.assertRaises(ValueError,
+                               msg = "no ValueError on invalid index type"):
+            self.pl[{}]
```
jan-janssen, line 106:
```suggestion
            self.assertEqual(self.pl[{}], 42)
```

tests/base/generic/test_inputlist.py
```diff
@@ -0,0 +1,346 @@
...
+            keys and values are strings.  This should be the output format of
+            _repr_json_
+            """
+
+            if isinstance(m, list):
+                for v in m:
+                    if isinstance(v, (list, dict)):
+                        if not rec(v):
+                            return False
+                    elif type(v) != str:
```
jan-janssen, line 283: Is there a specific reason to not use `isinstance` here?

tests/base/generic/test_inputlist.py
```diff
@@ -0,0 +1,346 @@
...
+                        return False
+            elif isinstance(m, dict):
+                for k, v in m.items():
+                    if type(k) != str:
```
jan-janssen, line 287: Is there a specific reason to not use `isinstance` here?

tests/base/generic/test_inputlist.py
```diff
@@ -0,0 +1,346 @@
...
+                        return False
+                    if isinstance(v, (list, dict)):
+                        if not rec(v):
+                            return False
+                    elif type(v) != str:
```
jan-janssen, line 292: Is there a specific reason to not use `isinstance` here?

tests/base/generic/test_inputlist.py
pmrv: I feel that this would hide the true intent of the test, checking whether invalid indices raise exceptions, how about `print()`ing it?

tests/base/generic/test_inputlist.py
pmrv: My thinking was since the docstring promises `str` and there's no (reasonable) way `repr()` returns anything but a `str`, that's what should get checked.  I'm not married to it, though.

tests/base/generic/test_inputlist.py
jan-janssen: print is fine for me

tests/base/generic/test_inputlist.py
jan-janssen: `not isinstance(v, str):` ? I am just against using `type` and prefer using `isinstance` https://premaseem.wordpress.com/2017/10/11/python-difference-between-type-and-isinstance/

jan-janssen:
Ok, once the tests pass you can merge it so we have it in the code for the pyiron meeting.

PR #753, opened 2020-06-22
add utility scripts

pmrv:
As we discussed I added two small utilities I use to make a start.  Feel free to push your script to this branch as well.


jan-janssen:
Can you add some documentation? Maybe some `--help` flag.

pmrv:
argparse generates -h/--help automatically, but I added some examples.  I also added some filtering options, so to show all or your jobs that finished today or yesterday, you could do
```bash
pyiron-ls ~ --recursive --since 1d
```
(assuming ~ is in your pyiron project path)

Here's the current output of -h

```
usage: pyiron-ls [-h] [-r] [-n NAME] [-e ELEMENTS [ELEMENTS ...]] [-s status [status ...]] [-i SINCE] [-c column [column ...]] [-a] [project]

Filter and prints jobs from pyiron project.

positional arguments:
  project               path to pyiron project

optional arguments:
  -h, --help            show this help message and exit

filter:
  select which jobs to show, all filters must be true for a job to be listed

  -r, --recursive       recurse into sub projects
  -n NAME, --name NAME  job name must contain this, regex allowed
  -e ELEMENTS [ELEMENTS ...], --elements ELEMENTS [ELEMENTS ...]
                        chemical elements that must be present in unit cell
  -s status [status ...], --status status [status ...]
                        job status must be one of the given, one of initialized, appended, created, submitted, running, aborted, collect, suspended, refresh, busy, finished, not_converged,
                        warning
  -i SINCE, --since SINCE
                        timestop must be less then the given duration before now, must be an integer and a time unit, i.e. one of d (days), h (hours), m (minutes) and s (seconds).
                        Combinations are possible, see examples.

output:
  control output style

  -c column [column ...], --columns column [column ...]
                        table columns to print, pass 'all' to print whole table, one of id, status, chemicalformula, job, subjob, project, projectpath, timestart, timestop, totalcputime,
                        computer, hamilton, hamversion, parentid, masterid
  -a, --all             show all job attributes

examples:
    Print the run time of all finished jobs:
        pyiron-ls -c job totalcputime -s finished

    Print all jobs with iron:
        pyiron-ls -e Fe

    Print all jobs that successfully finished yesterday and a bit:
        pyiron-ls -s finished -i 1d5h

    Print all jobs that were aborted less than 5 hours ago and match
    "spx.*restart":
        pyiron-ls -n "spx.*restart" -i 5h -s aborted
```

jan-janssen:
Can you add the entry point to the `setup.py` ? https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html?highlight=entrypoint#the-console-scripts-entry-point

pmrv:
Sure, but should I move the whole `util` directory then under the `pyiron` module?

jan-janssen:
@pmrv Yes

jan-janssen:
Can you add the copyright stuff ? Then it should be fine.

jan-janssen:
I am sorry for the incremental feedback. I was wondering how to test the code and the only suggestion I cam up with so far is calling the commands as subprocesses. 

In addition we should maybe try to unify the available command line utilities: 

- https://github.com/pyiron/pyiron/blob/master/pyiron/base/job/reloadfile.py
- https://github.com/pyiron/pyiron/blob/master/pyiron/base/job/wrappercmd.py
- https://github.com/pyiron/pyiron/blob/master/pyiron/base/settings/install.py

For example the other utilities use `getopt` rather than `argparse` and I just realised that they were not even included in the entry points.

pmrv:
I was thinking about a single executable before, but decided it wasn't worth it for two utils only.  But it makes more sense if the other ones should also be part of it.  It should be easy with argparse and I'll have a look, but probably not today.

For the tests I guess the easiest might to make functions that do all the work and then just call them from a "main" in the executable.  We could intercept `sys.stdout` if we wanted to check the output is correct also.

jan-janssen:
```
ERROR: base.settings.test_install (unittest.loader._FailedTest)
ImportError: Failed to import test module: base.settings.test_install
  File "/usr/share/miniconda/envs/pyiron/lib/python3.8/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/share/miniconda/envs/pyiron/lib/python3.8/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/home/vsts/work/1/s/tests/base/settings/test_install.py", line 8, in <module>
    from pyiron.base.settings.install import command_line
ImportError: cannot import name 'command_line' from 'pyiron.base.settings.install' (/home/vsts/work/1/s/pyiron/base/settings/install.py)


Ran 453 tests in 163.192s
```

jan-janssen:
I guess we just have to change the install function: https://github.com/pyiron/pyiron/blob/master/tests/base/settings/test_install.py#L17

pmrv:
Yes, I'll just change it to the `install_pyiron` function and remove `test_install_help` since the script is now just [calling](https://github.com/pyiron/pyiron/pull/753/files#diff-a0c2dd80095f8f96bb0d914151e07d69R50) that function and testing argparse is probably not useful.

pmrv:
I couldn't get rid of the `pyiron.log` even when I set the logging threshold very high (it seems to create an empty file in any case), so I just remove it as @samwaseda suggested.  Additionally I added a small [file](https://github.com/pyiron/pyiron/blob/utils/docs/source/commandline.rst) to the docs that describes the utilities and how to add new ones, do I have to add it somewhere to an index or so that it is correctly linked in the webpage?  I just checked that it is correctly built locally.

jan-janssen:
I am not so happy with deleting the log file, because it might contain valuable information.

jan-janssen:
Regarding the documentation, yes you can either add a link in the index https://github.com/pyiron/pyiron/blob/master/docs/index.rst#L122 to your document or from a more specific document like the FAQ https://github.com/pyiron/pyiron/blob/master/docs/source/faq.rst

pmrv:
> I am not so happy with deleting the log file, because it might contain valuable information.

In that case, it is possible to pass `-d/--dirty` to the main program so that it does not remove the log.  But I could also revert 498e8ef (or invert the option), I have no hard feelings personally that it would be necessary.

PR #758, opened 2020-07-01
maximum forces for the output added

samwaseda:
After having had to write `np.linalg.norm(job.output.forces, axis=-1).max(axis=-1)` so many times in order to check force convergence, I finally implemented the output property. User just has to put `job.output.f_max` -> I'm not really sure if this is a good name (although it's essentially equivalent to LAMMPS notation). Let me know what you think.

And maybe we should put in pyiron documentation that force convergence is the maximum force magnitude for structure optimisations (if it's not already written).

### DIFF
--- a/pyiron/atomistics/job/atomistic.py
+++ b/pyiron/atomistics/job/atomistic.py
@@ -836,7 +836,7 @@ class GenericOutput(object):
     @property
     def f_max(self):
         """
-            returns maximum force magnitude of each step which is used for
+            maximum force magnitude of each step which is used for
             convergence criterion of structure optimizations
         """
         return np.linalg.norm(self.forces, axis=-1).max(axis=-1)
--- a/pyiron/atomistics/job/atomistic.py
+++ b/pyiron/atomistics/job/atomistic.py
@@ -834,6 +834,14 @@ class GenericOutput(object):
         return self._job["output/generic/forces"]
 
     @property
+    def f_max(self):
+        """
+            returns maximum force magnitude of each step which is used for
+            convergence criterion of structure optimizations
+        """
+        return np.linalg.norm(self.forces, axis=-1).max(axis=-1)
+
+    @property
     def positions(self):
         return self._job["output/generic/positions"]
 
--- a/pyiron/atomistics/job/atomistic.py
+++ b/pyiron/atomistics/job/atomistic.py
@@ -834,7 +834,7 @@ class GenericOutput(object):
         return self._job["output/generic/forces"]
 
     @property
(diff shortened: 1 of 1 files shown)


pyiron/atomistics/job/atomistic.py
```diff
@@ -833,6 +833,14 @@ def energy_tot(self):
     def forces(self):
         return self._job["output/generic/forces"]
 
+    @property
+    def f_max(self):
```
sudarsan-surendralal, line 837: Maybe `force_max` is a better name? But this is fine too

pyiron/atomistics/job/atomistic.py
```diff
@@ -833,6 +833,14 @@ def energy_tot(self):
...
+        """
+            returns maximum force magnitude of each step which is used for
```
sudarsan-surendralal, line 839: No need to write "returns" since this is a property

pyiron/atomistics/job/atomistic.py
samwaseda: Yeah I also wondered whether `f_max` or `force_max`. I just thought for auto complete it might be easier to have something that doesn't start with `force`, but it's true that according to our general strategy it's probably better to write it more explicitly.

pyiron/atomistics/job/atomistic.py
sudarsan-surendralal: Yes because the user doesn't see this as a function. Maybe also mention that this is less useful for MD calculations.

samwaseda: I guess since it's specifically mentioned that it's used for structure optimisation, it's fine, isn't it?

pyiron/atomistics/job/atomistic.py
samwaseda: But I guess before changing it back and forth, I'll wait for @jan-janssen and @liamhuber

jan-janssen: I prefer `force_max` as well, because the user wants the forces and then from the forces the maximum. I could also imagine having something like `force.max` in future and then `force_max` is closer to that already.

samwaseda: That's a very good point. I guess the biggest problem is that numpy doesn't have something like `forces.norm(axis=-1).max(axis=-1)`, which would be acceptable by me.

jan-janssen, line 837:
```suggestion
    def force_max(self):
```

PR #759, opened 2020-07-02
bugfix for setdefault

samwaseda:
The problem occurred when I was doing something like this:

```
job = pr.create_job('Sphinx', 'spx')
job.structure = pr.create_structure('Fe', 'bcc', 2.83)
job.calc_static()
job.structure.set_initial_magnetic_moments([2,2])
job.calc_static()
```

When the first `job.calc_static()` loads the default input, pyiron thinks it is not a spin-polarised calculation. Then after setting the initial magnetic moments, the changes were not reflected, because SPHInX does `self.input.sphinx.PAWHamiltonian.setdefault('spinPolarized', self._spin_enabled)` internally, which (i.e. `setdefault`) apparently does not overwrite the value.

### DIFF
--- a/pyiron/sphinx/base.py
+++ b/pyiron/sphinx/base.py
@@ -466,7 +466,7 @@ class SphinxBase(GenericDFTJob):
             "ekt", self.input["Sigma"]/HARTREE_TO_EV
             )
         self.input.sphinx.PAWHamiltonian.setdefault("xc", self.input["Xcorr"])
-        self.input.sphinx.PAWHamiltonian.setdefault("spinPolarized", self._spin_enabled)
+        self.input.sphinx.PAWHamiltonian["spinPolarized"] = self._spin_enabled
 
     def load_guess_group(self, update_spins=True):
         """
--- a/pyiron/sphinx/base.py
+++ b/pyiron/sphinx/base.py
@@ -743,7 +743,6 @@ class SphinxBase(GenericDFTJob):
         )
 
         new_job.input = self.input
-        new_job.load_default_groups()
 
         if from_charge_density and os.path.isfile(
             posixpath.join(self.working_directory, "rho.sxb")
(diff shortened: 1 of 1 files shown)


jan-janssen:
I thought this is anyway changed in https://github.com/pyiron/pyiron/pull/749

samwaseda:
> I thought this is anyway changed in #749

It's not up to me to decide whether @pmrv is going to finish his changes.

pmrv:
I cherry-picked this for #749, but setdefault is used pretty extensively in the sphinx job class and so there still might be similar issues popping up.

pyiron/sphinx/base.py
```diff
@@ -743,7 +743,6 @@ def restart(
         )
 
         new_job.input = self.input
-        new_job.load_default_groups()
```
pmrv, line 746: Not sure about this.  It was just recently added in 12a530d2a / a35d483b1 I guess with the intention to fill in input fields that are not touched by user input.  For restarting a calculation `self.input` might already be populated with defaults from the original run, though

samwaseda: Yes, I remember to have added this line for some reason, but removing it does not seem to cause problems so I removed it, since conceptually `restart` for me is equivalent to `create_job` and therefore should not do `load_default_groups()` which `create_job` doesn't do.

pmrv: Ah, alright, that works then.

pmrv:
Unless you'd rather discuss this first in the [meeting](https://github.com/orgs/pyiron/teams/pyiron/discussions/9) I'd think we merge this now and I rebase #749 on top of master or merge master into it.  I'm working on it, but can't promise a working version until Friday.

samwaseda:
> Unless you'd rather discuss this first in the [meeting](https://github.com/orgs/pyiron/teams/pyiron/discussions/9) I'd think we merge this now and I rebase #749 on top of master or merge master into it. 

I guess we should talk about this tomorrow because I'm not really sure what to overwrite and what not to. So I guess some of `setdefault` should be replaced but some shouldn't be.

jan-janssen:
I close this one because the changes were included in https://github.com/pyiron/pyiron/pull/749

PR #764, opened 2020-07-09
repeat_points added

samwaseda:
Points are repeated in the same way as `repeat`, e.g.:

```
structure.repeat_points(points=[0.1, 0.2, 0.3], rep=[3, 3, 3])
```

The return value is the set of points repeated in each direction.

Tests and DocString will be added shortly.


samwaseda:
By the way I'll show why I needed this function at the next ADIS meeting :)

PR #766, opened 2020-07-15
deleted get_equivalent_atoms 

samwaseda:
I'm not even sure if anyone has ever used this function, but I don't think it was working correctly and without DocString it's not even clear what's the difference between this one and `get_symmetry()` (although I guess there was no difference).

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -2695,53 +2695,6 @@ class Atoms(object):
         )
         return mapping, mesh_points
 
-    def get_equivalent_atoms(self, eps=1e-5):
-        """
-
-        Args:
-            eps:
-
-        Returns:
-
-        """
-        sym = self.get_symmetry()
-        coords = np.mod(self.get_scaled_positions(wrap=False) + eps, 1) - eps
-
-        trans_vec = []
-        rot_vec = []
-        id_vec = []
-
-        ind_ref = 0  # TODO: extend as loop over all inequivalent atoms
-        id_mat = np.identity(3, dtype="intc")
-        ref_id_list = []
-        for trans, rot in zip(sym["translations"], sym["rotations"]):
-            if np.linalg.norm(rot - id_mat) < eps:  # TODO: remove this limitation
-                id_list = []
-                for i_c, coord_new in enumerate(np.mod(coords - trans + eps, 1) - eps):
-                    no_match = True
-                    hash_id = None
-                    for hash_id, c in enumerate(coords):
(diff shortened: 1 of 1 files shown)


samwaseda:
It's possible that I implemented the function though...

jan-janssen:
The function is over three years old, which makes it a bit difficult to track back the origin https://gitlab.mpcdf.mpg.de/pyiron/pyiron/-/blob/38f03ec7e55659d06004bab14257e836845ad096/PyIron/Structure/Atoms.py#L1468 

@sudarsan-surendralal Do you remember if it is used for the water calculation?

jan-janssen:
In addition it is no an ASE function, so I am open to removing it https://gitlab.com/search?utf8=✓&snippets=&scope=&repository_ref=&search=get_equivalent_atoms&project_id=470007

jan-janssen:
Most likely the easiest way is to quickly discuss it in the next pyiron meeting.

samwaseda:
> The function is over three years old, which makes it a bit difficult to track back the origin https://gitlab.mpcdf.mpg.de/pyiron/pyiron/-/blob/38f03ec7e55659d06004bab14257e836845ad096/PyIron/Structure/Atoms.py#L1468
> 
> @sudarsan-surendralal Do you remember if it is used for the water calculation?

Ok I first thought I implemented it two years ago when I started doing SSA, but probably it wasn't me in the end. Then I guess Jörg was the one who implemented it ... ?

pmrv:
> The function is over three years old, which makes it a bit difficult to track back the origin https://gitlab.mpcdf.mpg.de/pyiron/pyiron/-/blob/38f03ec7e55659d06004bab14257e836845ad096/PyIron/Structure/Atoms.py#L1468
> 
> @sudarsan-surendralal Do you remember if it is used for the water calculation?

iirc, it's used in one of the handbook notebooks, would be easy for me to update though, if we merge this.

samwaseda:
> Most likely the easiest way is to quickly discuss it in the next pyiron meeting.

Maybe it's also time to separate various functionalities inside `atoms.py`, since it has close to 5,000 lines and does everything from box creation to structure analysis.

sudarsan-surendralal:
> 
> 
> The function is over three years old, which makes it a bit difficult to track back the origin https://gitlab.mpcdf.mpg.de/pyiron/pyiron/-/blob/38f03ec7e55659d06004bab14257e836845ad096/PyIron/Structure/Atoms.py#L1468
> 
> @sudarsan-surendralal Do you remember if it is used for the water calculation?

No I think only Jörg knows what this does actually! Not sure if any of us use it

PR #768, opened 2020-07-16
Get equivalent points

samwaseda:
Returns all symmetry-equivalent points of a given point.

```
structure.get_equivalent_points([0, 0, 0.5])
```

I guess this is what Jörg was suggesting (not sure).


samwaseda:
I'll wait for the pyiron meeting for this one as well, since I'm not perfectly sure if this is what Jörg was suggesting

Issue #773, opened 2020-07-17
version X.Y milestones


pmrv:
As we are just discussed, we should collect the milestones before we want before the next (major) version bump:

Some points we mentioned as candidates
- docs
  - [ ] consistent documentation & website #779 
  - [ ] [high-level docs](https://www.joelonsoftware.com/2000/10/02/painless-functional-specifications-part-1-why-bother/), how do modules interact, reponsibilities
  - [ ] website polishing? 
- increase test coverage, 80%, candidate:
  - [ ] gaussian
  - [ ] gpaw
  - [ ] thermodynamics
  - [ ] dft, base & project are around ~50-60%
- [ ] use `InputList` for all major codes & remove `GenericParameters`
- [ ] ground work for pyiron/experimental (vs. making a separate module)
- [ ] pyiron GUI via [ipytree](https://github.com/QuantStack/ipytree)
- [ ] new Atoms class from ase
- [ ] protocol integration, lives currently in https://github.com/pyiron/pyiron_contrib 
- [ ] remote SQL database access, HDF-only installations?
- [ ] dynamic object loading from hdf (automatic from object metadata?)
- [ ] identify parts that can be split into separate repositories
  - [ ] `pyiron.base` is essentially standalone, everything else could be `atomistic`

As for the minor bump maybe 65d76edfa55e2e87b607549c0f1f0098ba6f8f4c.  That would be the commit just before merging the inputlist branch.

ashtonmv:
Sudarsan's refactored Atoms class comes to mind- should we consider that to be a milestone?

jan-janssen:
I would be interested to work on the database interface, mainly supporting clusters which do not have a local SQL database.

pmrv:
> Sudarsan's refactored Atoms class comes to mind- should we consider that to be a milestone?

See I knew missed a bit.

samwaseda:
I'd love to talk about the consistency with the website. A long standing issue is for example the output quantities, partially given [here](https://pyiron.github.io/source/faq.html#which-output-quantities-are-stored-in-pyiron), which VASP doesn't follow. But then there are also some stuff that should be updated as well I guess.

raynol-dsouza:
I think a basic version of protocols (maybe NEB, for starters) could also be included.

pmrv:
See #792, #793, #794, #795

PR #779, opened 2020-07-20
Consistency of codes + website

samwaseda:
In response to the [major update issue](https://github.com/pyiron/pyiron/issues/773), I started to look at the consistency between codes and the content on the website. Feel free to add more stuff.

### DIFF
--- a/pyiron/lammps/base.py
+++ b/pyiron/lammps/base.py
@@ -573,8 +573,10 @@ class LammpsBase(AtomisticGenericJob):
 
     def calc_minimize(
             self,
-            e_tol=0.0,
-            f_tol=1e-4,
+            ionic_energy=0.0,
+            ionic_forces=1e-4,
+            e_tol=None,
+            f_tol=None,
             max_iter=1000000,
             pressure=None,
             n_print=100,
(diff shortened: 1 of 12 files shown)


samwaseda:
@jan-janssen Please add this to the list of topics for this Friday.

pyiron/lammps/control.py
```diff
@@ -250,8 +250,10 @@ def pressure_to_lammps(self, pressure, rotation_matrix):
 
     def calc_minimize(
```
jan-janssen, line 251: This function is never called from the user directly, so we do not need `e_tol` and `f_tol` here.

pyiron/lammps/base.py
```diff
@@ -573,25 +573,41 @@ def collect_output_log(self, file_name="log.lammps", cwd=None):
...
+            ionic_energy_tolerance = e_tol
+            e_tol = None
+        if f_tol is not None:
+            warnings.warn(
+                "f_tol is deprecated as of vers. 0.3.0. It is not guaranteed to be in service in vers. 0.4.0. Use ionic_force_tolerance instead.",
+                DeprecationWarning
+            )
+            ionic_force_tolerance = f_tol
+            f_tol = None
         super(LammpsBase, self).calc_minimize(
```
jan-janssen, line 601: Here we should set both - e_tol, f_tol and ionic_energy_tolerance, ionic_force_tolerance so the warnings are raised from AtomisticGenericJob.

pyiron/lammps/control.py
samwaseda: Shouldn’t it be a private function then? In general I was not really sure which functions can be used by the user and which ones cannot be.

jan-janssen: The issue is that pep8 defines private functions as those which should only be used within the same class and derived. But in pyiron we have three levels: 

- user level - what is used in the notebook
- developer level - what is used in derived classes but should never be called from the user directly
- class level - what is only used within one class but not in other classes - _private functions_

jan-janssen:
@samwaseda Can you merge this one?

PR #785, opened 2020-07-22
time replaced by steps

samwaseda:
Compatibility with [this](https://pyiron.github.io/source/faq.html)


jan-janssen:
I am confused, `"time"` and `"steps"` both were in the output file before, do not you have to remove the old `"steps"` part ?

sudarsan-surendralal:
Technically, isn't time more relevant for MD simulations?

samwaseda:
I agree that time is important, the problem is just what was parsed as `time` was simply not time.

PR #786, opened 2020-07-22
Simplify accessing the project directory for script jobs

jan-janssen:


pyiron/base/project/generic.py
```diff
@@ -121,7 +124,14 @@ class Project(ProjectPath):
...
     """
 
-    def __init__(self, path="", user=None, sql_query=None):
+    def __init__(self, path="", user=None, sql_query=None, default_working_directory=False):
+        if default_working_directory:
+            inputdict = Notebook.get_custom_dict()
+            if "project_dir" in inputdict.keys():
+                path = inputdict["project_dir"]
+            else:
+                path = "."
```
samwaseda, line 133: I guess we can give a warning here, because I agree with you that this is not recommended.

samwaseda: (I mean `Project('.')` is not recommended)

pyiron/base/project/generic.py
```diff
@@ -121,7 +124,14 @@ class Project(ProjectPath):
 
     """
 
-    def __init__(self, path="", user=None, sql_query=None):
+    def __init__(self, path="", user=None, sql_query=None, default_working_directory=False):
+        if default_working_directory:
```
samwaseda, line 128: I would say `if default_working_directory and path==""`.

pyiron/base/project/generic.py
samwaseda: Maybe just to clarify: I guess if the user explicitly writes `Project(path='.')`, it's fine, because then it's likely that they know what they're doing, but if `path='.'` is automatically set because `default_working_directory` was turned on, then there's the danger that the user doesn't know what's happening.

jan-janssen: If no script job is found it at least prints that no HDF5 file is found: 
https://github.com/pyiron/pyiron/blob/script_job/pyiron/base/job/external.py#L53

jan-janssen:
@samwaseda Can you test it? Does it work for your workflow?

pyiron/base/project/generic.py
samwaseda: Okay true. Maybe that's good enough. But shouldn't it be a warning anyway?

pyiron/base/project/generic.py
```diff
@@ -121,7 +124,14 @@ class Project(ProjectPath):
 
     """
 
-    def __init__(self, path="", user=None, sql_query=None):
+    def __init__(self, path="", user=None, sql_query=None, default_working_directory=False):
+        if default_working_directory and path="":
```
samwaseda, line 128: `path==""`

pyiron/base/project/generic.py
jan-janssen: I updated it to use a warning now.

pyiron/base/project/generic.py
jan-janssen: Thx

samwaseda:
If we do

```
from pyiron import Project
Project(default_working_directory=True)
```

then shouldn't `pyiron/project.py` also be changed?

jan-janssen:
@samwaseda Did you test it?

samwaseda:
> @samwaseda Did you test it?

Yes

Issue #762, opened 2020-07-08
View structure along an axis using plot3d() 


raynol-dsouza:
I think it would be helpful (sometimes) to be able to view structures/surfaces from plot3d() along an axis, than to have to manually rotate it. @samwaseda could you maybe look into adding this feature to plot3d()?

samwaseda:
Sorry I still haven't found time but I'll surely take on this one

samwaseda:
ok this one turned out to be a bit trickier than I thought. I must be somehow related to [this function](https://github.com/arose/nglview/blob/e3b1bb0faf32be689ac4901469d9da721fb90863/nglview/viewer_control.py#L58) but I haven't really figured out how the orientation is given by a 16-dim (4-dim) array.

pmrv:
My lucky guess are [quaternions](https://en.wikipedia.org/wiki/Quaternion#Matrix_representations).  Maybe [this](https://github.com/arose/nglview/issues/596) old issue gives a clue where to look. I just skimmed it though. :|

ashtonmv:
Maybe I can help, since I had spent some time playing with this before. That flattened 4x4 matrix is just an affine spatial transformation (rotation + scaling + translation), like what they describe [here](https://www.brainvoyager.com/bv/doc/UsersGuide/CoordsAndTransforms/SpatialTransformationMatrices.html). The code below is basically what I have previously used to achieve API-based orientations.

```python
import numpy as np
import pyiron
```
```python
structure = pyiron.create_structure("Fe", "bcc", lattice_constant=2.87)
```
```python
# Just a helper function
def get_rotation_matrix(axis, theta):
    """
    Find the rotation matrix associated with counterclockwise rotation
    about the given axis by theta radians.
    Credit: http://stackoverflow.com/users/190597/unutbu
    Args:
        axis (list): rotation axis of the form [x, y, z]
        theta (float): rotational angle in radians
    Returns:
        array. 4x4 affine Rotation matrix (no translation).
    """

    axis = np.array(list(axis))
    axis = axis / np.linalg.norm(axis)
    axis *= -np.sin(theta/2.0)
    a = np.cos(theta/2.0)
    b, c, d = tuple(axis.tolist())
    aa, bb, cc, dd = a*a, b*b, c*c, d*d
    bc, ad, ac, ab, bd, cd = b*c, a*d, a*c, a*b, b*d, c*d
[...]

PR #788, opened 2020-07-22
View structure along an axis using plot3d() 

raynol-dsouza:
Fixes issue #762 

Added two flags `to structure.plot3d()`, `camera_axis` and `rotation_matrix`. 

`camera_axis` takes `'x'`, `'y'`, or `'z'` as an input to rotate the camera to view the `structure` along that axis.

The `rotation_matrix` is generated by the `_get_rotation_matrix()` helper function. More info on this can be found on Micheal's comment in #762.


pyiron/atomistics/structure/atoms.py
```diff
@@ -1710,6 +1741,32 @@ def plot3d(
 
         view.camera = camera
         view.background = background
+
+        if camera_axis != "x" and camera_axis != "y" and camera_axis != "z":
+            warnings.warn(
+                "Choose between x, y, or z"
+            )
+
+        zoom_out = 14  # higher = farther away
```
jan-janssen, line 1750: Should not this be an argument? I could see it being useful to set the zoom from the function even when no rotation is applied.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1710,6 +1741,32 @@ def plot3d(
 
         view.camera = camera
         view.background = background
+
+        if camera_axis != "x" and camera_axis != "y" and camera_axis != "z":
```
jan-janssen, line 1745: Instead of a warning it might be better to raise a value error in the else part (line 1767).

pyiron/atomistics/structure/atoms.py
```diff
@@ -1710,6 +1741,32 @@ def plot3d(
...
+            )
+
+        zoom_out = 14  # higher = farther away
+        scale = np.eye(4) * zoom_out
+        orientation = None
+
+        if rotation_matrix is not None:
+            if np.array(rotation_matrix).shape != (4, 4):
+                raise ValueError('The shape of the rotation matrix should be (4, 4)')
+            orientation = np.dot(scale, rotation_matrix).ravel().tolist()
```
jan-janssen, line 1757: Is the `tolist()` required? I would assume that NGLview also supports numpy arrays.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1499,6 +1499,31 @@ def _scalars_to_hex_colors(scalar_field, start=None, end=None, cmap=None):
             rgb2hex(cmap(scalar)[:3]) for scalar in remapped_field
         ]  # The slice gets RGB but leaves alpha
 
+    @staticmethod
+    def _get_rotation_matrix(axis, theta):
```
jan-janssen, line 1503: As somebody has to maintain this code in future I am always wondering if we can reduce the lines in the pyiron code base by using established libraries. In this case, can you do the same with https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html ?

pyiron/atomistics/structure/atoms.py
ashtonmv: I agree it's weird, but I'm pretty sure it is needed. NGLView internally serializes the list w/ JSON at some point, which won't work for a numpy array

pyiron/atomistics/structure/atoms.py
ashtonmv: scipy.spatial.transform.Rotation.from_euler seems like it would work. Just note that we need to extend the 3x3 rotation-only matrix to a 4x4 rotation + (zero) translation matrix, so

```python
from scipy.spatial.transform import Rotation
```
```python
R = np.eye(4)
R[:3, :3] = Rotation.from_euler(axis, theta).as_matrix()
```

and then flatten it.

samwaseda: This one worked:

```
from scipy.spatial.transform import Rotation
c = np.cos(theta/2)
s = np.sin(theta/2)

r = Rotation.from_quat(np.append(c, s*axis))

R = np.eye(4)
R[:3, :3] = r.as_matrix()

view.control.orient(R.ravel().tolist())
```

pyiron/atomistics/structure/atoms.py
samwaseda: I just saw indeed it doesn't work with ndarray.

pyiron/atomistics/structure/atoms.py
raynol-dsouza: Would it be better to just update` _get_rotation_matrix()` to Sam's method? Because otherwise, this snippet would need to be run in the notebook to obtain the necessary rotation matrix.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1553,6 +1580,10 @@ def plot3d(
             vector_color (numpy.ndarray): Colors for the vectors (only available with vector_field). (Default is None,
                 vectors are colored by their direction.)
             magnetic_moments (bool): Plot magnetic moments as 'scalar_field' or 'vector_field'.
+            camera_axis (str): View the structure along 'x', 'y' or 'z' axis. (Default is 'z'.)
```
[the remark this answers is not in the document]
raynol-dsouza: This can be done, yes. But I think it would be nicer to have them separated. `camera_axis` specifically sets the `orientation`. The input `rotation_matrix` will need to flattened and scaled (zoom_out) to get said `orientation`.

pyiron/atomistics/structure/atoms.py
raynol-dsouza: I will put in a `camera_zoom` flag which defaults to `None`, and then sets it to 14, if no other value is specified. I think users will be pretty confused to see this random number 14 as a default.

pyiron/atomistics/structure/atoms.py
raynol-dsouza: On second thought, I will use Micheal's suggestion. Works with fewer lines of code.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1710,6 +1741,32 @@ def plot3d(
...
+        orientation = None
+
+        if rotation_matrix is not None:
+            if np.array(rotation_matrix).shape != (4, 4):
+                raise ValueError('The shape of the rotation matrix should be (4, 4)')
+            orientation = np.dot(scale, rotation_matrix).ravel().tolist()
+        else:
+            if camera_axis == 'x':
+                rotation_matrix = self._get_rotation_matrix([0, 1, 0], np.pi / 2)
+                orientation = np.dot(scale, rotation_matrix).ravel().tolist()
```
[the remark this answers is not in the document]
raynol-dsouza: Yes, it is.

pyiron/atomistics/structure/atoms.py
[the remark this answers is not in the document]
samwaseda: My arguments:

 - They do conceptually the same thing (i.e. rotation)
 - They are not compatible with each other (one will be simply ignored when the other is set)

Then I would rather say they shouldn’t coexist. I would simply set something like `orientation` instead and internally check whether it was a string or an array.

pyiron/atomistics/structure/atoms.py
samwaseda: > On second thought, I will use Micheal's suggestion. Works with fewer lines of code.

It didn't work for me.

pyiron/atomistics/structure/atoms.py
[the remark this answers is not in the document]
raynol-dsouza: Sound like a sane argument. Shall I change `camera_axis` to `camera_orientation`, which takes axis(str) or the rotation_matrix(3, 3 array) as input?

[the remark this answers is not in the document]
samwaseda: sounds brilliant 👍

pyiron/atomistics/structure/atoms.py
raynol-dsouza: Try this:

```
from scipy.spatial.transform import Rotation

R = np.eye(4)
axis = 'x'
theta = 90  # in degrees
rotation_matrix = Rotation.from_euler(axis, theta, degrees=True).as_matrix()
R[:3, :3] = rotation_matrix

view.control.orient(R.ravel().tolist())
```

pyiron/atomistics/structure/atoms.py
```diff
@@ -1710,6 +1719,31 @@ def plot3d(
...
         view.camera = camera
         view.background = background
+
+        if camera_axis not in ['x', 'y', 'z']:
+            raise ValueError(
+                "Choose between x, y, or z"
+            )
+
+        if camera_zoom_out is None:
+            camera_zoom_out = 14
```
samwaseda, line 1729: I guess you can put this as default value (i.e. `plot3d(self, ..., camear_zoom_out=14, ...)`)

pyiron/atomistics/structure/atoms.py
samwaseda: Sorry I had erased my comment because I saw how you did it. It was my mistake. Thanks!

pyiron/atomistics/structure/atoms.py
raynol-dsouza: Updated.

samwaseda:
Following @pmrv 's suggestion yesterday (which I find was a brilliant one in hindsight), I suggest the following lines:

```
if not isinstance(camera_orientation, str) and len(np.array(camera_orientation).flatten())%3==0:
    camera_orientation = np.array(camera_orientation).reshape(-1, 3)
    rotation_matrix = np.roll(np.eye(3), 1, axis=0)
    rotation_matrix[:len(camera_orientation)] = camera_orientation
    rotation_matrix = rotation_matrix/np.linalg.norm(rotation_matrix, axis=-1)[:, np.newaxis]
    rotation_matrix[1] -= np.dot(rotation_matrix[0], rotation_matrix[1])*rotation_matrix[0]
    rotation_matrix[2] = np.cross(rotation_matrix[0], rotation_matrix[1])
    if np.isclose(np.linalg.det(rotation_matrix), 0):
        rotation_matrix = np.eye(3)
    else:
        rotation_matrix = np.roll(rotation_matrix/np.linalg.norm(rotation_matrix, axis=-1)[:, np.newaxis], 2, axis=0).T
```

And DocString has to be changed to something like:

> camera_orientation (str or numpy.ndarray): View the structure along 'x', 'y' or 'z' axis, or specify plane(-s) in Nx3-array (N=1,2,3); the first 3d-component of the array specifies the direction through screen, the second 3d-component (if specified, otherwise [1, 0, 0]) gives the horizontal direction, and the third component (if specified) is ignored and calculated internally. The orthonormality of the orientation is internally insured and therefore is not required in the function call. (Default is 'z', view along the 'z' axis)

raynol-dsouza:
@samwaseda Sorry for the delayed response. I don't think I really follow the input Nx3 array. 

From what I understood, the 1st row essentially specifies 'x', 'y' or 'z' (of input [1, 0, 0], 1 corresponding to 'x' and likewise), while the 2nd row gives the cell further rotations to align it along a given plane. I do not know what the 3rd row does.

I ran this snippet, and specifying the first row as [1, 0, 0] or [0, 0, 1] returns the same result, which is view along the 'z' axis, while [0, 1, 0] returns view along 'y' axis, as I thought it should. I thought [1, 0, 0] should return a view along the 'x' axis.

Could you maybe give a short example of how this snippet is supposed to work?

ashtonmv:
Another small note, since coveralls is complaining and we talked a lot about test coverage on friday: it would be nice to modularize the code so that as many lines as possible are covered by one or more unit tests, and then controlling the NGLWidget (which I think only exists in the Jupyter world(?)) only adds one untested line:

```python
view.control.orient(get_camera_orientation(axis, theta))
```

where `get_camera_orientation`, however you write it, should return the flat len=16 list and would get a unit test. It's probably not a bad idea to add a notebook test that tests the actual plot3d part too.

raynol-dsouza:
> 
> 
> Another small note, since coveralls is complaining and we talked a lot about test coverage on friday: it would be nice to modularize the code so that as many lines as possible are covered by one or more unit tests, and then controlling the NGLWidget (which I think only exists in the Jupyter world(?)) only adds one untested line:
> 
> ```python
> view.control.orient(get_camera_orientation(axis, theta))
> ```
> 
> where `get_camera_orientation`, however you write it, should return the flat len=16 list and would get a unit test. It's probably not a bad idea to add a notebook test that tests the actual plot3d part too.

I agree with this, and can do that. But as Jan had commented earlier, this would mean adding a new method, that does what `scipy.spatial.transform.Rotation.from_euler()` does, minus the flattening. If adding another method is okay, I will go ahead and modify this snippet.

samwaseda:
> @samwaseda Sorry for the delayed response. I don't think I really follow the input Nx3 array.
> 
> From what I understood, the 1st row essentially specifies 'x', 'y' or 'z' (of input [1, 0, 0], 1 corresponding to 'x' and likewise), while the 2nd row gives the cell further rotations to align it along a given plane. I do not know what the 3rd row does.
> 
> I ran this snippet, and specifying the first row as [1, 0, 0] or [0, 0, 1] returns the same result, which is view along the 'z' axis, while [0, 1, 0] returns view along 'y' axis, as I thought it should. I thought [1, 0, 0] should return a view along the 'x' axis.
> 
> Could you maybe give a short example of how this snippet is supposed to work?

Sorry I edited the code - now it should work again (there was only one minus to add in roll).

For example, when you put `camera_orientation=[1, 1, 1]`, you see the system in the [1, 1, 1]-plane. In this case, pyiron chooses a horizontal direction. Otherwise the user can also specify the horizontal direction by putting something like `camera_orientation=[[1, 1, 1], [1, 1, 0]]` etc. The vertical direction is then calculated internally.

But you're probably right that the DocString that I proposed was not particularly easy to understand :D

raynol-dsouza:
@samwaseda This works beautifully! I think now we might as we might as well remove the 'x', 'y', 'z' `str` inputs, as the user can specify the same using [1, 0, 0] etc. And maybe rename the flag in `plot3d` as `view_plane` instead of `camera_orientation`?

I should then also be able to put this in a modular form as @ashtonmv suggested.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1499,6 +1499,42 @@ def _scalars_to_hex_colors(scalar_field, start=None, end=None, cmap=None):
...
+
+        Returns:
+            (list): Flattened list of len = 16, which is the input argument to `view.contol.orient`
+        """
+        if len(np.array(view_plane).flatten()) % 3 != 0:
+            raise ValueError("The shape of view plane should be (N, 3), where N = 1, 2 or 3. Refer docs for more info.")
+        view_plane = np.array(view_plane).reshape(-1, 3)
+        rotation_matrix = np.roll(np.eye(3), -1, axis=0)
+        rotation_matrix[:len(view_plane)] = view_plane
+        rotation_matrix = rotation_matrix / np.linalg.norm(rotation_matrix, axis=-1)[:, np.newaxis]
```
jan-janssen, line 1525:
```suggestion
        rotation_matrix /= np.linalg.norm(rotation_matrix, axis=-1)[:, np.newaxis]
``

File: pyiron/atomistics/structure/atoms.py [hunk 9]
```diff
@@ -1553,6 +1590,14 @@ def plot3d(
             vector_color (numpy.ndarray): Colors for the vectors (only available with vector_field). (Default is None,
                 vectors are colored by their direction.)
             magnetic_moments (bool): Plot magnetic moments as 'scalar_field' or 'vector_field'.
+            view_plane (numpy.ndarray): A Nx3-array (N = 1,2,3); the first 3d-component of the array specifies
+                which plane of the system to view (for example, [1, 0, 0], [1, 1, 0] or the [1, 1, 1] planes), the
+                second 3d-component (if specified, otherwise [1, 0, 0]) gives the horizontal direction, and the third
+                component (if specified) is the vertical component, which is ignored and calculated internally. The
+                orthonormality of the orientation is internally ensured, and therefore is not required in the function
+                call. (Default is np.array([0, 0, 1]), which is view normal to the x-y plane.)
+            distance_from_camera (float): Distance of the camera from the structure. Higher = farther away.
```
samwaseda, line 1599: I see that setting `distance_from_camera=0` raises no error and NGLView shows nothing. So we could add

```
if distance_from_camera <= 0:
    raise ValueError('distance_from_camera must be a positive integer')
```

Or we could simply do:
```
if distance_from_camera <= 0:
    warnings.warn('distance_from_camera changed to 1')
    distance_from_camera = 1
```

I know this sounds stupid, but the most complicated problems often come from stupid mistakes...

samwaseda:
You can add these lines in the unit test:

```
def test_get_flattened_orientation(self):
    R = np.random.random(9).reshape(-1, 3)
    R = np.array(_get_flattened_orientation(R, 1)).reshape(4, 4)
    self.assertAlmostEqual(np.linalg.det(R), 1)
```

samwaseda:
> You can add these lines in the unit test:
> 
> ```
> def test_get_flattened_orientation(self):
>     pos, cell = generate_fcc_lattice()
>     basis = Atoms(symbols="Al", positions=pos, cell=cell)
>     R = np.random.random(9).reshape(-1, 3)
>     R = np.array(basis._get_flattened_orientation(R, 1)).reshape(4, 4)
>     self.assertAlmostEqual(np.linalg.det(R), 1)
> ```

Sorry there was an error in the lines. This should work now.

jan-janssen:
@raynol-dsouza I guess you can merge it now - congrats

PR #796, opened 2020-07-24
Validate lammps run

sudarsan-surendralal:
Check if atoms are within box (along non-periodic sides)

### DIFF
--- a/pyiron/lammps/base.py
+++ b/pyiron/lammps/base.py
@@ -212,15 +212,20 @@ class LammpsBase(AtomisticGenericJob):
 
     def validate_ready_to_run(self):
         """
-
-        Returns:
-
+        Validating input parameters before LAMMPS run
         """
         super(LammpsBase, self).validate_ready_to_run()
         if self.potential is None:
             raise ValueError(
                 "This job does not contain a valid potential: {}".format(self.job_name)
             )
+        scaled_positions = self.structure.get_scaled_positions()
+        # Check if atoms located outside of non periodic box
+        conditions = [(np.min(scaled_positions[:, i]) < 0.0 or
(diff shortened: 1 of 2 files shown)


samwaseda:
I have the feeling that this could occur due to python numerical precision (i.e. something like -1.0e-36 was set instead of 0). Wouldn't it be better if we replace the error by a warning and we perform `center_coordinates_in_unit_cell` behind the scenes? But I'd follow what everyone prefers.

In this regard I add @liamhuber

sudarsan-surendralal:
> 
> 
> I have the feeling that this could occur due to python numerical precision (i.e. something like -1.0e-36 was set instead of 0). Wouldn't it be better if we replace the error by a warning and we perform `center_coordinates_in_unit_cell` behind the scenes? But I'd follow what everyone prefers.
> 
> In this regard I add @liamhuber

I agree with the python precision part however, I don't think we should center the coordinates for a structure without pbc since it does not make sense physically (esp. for molecules).

samwaseda:
Oh sorry I didn't see that pbc was also checked. Ok then I'm fine.

pyiron/lammps/base.py
```diff
@@ -212,15 +212,20 @@ def set_input_to_read_only(self):
...
-        Returns:
-
+        Validating input parameters before LAMMPS run
         """
         super(LammpsBase, self).validate_ready_to_run()
         if self.potential is None:
             raise ValueError(
                 "This job does not contain a valid potential: {}".format(self.job_name)
             )
+        scaled_positions = self.structure.get_scaled_positions()
```
samwaseda, line 222: `get_scaled_positions(wrap=False)`

samwaseda: I mean, it's unlikely that this is required, but I guess it's safer.

sudarsan-surendralal: Done!

Issue #791, opened 2020-07-24
Float input in structure.repeat() does not work


prince-mathews:
```
Project('.').create_structure('Ni', 'fcc', a).repeat(n_repeat)
```

On using a value of n_repeat = 5.0,  the function raises an error, but using an integer value works.

jan-janssen:
I guess it is similar like an index for a list:
`[1, 2, 3, 4, 5, 6][n_repeat]`
It works with int but does not work with float. So to me that is a feature not a bug.

samwaseda:
@prince-mathews actually indeed used an integer, but `GenericParameter` for `ScriptJob` for some reason changed it to a float when the job was executed in a different notebook. I guess a safer behavior for `repeat` is that it changes the value an integer internally and maybe we can issue a warning if float was inserted or if the value was not the same before and after the conversion.

jan-janssen:
@samwaseda I guess a saver way would be to fix `GenericParameter` if that is where the error is coming from. Or even better use `inputlist` for the script job because that should already fix the issue.

samwaseda:
I’m also okay with this solution. Then for now I’d like to include an error in `repeat` that will make the user understand that a float was inserted (because the error message that @prince-mathews got was absolutely impossible to understand). If someone could do it right away, it would be nice. Otherwise I’ll do it tonight when I’m at home.

PR #799, opened 2020-07-26
error when repeat gets a float

samwaseda:
Problem addressed [here](https://github.com/pyiron/pyiron/issues/791)


jan-janssen:
I guess this was part of a rather old code change - @samwaseda thanks for fixing it -
https://github.com/pyiron/pyiron/commit/b7b4153cb75c6df86d31fa31c0e518ad53150c45

Issue #802, opened 2020-07-29
force_reset in job creation?


samwaseda:
I very often write lines like:

```
lmp = pr.create_job('Lammps', 'test')
lmp.remove()
lmp = pr.create_job('Lammps', 'test')
```

or

```
pr.remove_jobs(recursive=True)
lmp = pr.create_job('Lammps', 'test')
```

pyiron complains with the first option (and also it looks ugly) and the second option is anyway not recommended. I'd like to have something like `pr.create_job(..., force_rerun=True)`. What do you think?

jan-janssen:
I like the idea, I am just not sure if `force_rerun` is clear to everybody. What about `delete_existing_job` ?

samwaseda:
I'm fine with anything, but I think it's important that `run(run_again=True)` gets a similar name (or this new argument gets a similar name), since it does essentially the same thing (even though `run(run_again=True)` doesn't work for all job types),

jan-janssen:
Ok, we can also update `run(run_again=True)` to `run(delete_existing_job=True)`. Can you open the pull request?

samwaseda:
Another consideration: since we set `job.set_input_to_read_only()` when the job finishes, isn't it better to drop `run(run_again=True)` (or `run(delete_existing_job=True)`) altogether?

appassionate:
Hi, @jan-janssen, for the "backup job which condition is not converged"reason, I need to backup job using ".copy_to" but there seems have no method such as "reset_job".  In the "job.run" implement I followed the reset step to build a function as "reset" action.
Here wiil be a example If you are interested, Can it be a method in jobs? Thanks.

```python
def reset_job(job):
    
    if job.job_id:
        #self._logger.info("run repair " + str(self.job_id))
        
        master_id, parent_id = job.master_id, job.parent_id
        job.remove(_protect_childs=False)
        job.reset_job_id()
        job.project.set_job_status(job.job_id, "initialized")
        job.master_id, job.parent_id = master_id, parent_id
```

PR #803, opened 2020-07-29
Implement delete_existing_job in create_job

samwaseda:
Following [this discussion](https://github.com/pyiron/pyiron/issues/802)

### DIFF
--- a/pyiron/base/job/generic.py
+++ b/pyiron/base/job/generic.py
@@ -645,22 +645,24 @@ class GenericJob(JobCore):
         self._job_id = job_id
         self._status = JobStatus(db=self.project.db, job_id=self._job_id)
 
-    def run(self, run_again=False, repair=False, debug=False, run_mode=None):
+    def run(self, delete_existing_job=False, repair=False, debug=False, run_mode=None, run_again=False):
         """
         This is the main run function, depending on the job status ['initialized', 'created', 'submitted', 'running',
         'collect','finished', 'refresh', 'suspended'] the corresponding run mode is chosen.
 
         Args:
-            run_again (bool): Delete the existing job and run the simulation again.
+            delete_existing_job (bool): Delete the existing job and run the simulation again.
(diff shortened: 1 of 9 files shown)


jan-janssen:
Can you also update the `run(run_again=True)` part and maybe add a depreciation warning when people use `run_again` ?

samwaseda:
> Can you also update the `run(run_again=True)` part and maybe add a depreciation warning when people use `run_again` ?

Oh sorry I forgot this one. I'll do it tomorrow

pyiron/base/job/generic.py
```diff
@@ -656,6 +656,11 @@ def run(self, run_again=False, repair=False, debug=False, run_mode=None):
             debug (bool): Debug Mode - defines the log level of the subprocess the job is executed in.
```
jan-janssen, line 656: Can you implement `delete_existing_job here`?

pyiron/base/job/generic.py
```diff
@@ -1481,7 +1486,7 @@ def _run_if_finished(self, run_again=False):
             self.run()
```
jan-janssen, line 1486: Can you implement `delete_existing_job here`?

samwaseda: I'm not really sure if such an option should be available in `run`, because it doesn't work in the interactive mode, and we know that the changes in input won't be reflected in this case. Isn't it better to drop it altogether?

jan-janssen: But we use it, for example here https://github.com/pyiron/pyiron/blob/master/pyiron/interactive/sxextoptint.py#L359

samwaseda: But isn't it in conflict with `job.set_input_to_read_only()`? I don't understand when exactly the input can be changed and when it cannot.

jan-janssen: They basically only change the structure.

PR #804, opened 2020-07-29
Add full_table parameter

jan-janssen:


jan-janssen:
@Leimeroth  This should fix the job_table issue when using no database.

PR #805, opened 2020-07-29
update transfer_from_remote()

jan-janssen:
Requires https://github.com/pyiron/pysqa/pull/26


jan-janssen:
@Leimeroth This should at least keep the files on the remote host.

PR #833, opened 2020-08-12
check if potential supports correct elements

pmrv:
When setting a custom potential on lammps jobs via pandas dataframe, we used to accept everything.  This checks at least whether the potential advertises the correct species list for the species present in the structure.  I could also add checks for the other members as per [FAQ](https://pyiron.readthedocs.io/en/latest/source/faq.html#how-to-use-a-custom-potential-in-lammps), but I'm not 100% sure what all of them are or what downstream code expects there.


jan-janssen:
Should not this be applied to all kinds of potentials, basically below line 171? I mean in principle I can set a potential by giving the name as a string which does not support the structure.

pmrv:
True, I sort of assumed vaguely that the other code paths check this already.  Will move the check and add tests after the meeting.

PR #858, opened 2020-08-18
pyiron atoms changed to pbc when not specified

samwaseda:
I did this mainly because `get_structure()` was returning a non-periodic box


pmrv:
I thought our consensus was to keep with the ASE default in the constructor and overrride `pbc` when our functions create `Atoms` objects.  Can we pass `pbc = True` when `get_structure()` makes a new Atoms object instead?

samwaseda:
Ok I wasn't really sure about our `Atoms` class. I'm fine with changing it in `get_structure`

jan-janssen:
@samwaseda Most likely we have to update this everywhere where `get_structure()` is defined.

jan-janssen:
I found two similar issues https://github.com/pyiron/pyiron/pull/862 and https://github.com/pyiron/pyiron/pull/861

PR #863, opened 2020-08-18
FlexibleMaster: submit child jobs to queue

jan-janssen:


SanderBorgmans:
@jan-janssen it seems that in the pipeline example everything finishes, but the struct_master keeps running (and thus lstmaster stays suspended) . Do you have any idea why?

jan-janssen:
That is a bit confusing for me, in which line do you submit to the queue? I basically set:
```
job_lst_master.server.queue = "cm"
job_lst_master.run()
```
To me it seems like you set the queue for the Lammps job or something similar. While I agree that in principle both should work, I have to admit that we have not tested it so far.

SanderBorgmans:
> That is a bit confusing for me, in which line do you submit to the queue? I basically set:
> 
> ```
> job_lst_master.server.queue = "cm"
> job_lst_master.run()
> ```
> 
> To me it seems like you set the queue for the Lammps job or something similar. While I agree that in principle both should work, I have to admit that we have not tested it so far.

I am using the following code:


```
from pyiron import Project

def calc_minimize(job):
    job_mini = job.copy_to(
        new_job_name=job.job_name + '_mini', 
        new_database_entry=False
    )
    job_mini.calc_minimize(pressure=0)
    return job_mini

def calc_murnaghan(job):
    job_murn = job.copy_to(
        new_job_name=job.job_name + '_murn_lmp', 
        new_database_entry=False
    )
    return job_murn.create_job(
        job_type=pr.job_type.Murnaghan, 
        job_name=job.job_name + '_murn'
    )

def calc_phonons_for_each_structure(job):
    job_phono = job.copy_to(
        new_job_name=job.job_name + '_phono_lmp', 
        new_database_entry=False
    )
    phono = job_phono.create_job(
        job_type=pr.job_type.PhonopyJob, 
        job_name=job.job_name + '_phono'
    )
    return phono.create_job(
        job_type=pr.job_type.StructureListMaster, 
        job_name='struct_master'
    )

pr = Project('tmp')
pr.remove_jobs()
potential = 'Fe_C_Hepburn_Ackland_eam'
element = 'Fe'

structure = pr.create_ase_bulk(element, cubic=True)

job = pr.create_job(pr.job_type.Lammps, 'lmp')
job.structure = structure
job.potential = potential

job.server.queue = 'victini'
job.server.cores = 1
job.server.run_time = 5*60 # in seconds

job_lst_master = job.create_pipeline(
    step_lst=[
        calc_minimize,
        calc_murnaghan,
        calc_phonons_for_each_structure,
    ]
)

job_lst_master.run()
```

In this way, I expect that every job spawned by the pipeline through job.copy_to() gets submitted to the queue. This leaves the phonopy and structurelistmaster jobs to be modal jobs.

SanderBorgmans:
With the current code it seems that the master job waits for queue jobs to start the next, but not with modal jobs. In the notebook example, the pipeline jobs waits for the first lammps job and then the other steps are exectued in parallel. Moreover, the childcounter of the structurelistmaster does not appear to be working (it stays 0).

jan-janssen:
Yes, typical master jobs use the `update_master()` function which is called from the child job, but I am not sure if this is working for these pipeline jobs. In any way it requires a central database, which is something not many clusters support. Therefore my current idea is to submit the master as one big job to the queue and then the master executes all jobs in one queue reservation rather than submitting each job individually. Does this work for you as well?

SanderBorgmans:
> Yes, typical master jobs use the `update_master()` function which is called from the child job, but I am not sure if this is working for these pipeline jobs. In any way it requires a central database, which is something not many clusters support. Therefore my current idea is to submit the master as one big job to the queue and then the master executes all jobs in one queue reservation rather than submitting each job individually. Does this work for you as well?

I'll give this a try.

SanderBorgmans:
> > > Yes, typical master jobs use the `update_master()` function which is called from the child job, but I am not sure if this is working for these pipeline jobs. In any way it requires a central database, which is something not many clusters support. Therefore my current idea is to submit the master as one big job to the queue and then the master executes all jobs in one queue reservation rather than submitting each job individually. Does this work for you as well?
> > 
> > 
> > I'll give this a try.
> 
> It appears this is working. However, the ParallelMaster jobs are now also being executed sequentially (so the jobs spawned by the e.g. the Murnaghan job are only spawned one at the time). For the structuremasterlist, two jobs are executed at one time, the phonopy and lammps jobs, where again the lammps jobs all have the same name due to childcounter not incrementing.

The childcounter not being incremented makes sense when considering the parent_id and master_id of the jobs spawned by the pipeline. The phonopy jobs are coupled to the structuremasterlist id, while the lammps jobs are coupled to the phonopy job id, and as the phonopy job ids keep changing the childcounter is 0 for everyone of them

jan-janssen:
You could try in the functions to add something like: 
```
phono.server.run_mode.non_modal = True 
```
Then the structure master should be able to run them in parallel - but I have not tested this yet.

jan-janssen:
Currently `run_if_refresh()` directly calls `run_static()` and does not support two refresh calls by setting the status to busy. So I assume that is what causes the non modal mode to fail:
https://github.com/pyiron/pyiron/blob/master/pyiron/base/master/flexible.py#L184

jan-janssen:
@SanderBorgmans The `non_modal` mode works now, which allows the jobs at the individual stages to be executed in parallel. That's why I merge this pulls request. Still the mode you suggested where every individual calculation is submitted to the queue is not working primarily because it requires the compute nodes to be able to submit to the queuing system again, which is less flexible compared to submitting one queuing system allocation and then run the individual jobs within this allocation. Maybe you can try to run your workflow in one queuing system allocation. If this does not work feel free to open a new issue any time.

SanderBorgmans:
@jan-janssen thanks for the fixes. It can't get my local version working, but I pulled the latest pyiron version and the pipeline example work perfectly aside from the fact that all Lammps jobs get the name 'supercell_phonon_0' instead of a unique name wich could give issues

jan-janssen:
@SanderBorgmans The name is the same because it is a simple fcc cell, typically each phonopy calculation would have multiple calculation named `supercell_phonon_0`, `supercell_phonon_1`, ... `supercell_phonon_n`. Still it is correct that each phonopy calculation would start the names with `supercell_phonon_0`. The only way to differentiate the jobs is based on the project folder, as each strain gets a separate project.

SanderBorgmans:
@jan-janssen I see, I was just comparing to the example pipeline notebook, where each lammps job had a unique name based on the phonopy job name. 

Is there some function to force the struct master job to refresh its 'busy' state?

jan-janssen:
@SanderBorgmans you can simply update the job status manually: 
```
struct_master.status.refresh = True 
struct_master.run()
```

SanderBorgmans:
@jan-janssen It seems the issue is that for the Master jobs such as the Murnaghan or StructureListMaster jobs, the respective entries in the job_lst_master._job_name_lst are not popped when the all the child jobs of the Master job are complete. This seems to be causing my issues with the job status.

PR #873, opened 2020-08-20
Structurelist

pmrv:
Quick job type to store a list of structure with some meta data.  The motivation is store structures we get externally and keep track of some tags inside pyiron.

Once `InputList` can store arbitrary object in HDF5, I'd like to merge this with the `StructureContainer` that we have already. Midterm I feel this could also be a starting point for storage of experimental results.


jan-janssen:
Can you provide a notebook how to use this class? Or maybe integrate it in one of the examples like https://github.com/pyiron/pyiron/blob/master/notebooks/structures.ipynb

pmrv:
Will update the structures notebook.

jan-janssen:
Maybe you could even derive your class from the structure container used in https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/job/atomistic.py#L283 ? From my perspective your class is basically an extension or moreover a replacement for the structure container. https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/job/structurecontainer.py#L21

pmrv:
I didn't derive from `StructureContainer` as there is no shared implementation or overlapping API, even though the new class is basically a straight upgrade.  I can add a `structure` property so that the `StructureList` behaves like a `StructureContainer` and then throw that out though.

pyiron/atomistics/job/structurecontainer.py
```diff
@@ -2,62 +2,160 @@
...
-        self.__version__ = "0.1"
-        self.__name__ = "StructureContainer"
+        super().__init__(project, job_name)
+        self.__structures = InputList(table_name = "structures")
         self.server.run_mode.interactive = True
 
     @property
-    def structure(self):
-        return self._structure
+    def structures(self):
```
jan-janssen, line 37: To be a bit more explicit we decided to name the object which returns multiple structures `structure_lst()` 
https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/master/structure.py#L174

pyiron/atomistics/job/structurecontainer.py
```diff
@@ -2,62 +2,160 @@
...
+
+    @structure.setter
+    def structure(self, structure):
+        item = self.__to_structure(structure_or_job)
+        if len(self.structures) >= 1:
+            self.structures[0] = item
+        else:
+            self.structures.append(item)
+
+    def append(self, structure_or_job, **kwargs):
```
jan-janssen, line 86: I am a bit reluctant to change the signature of append, can not we require a dictionary to begin with? And when the user tries to append only a single structure create the dictionary internally?

pyiron/atomistics/job/structurecontainer.py
```diff
@@ -2,62 +2,160 @@
...
+        hdf = self.project_hdf5.create_group("structures")
+
+        for i, s in enumerate(self.__structures.values()):
+            sub = hdf.create_group("index_{}".format(i))
+            for k, v in s.items():
+                if k == "atoms":
+                    # use to_hdf from Atoms class
+                    v.to_hdf(hdf = sub, group_name = "atoms")
+                else:
+                    # otherwise use the normal hdfio implementation
```
jan-janssen, line 141: This is dangerous, because by default it creates a separate HDF5 node for each key, as our nodes have a rather large byte length we are wasting a lot of space. But you can also store the whole dictionary in HDF5 - once he structure is removed - then h5io is going to store it as a single JSON string. This is what we do for the server object 
https://github.com/pyiron/pyiron/blob/master/pyiron/base/server/generic.py#L410

pyiron/atomistics/job/structurecontainer.py
```diff
@@ -2,62 +2,160 @@
...
+        :meth:`~.StructureContainer.append()`
+        """
+        return self.__structures
 
-    @structure.setter
-    def structure(self, structure):
-        if isinstance(structure, AtomisticGenericJob):
+    @staticmethod
+    def __to_structure(structure_or_job):
+        """
```
jan-janssen, line 48: Single _ or two __ ? Currently in pyiron we use just a single one.

pmrv: Ah, just old habit, will change.

pyiron/atomistics/job/structurecontainer.py
pmrv: I thought
```python
job.append(..., foo = bar)
```
feels more concise that
```python
job.append({'atoms': ..., 'foo': bar})
```
This also requires the user to know about the correct key to store the structure under, which I feel is an implementation detail.  I could rename it to something else, though.

jan-janssen: To me:
```
job.append(..., foo = bar)
```
does not feel like python. Still we might also be able to store this description as tags for the atomistic structure or like ASE is using an "info" dictionary, we should be able to use that as well https://wiki.fysik.dtu.dk/ase/_modules/ase/atoms.html

pyiron/atomistics/job/structurecontainer.py
pmrv: My intention was to make filtering the structures before loading the whole jobs as easy as possible.

As a compromise I just tried setting the attributes of the structure group instead of creating new datasets for each tag, that reduces the final size by ~3.  Naturally the json is still ~10 smaller than this.

pyiron/atomistics/job/structurecontainer.py
pmrv: Forgot about `Atoms.info`! Will use that.

pyiron/atomistics/job/structurecontainer.py
jan-janssen: I like this idea - still loading the full dictionary is also available in inspect mode, so that should not be an issue to filter the structures.

jan-janssen:
```
ERROR: setUpClass (atomistics.job.test_StructureContainer.TestStructureContainer)
  File "/home/vsts/work/1/s/tests/atomistics/job/test_StructureContainer.py", line 23, in setUpClass
    cls.structure_container.run()
  File "/home/vsts/work/1/s/pyiron/base/job/generic.py", line 684, in run
    self._run_if_new(debug=debug)
  File "/home/vsts/work/1/s/pyiron/base/job/generic.py", line 1375, in _run_if_new
    self.save()
  File "/home/vsts/work/1/s/pyiron/base/job/generic.py", line 1201, in save
    self.to_hdf()
  File "/home/vsts/work/1/s/pyiron/atomistics/job/structurecontainer.py", line 125, in to_hdf
    GenericJob.to_hdf(self, hdf = hdf, group_name = group_name)
NameError: name 'GenericJob' is not defined

Ran 494 tests in 198.887s
```

pyiron/atomistics/job/structurecontainer.py
jan-janssen: @pmrv Can you fix this?

pmrv: Didn't see this, it's fixed now.

PR #876, opened 2020-08-20
allow abitrary object loading with `to_object()`

pmrv:
It turns out there's only small changes `to_object()` needed to allow loading any type of object from HDF5. I'll add tests & more docs, but I also wrote up a bit about how writing and reading works [here](https://github.com/pyiron/pyiron/blob/jobcore/docs/source/hdf5.rst) and I'd like your comments for the meeting tomorrow.  Naturally it's not quite complete and I'm not sure the page should mesh with the rest of the docs, but it's a start.

### DIFF
--- a/pyiron/base/generic/hdfio.py
+++ b/pyiron/base/generic/hdfio.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 import h5py
 import os
 import importlib
+import inspect
 import pandas
 import posixpath
 import h5io
@@ -1231,19 +1232,18 @@ class ProjectHDFio(FileHDFio):
         if not os.path.isdir(self.working_directory):
             os.makedirs(self.working_directory)
 
-    def create_object(self, class_name, **qwargs):
+    def import_class(self, class_name):
         """
-        Internal function to create a pyiron object
+        Import given class from fully qualified name and return class object.
 
         Args:
-            class_name (str): name of a pyiron class
-            **qwargs: object parameters
+            class_name (str): fully qualified name of a pyiron class
 
         Returns:
-            pyiron object: defined by the pyiron class in class_name with the input from **qwargs
+            class object: class object of the given name
         """
         internal_class_name = class_name.split(".")[-1][:-2]
-        if internal_class_name in self._project.job_type.job_class_dict.keys():
+        if internal_class_name in self._project.job_type.job_class_dict:
             module_path = self._project.job_type.job_class_dict[internal_class_name]
         else:
             class_path = class_name.split()[-1].split(".")[:-1]
@@ -1252,14 +1252,51 @@ class ProjectHDFio(FileHDFio):
         return getattr(
             importlib.import_module(module_path),
             internal_class_name,
-        )(**qwargs)
+        )
 
-    def to_object(self, object_type=None, **qwargs):
+    @staticmethod
+    def make_from_hdf(cls, **kwargs):
+        """
+        Create new instance of the given class from HDF5 file.
+
+        Uses the given **kwargs and a special classmethod "from_hdf_args" that
+        may be defined on cls to construct a dictionary of arguments and then
+        instatiate cls with them.
+
+        kwargs must contain all necessary arguments for `cls.__init__`, even
+        positional ones.  They will be passed by name, i.e. `cls.__init__` must
+        not have positional-only parameters to be correctly instantiated by
+        this method.
+
+        Args:
+            cls (type): pyiron type to instantiate
+            **kwargs: arguments for instance creation
+        """
+
+        # kwargs might provide more arguments than needed for this class, since
+        # it has to provide init arguments for all possible pyiron types.  Here
+        # we just pick the ones needed for this specific instance based on the
+        # signature for cls.__init__
+        if hasattr(cls, "from_hdf_args"):
+            init_args = cls.from_hdf_args(hdf)
+        else:
+            init_args = {}
+        for name, param in inspect.signature(cls.__init__).parameters.items():
+            # skip parameter if it is a varargs-style parameter *args/**kwargs
+            if param.kind in (param.VAR_POSITIONAL, param.VAR_KEYWORD):
+                continue
+
+            if name in kwargs:
+                init_args[name] = kwargs[name]
+
+        return cls(**init_args)
+
+    def to_object(self, class_name=None, **qwargs):
         """
         Load the full pyiron object from an HDF5 file
 
         Args:
-            object_type: if the 'TYPE' node is not available in the HDF5 file a manual object type can be set - optional
+            class_name: if the 'TYPE' node is not available in the HDF5 file a manual object type can be set - optional
             **qwargs: optional parameters ['job_name', 'project'] - to specify the location of the HDF5 path
 
         Returns:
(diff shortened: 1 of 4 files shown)


pyiron/base/generic/hdfio.py
```diff
@@ -1231,19 +1232,18 @@ def create_working_directory(self):
...
-            **qwargs: object parameters
+            class_name (str): fully qualified name of a pyiron class
 
         Returns:
-            pyiron object: defined by the pyiron class in class_name with the input from **qwargs
+            class object: class object of the given name
         """
         internal_class_name = class_name.split(".")[-1][:-2]
-        if internal_class_name in self._project.job_type.job_class_dict.keys():
+        if internal_class_name in self._project.job_type.job_class_dict:
```
jan-janssen, line 1246: I do not think this can work, with a dictionary

pyiron/base/generic/hdfio.py
```diff
@@ -1252,14 +1252,51 @@ def create_object(self, class_name, **qwargs):
...
+            cls (type): pyiron type to instantiate
+            **kwargs: arguments for instance creation
+        """
+
+        # kwargs might provide more arguments than needed for this class, since
+        # it has to provide init arguments for all possible pyiron types.  Here
+        # we just pick the ones needed for this specific instance based on the
+        # signature for cls.__init__
+        if hasattr(cls, "from_hdf_args"):
+            init_args = cls.from_hdf_args(hdf)
```
jan-janssen, line 1281: Currently only the job object has default arguments, while all other objects can be initialize with empty arguments and the call `from_hdf()`. I would prefer to initialize all objects with empty arguments and then load everything in the `from_hdf()`.

pyiron/base/generic/hdfio.py
```diff
@@ -1252,14 +1252,51 @@ def create_object(self, class_name, **qwargs):
         return getattr(
             importlib.import_module(module_path),
             internal_class_name,
-        )(**qwargs)
+        )
+
+    @staticmethod
+    def make_from_hdf(cls, **kwargs):
```
jan-janssen, line 1258: Maybe better -` create_instance()` - to be consistent

pyiron/base/generic/hdfio.py
```diff
@@ -1271,31 +1308,27 @@ def to_object(self, object_type=None, **qwargs):
...
-        return new_obj
+        class_name = class_name or self.get("TYPE")
+        class_object = self.import_class(class_name)
+
+        # TODO: what does that mean?
+        if class_name != str(class_object):  # Backwards compatibility
+            self["TYPE"] = str(class_object)
+
+        obj = self.make_from_hdf(class_object, **qwargs)
+        obj.from_hdf(
```
jan-janssen, line 1327: This part does not work for the `Atoms()` class

pyiron/base/generic/hdfio.py
```diff
@@ -1271,31 +1308,27 @@ def to_object(self, object_type=None, **qwargs):
...
+        elif class_name is not None and class_name != self.get("TYPE"):
+            raise ValueError(
+                "Object type in hdf5-file must be identical to input parameter"
             )
-        return new_obj
+        class_name = class_name or self.get("TYPE")
+        class_object = self.import_class(class_name)
+
+        # TODO: what does that mean?
+        if class_name != str(class_object):  # Backwards compatibility
```
jan-janssen, line 1323: When we refactored the code, the import paths stored in the HDF5 files were not updated automatically, so this basically updates the path once the class is first loaded after the refactoring.

pyiron/base/generic/hdfio.py
pmrv: Why not? Iterating over a dict iterates over the keys by default.

pyiron/base/generic/hdfio.py
pmrv: Fair, does that work for jobs or is it easy to make it work?  This is just me trying to be future-proof, I didn't have a use case in mind yet.

pyiron/base/generic/hdfio.py
pmrv: I tested it and it works when I call `Atoms.to_hdf` on a `ProjectHDFio`, then reload it and call `to_object`.  `make_from_hdf` takes care to pass `job_id` and `project` only when needed.

pmrv:
Thanks for the comments!

pyiron/base/generic/hdfio.py
jan-janssen: I just did not know about this one.

pyiron/base/generic/hdfio.py
jan-janssen: jobs get the `job_name` and `project_hdf5` as input - which could be loaded from the HDF5 file as well.

jan-janssen:
```
ERROR: test_run (atomistics.master.test_murnaghan.TestMurnaghan)
  File "/home/vsts/work/1/s/tests/atomistics/master/test_murnaghan.py", line 57, in test_run
    murn.run()
  File "/home/vsts/work/1/s/pyiron/base/job/generic.py", line 685, in run
    self._run_if_new(debug=debug)
  File "/home/vsts/work/1/s/pyiron/base/job/generic.py", line 1392, in _run_if_new
    self.run()
  File "/home/vsts/work/1/s/pyiron/base/job/generic.py", line 687, in run
    que_id = self._run_if_created()
  File "/home/vsts/work/1/s/pyiron/base/job/generic.py", line 1409, in _run_if_created
    self.run_static()
  File "/home/vsts/work/1/s/pyiron/base/master/parallel.py", line 680, in run_static
    job = next(self._job_generator, None)
  File "/home/vsts/work/1/s/pyiron/base/master/parallel.py", line 909, in __next__
    return self.next()
  File "/home/vsts/work/1/s/pyiron/base/master/parallel.py", line 924, in next
    job = self._job.create_child_job(
  File "/home/vsts/work/1/s/pyiron/base/master/parallel.py", line 819, in create_child_job
    job = self._load_all_child_jobs(job_to_load=job)
  File "/home/vsts/work/1/s/pyiron/base/master/generic.py", line 464, in _load_all_child_jobs
    job_to_load._load_job_from_cache(sub_job_name)
  File "/home/vsts/work/1/s/pyiron/base/master/generic.py", line 481, in _load_job_from_cache
    ham_obj = self.project_hdf5.create_object(
AttributeError: 'ProjectHDFio' object has no attribute 'create_object'
```

pmrv:
I had trouble understanding that one, because `create_object` is a method of `Project`, but `project_hdf5` is documented to be a `ProjectHDFio`.  Do I need to pass a `Project` or a `ProjectHDFio` to the `project parameter of a job?  How do I construct a project from a `ProjectHDFio` object?  This seems to do it

https://github.com/pyiron/pyiron/blob/33bdcd8e4dcf4c4b7b1670d6681d953196a457b1/pyiron/base/generic/hdfio.py#L1370

but then why does this seem to complicated?

https://github.com/pyiron/pyiron/blob/33bdcd8e4dcf4c4b7b1670d6681d953196a457b1/pyiron/base/generic/hdfio.py#L1271

jan-janssen:
`self.__class__` is type `ProjectHDFio` while `_create_project_from_hdf5(self)` returns a `Project`

pmrv:
Turns out I just confused `Project.create_object()` and `ProjectHDFio.create_object` (now known as `ProjectHDFio.create_instance()`).  I had to pass slightly different parameters to `to_object()` in GenericMaster than is read in `from_hdf_args()` here
https://github.com/pyiron/pyiron/blob/9759d027e64f1a45797e645b37631f86ee062098/pyiron/base/master/generic.py#L457
and I'm not quite sure why.  But if we want to cut that out later then it's not so relevant maybe.

Issue #843, opened 2020-08-14
Possibly too easy to accidentally remove_jobs(recursive=True)?


ashtonmv:
I might be the only one, but I very often have a `pr.remove_jobs(recursive=True)` statement in my notebooks that I forget to comment out, and sometimes I accidentally rerun it when I later revisit the notebook.

This is obviously a user error on my part, but I'm wondering if there's any argument against accomodating my stupidity with something like

```python
confirmed = input(f"Are you sure you want to delete all jobs from '{project_name}'? y/(n)")`
if confirmed.lower() == "y":
    # proceed to delete everything
else:
    # alert the user that nothing was deleted
```

so that the user has to confirm the deletion each time. For unit tests and users who are not as careless as I am, perhaps this safety could also be bypassed using an `enable` or `force_enable` keyword like we have in place for `pr.remove`?

Just a suggestion/question.

pmrv:
I think the idea is ok, especially since we have it on `pr.remove` and that does essentially the same. My only worry is that we'd need to update a bunch of tests & example notebooks.  On the other hand those should probably use `pr.create_job(delete_existing=True)` anyway.

jan-janssen:
I am a bit reluctant as I got used to using `pr.remove_jobs(recursive=True)` for the interactive development - basically running the the notebook with "Kernel Restart & Run All" until it produces the correct results. But once we have the ability to set the flag on a project level I am open to it. For example: 
```
pr.enable_remove_jobs = True
```
which then results in the current behaviour.

ashtonmv:
Yeah I guess it mostly only matters for people running heavier DFT calculations. So I'm happy to leave it as-is.

PR #883, opened 2020-08-21
require user input in `remove_jobs` + create `remove_jobs_silently`

ashtonmv:
As discussed in #843 and in today's meeting, I built a user input call into `Project.remove_jobs()` to confirm the removal:

```python
def remove_jobs(self, recursive=False):
    """
    Remove all jobs in the current project and in all subprojects if recursive=True is selected - see also
    remove_job().

    For safety, the user is asked via input() to confirm the removal. To bypass this
    interactive interruption, use `remove_jobs_silently()`.

    Args:
        recursive (bool): [True/False] delete all jobs in all subprojects - default=False
    """
    if not isinstance(recursive, bool):
        raise ValueError('recursive must be a boolean')
    if not self.view_mode:
        confirmed = None
        while confirmed not in ["y", "n"]:
            if confirmed is None:
                confirmed = input(
                    "Are you sure you want to delete all jobs from "
                    + f"'{self.base_name}'? y/(n)"
                    ).lower()
            else:
                confirmed = input(
                    "Invalid response. Please enter 'y' (yes) or 'n' (no): "
                    ).lower()
        if confirmed == "y":
            # proceed to delete the jobs
            for job_id in self.get_job_ids(recursive=recursive):
                if job_id not in self.get_job_ids(recursive=recursive):
                    continue
                else:
                    try:
                        self.remove_job(job_specifier=job_id)
                        s.logger.debug("Remove job with ID {0} ".format(job_id))
                    except (IndexError, Exception):
                        s.logger.debug(
                            "Could not remove job with ID {0} ".format(job_id)
                        )
        else:
            print(f"No jobs removed from '{self.base_name}'.")
    else:
        raise EnvironmentError("copy_to: is not available in Viewermode !")
```

To leave the tests and notebooks unaffected, I replaced their calls to `remove_jobs` with `remove_jobs_silently`, which does exactly what the original function did.

I added a test for the new "interactive" `remove_jobs` in test_project.py.

### DIFF
--- a/notebooks/bandstructure.ipynb
+++ b/notebooks/bandstructure.ipynb
@@ -53,7 +53,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "pr.remove_jobs(recursive=True)"
+    "pr.remove_jobs_silently(recursive=True)"
    ]
   },
   {
--- a/notebooks/data_mining.ipynb
+++ b/notebooks/data_mining.ipynb
@@ -33,7 +33,7 @@
     "pr = Project(\"potential_scan\")\n",
     "\n",
     "## Uncomment the next line if you want to remove all jobs and start again\n",
-    "# pr.remove_jobs(recursive=True)"
+    "# pr.remove_jobs_silently(recursive=True)"
    ]
   },
   {
--- a/notebooks/energy_volume_curve.ipynb
+++ b/notebooks/energy_volume_curve.ipynb
@@ -59,7 +59,7 @@
    "outputs": [],
    "source": [
     "pr = Project(path='thermo')\n",
-    "# pr.remove_jobs(recursive=True)"
+    "# pr.remove_jobs_silently(recursive=True)"
    ]
   },
   {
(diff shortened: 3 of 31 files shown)


jan-janssen (approved):
Looks good to me - feel free to merge it once the tests passed

ashtonmv:
I'm not sure what's happening with the coverage report. The 849 lines that "lost coverage" don't seem to have anything to do with what I changed, but I of course don't want to decrease the master branch's coverage. How strictly are we adhering to that build status?

sudarsan-surendralal:
> 
> 
> I'm not sure what's happening with the coverage report. The 849 lines that "lost coverage" don't seem to have anything to do with what I changed, but I of course don't want to decrease the master branch's coverage. How strictly are we adhering to that build status?

My guess is that it's because of some examples (Murnaghan examples I think) that covers areas of code depending on random initialization which could cause the decrease. @jan-janssen am I right?

jan-janssen:
The issue is that we run some tests for the non modal mode, which are executed in the background. As the order of the tests is a bit random the coverage is some times a bit fluctuating.

Issue #864, opened 2020-08-19
Trouble saving data into pyiron table


sravankumarkb:
Hello,

I have been trying to save the data from DFT calcs into a pyiron table (tot_DFT_data_table). I created the pyiron table using 

```
tot_DFT_data_table = pr.create_table("tot_DFT_data_table")
```
I get the following error at the end of the collection of my data into the table. I think that the error has to do with storing all the data into a h5 file with the table's name. I can see that all the data has been copied into the table when I print the dataframe using

```
tot_DFT_data_table.get_dataframe()
```
The status of the job is aborted. 
```
tot_DFT_data_table.status
```
I am not sure what to make of the error and if the data will load if I shutdown the notebook and open it back. Any help is greatly appreciated. Thank you.

Error
```
100%|██████████| 14784/14784 [15:09<00:00, 16.25it/s]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-358-ddb08d961dc7> in <module>
     13 
     14 # Running the table to generate the data
---> 15 tot_DFT_data_table.run(run_again = True)

/u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/site-packages/pyiron/base/job/generic.py in run(self, run_again, repair, debug, run_mode)
    673                 status = "created"
    674             if status == "initialized":
--> 675                 self._run_if_new(debug=debug)
    676             elif status == "created":
    677                 que_id = self._run_if_created()

/u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/site-packages/pyiron/base/job/generic.py in _run_if_new(self, debug)
   1350         else:
   1351             self.save()
-> 1352             self.run()
   1353 
   1354     def _run_if_created(self):

/u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/site-packages/pyiron/base/job/generic.py in run(self, run_again, repair, debug, run_mode)
    675                 self._run_if_new(debug=debug)
    676             elif status == "created":
--> 677                 que_id = self._run_if_created()
    678                 if que_id:
    679                     self._logger.info(

/u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/site-packages/pyiron/base/job/generic.py in _run_if_created(self)
   1367             self.run_if_manually()
   1368         elif self.server.run_mode.modal:
-> 1369             self.run_static()
   1370         elif self.server.run_mode.non_modal or self.server.run_mode.thread:
[...]

PR #897, opened 2020-08-25
Update CONTRIBUTING.rst

pmrv:
Since it came up in #864 and it's also part of on-boarding new devs, I added some line of how to get setup on the cluster.


jan-janssen (approved):
Looks good to me

PR #901, opened 2020-08-26
convert server.run_time to int

### LINKED ISSUE #900: Non integer runtimes break submission script
SLURM wants to have the submission time in integer minutes, setting a float silently breaks submission of job (due to #899).  We should add a type check at some point add coerce to int, I could do it in `job.server.run_time`, when passing to pysqa or in the queue template, not sure where the best location would be.

pmrv:
fixes #900


jan-janssen (approved):
Looks good to me - feel free to merge it once the tests passed

Issue #899, opened 2020-08-26
Submission errrors not caught


pmrv:
`GenericJob` checks whether submission to the queue is successful here by try/except any subprocess errors
https://github.com/pyiron/pyiron/blob/8b91247d8fb507b606e54e956655ae64f3db95ec/pyiron/base/job/generic.py#L983
but pysqa already catches those here
https://github.com/pyiron/pysqa/blob/a1792778a26be794170581e3a5d4459e219d2a82/pysqa/utils/basic.py#L422
and returns `None`.  That means any errors during submission are not caught and logged and pyiron trips a few lines later when trying to set the queue id to `None`.  Should pyiron be updated to expect `None` as failure condition or should pysqa delegate the exception?

jan-janssen:
I try to do all the queue system specific implementation inside pysqa - most likely it makes sense to have two return values, the `queue_id` and an `error.` When no error appears the `queue_id` is an integer and the `error` is `None` and when there is an error the `queue_id` is None. In `pyiron` we can check if the `queue_id` is `None`.

pmrv:
Agree that checking on the pyiron side for `None` is simplest and that the exact subprocess error should be hidden from consumers of pysqa.  Maybe `submit_job` could raise a generic `SubmissionError` when it ends up with a `None` queue id after calling its internal functions?

jan-janssen:
I guess that is something we would have to check. I am not sure if the pysqa error includes any particular information from the queueing system - why the submission failed - or if it just reports that the subprocess crashed. In particular in the case of submitting jobs to a remote host I am a bit again raising errors in pysqa and instead prefer transferring them as stings.

pmrv:
> I am not sure if the pysqa error includes any particular information from the queueing system - why the submission failed - or if it just reports that the subprocess crashed.
The subprocess error contains stderr from the queuing system command, so it's useful to have it in the log maybe.

> In particular in the case of submitting jobs to a remote host I am a bit again raising errors in pysqa and instead prefer transferring them as stings.
Fair, I'm just saying this because I had to manually submit the script from python to find out the problem in #900 and it just feels a bit hackish. If that makes problems with remote submission I understand though, given the low frequency of this type of bug, it's probably not worth the engineering cost. Will add a `None` check to pyiron later.

PR #902, opened 2020-08-26
check for None after submit_job

pmrv:
pysqa returns None now if it encounters a submission error instead of
raising an Exception

See discusion in #899.


jan-janssen (approved):
Looks good to me

PR #904, opened 2020-08-26
Increase precision in POSCAR

jan-janssen:


sudarsan-surendralal (approved):
Looks good! But I'm curious, why do we need so much precision? It it necessary for symmetry operations?

jan-janssen:
It is for the collaboration with Alexander Shapeev and Edgar Makarov.

PR #908, opened 2020-08-27
Simplify setting potentials for VASP

jan-janssen:
Use:

```
job.potential["Al"] = "Al_sv_GW"
job.potential["Ca"] = "Ca_sv_GW"
```

Or:

```
job.potential.Al = "Al_sv_GW"
job.potential.Ca = "Ca_sv_GW"
```

In addition you can list the available potentials with:
```
job.list_potentials()
```
Or get more detailed information with:
```
job.potential_view
```

Finally to change the pseudo potential use:
```
job.input.potcar["xc"] = "LDA"   # "GGA" or "LDA"
```

To mix different pseudo potential functionals you need to specify the absolute path to the corresponding potentials. Anyway mixing different pseudo potential functionals is not recommended.

### DIFF
--- a/pyiron/vasp/base.py
+++ b/pyiron/vasp/base.py
@@ -270,9 +270,12 @@ class VaspBase(GenericDFTJob):
         if self.structure is None:
             raise ValueError("Can't list potentials unless a structure is set")
         else:
-            return VaspPotentialFile(xc=self.input.potcar["xc"]).find(
+            df = VaspPotentialFile(xc=self.input.potcar["xc"]).find(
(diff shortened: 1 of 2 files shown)


sudarsan-surendralal (commented):
This looks good. However, isn't it better to implement this in `GenericDFTJob`? In principle, we can do this for SPHINX too.

jan-janssen:
I agree and I guess that is also what we discussed back in May. Still as nobody was working on this I decided to at least implement a quick fix for VASP to get started. I am definitely open towards making this more general, but that would be another pull request.

PR #911, opened 2020-08-27
Fix selective dynamics

jan-janssen:


sudarsan-surendralal (approved):
Great! I didn't realize it's this simple. Thanks!

PR #912, opened 2020-08-27
Support multiple cluster

jan-janssen:
Based on https://github.com/pyiron/pysqa/pull/37

### DIFF
--- a/pyiron/base/database/filetable.py
+++ b/pyiron/base/database/filetable.py
@@ -37,8 +37,15 @@ class FileTable(with_metaclass(Singleton)):
                          'username']
         self.force_reset()
 
+    @property
+    def viewer_mode(self):
+        return None
+
     def force_reset(self):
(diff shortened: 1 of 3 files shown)


SanderBorgmans:
@jan-janssen should the ModularQueueAdapter for the 'Gent' implementation be replaced by the RemoteQueueAdapter? The ModularQueueAdapter was initially created to be able to submit jobs on different clusters from a login node.

jan-janssen:
@SanderBorgmans I guess your setup is different as you can switch between the clusters by just loading a module, while the setup here is designed for people who ssh into different clusters and then copy all their results back to some group server. So from my perspective these are two different solutions. Still if you are interested we could also discuss using the newly implemented remote interface for your setup.

SanderBorgmans:
@jan-janssen Thanks for the explanation. I agree with your perspective then, the two designs are sufficiently different to allow different adapters to exist.

PR #917, opened 2020-08-28
Fix vasp calc

sudarsan-surendralal:
Setting the convergence precision when calling these functions


jan-janssen (approved):
Looks good to me - is set_convergence_precision() a generic function? I guess It would be great to have the same interface in Sphinx and maybe even require it for every DFT code.

sudarsan-surendralal:
> 
> 
> Looks good to me - is set_convergence_precision() a generic function? I guess It would be great to have the same interface in Sphinx and maybe even require it for every DFT code.

Yes it's already there in both VASP and SPHINX

PR #918, opened 2020-08-28
Vasp parsing

sudarsan-surendralal:
Parsing to getter better precision on the total energies (free energy). The high precision output is now stored in `job["output/genreic/dft/energy_free"]`


jan-janssen:
It is strange that the tests on Azure fail

sudarsan-surendralal:
> 
> 
> It is strange that the tests on Azure fail

I find this strange too since the precision check is the one that fails. Does the `numpy` float precision change based on environment?

```
FAIL: test_energy_pot (vasp.test_oszicar.TestOszicar)
  File "/home/vsts/work/1/s/tests/vasp/test_oszicar.py", line 38, in test_energy_pot
    self.assertTrue(np.array_equal(self.oszicar_parser.parse_dict["energy_pot"], [-17.7379867884]))
AssertionError: False is not true

```

jan-janssen:
For comparing arrays I guess it is better to use https://numpy.org/doc/stable/reference/generated/numpy.allclose.html and set the precision higher than the number of digits we have.

.ci_support/azure-unix-test-template.yml
```diff
@@ -11,7 +11,7 @@ steps:
     conda update -q conda
   displayName: conda setup
 
-- bash: .ci_support/environment.yml > environment.yml
+- bash: sed .ci_support/environment.yml > environment.yml
```
jan-janssen, line 14:
```suggestion
- bash: cp .ci_support/environment.yml environment.yml
```

jan-janssen:
It still fails - I do not really understand this ...

PR #909, opened 2020-08-27
Backwards compatible loading for StructureContainer & tests

pmrv:
The serialization format for `StructureContainer` changed due to #873 and I didn't catch that.  This bumps the HDF version for that job and adds a compatible `from_hdf`.

PR #920, opened 2020-08-29
Job core

jan-janssen:
@pmrv I reverted the pull request as they crashed the build system - I posted the error here https://github.com/pyiron/pyiron/pull/909 - as I was not sure about which part is broken I reverted both.

### DIFF
--- a/docs/source/hdf5.rst
+++ b/docs/source/hdf5.rst
@@ -1,0 +1,81 @@
+===============================
+HDF5 Serialization Architecture
+===============================
+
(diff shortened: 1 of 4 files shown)


pmrv:
I'm still not sure what caused the error last week, but I tested this PR by merging locally and running the relevant tests.  I think the error is in #921, so I'd suggest merging this (as all tests pass even after merging locally) and then I'd rebase #921 on top of master and hash out the problem there.

PR #921, opened 2020-08-29
Struct back

jan-janssen:
@pmrv I reverted the pull request as they crashed the build system - I posted the error here https://github.com/pyiron/pyiron/pull/909 - as I was not sure about which part is broken I reverted both.


pmrv:
[Merging](https://travis-ci.org/github/pyiron/pyiron/builds/725849530) works now, so I suppose the problem was in 68a5bc3.

PR #926, opened 2020-09-01
:bug: in `calc_minimize`

sudarsan-surendralal:
Also this function should not overwrite the existing electronic convergence.

### DIFF
--- a/pyiron/vasp/base.py
+++ b/pyiron/vasp/base.py
@@ -942,8 +942,29 @@ class VaspBase(GenericDFTJob):
             self.write_charge_density = retain_charge_density
         if retain_electrostatic_potential:
             self.write_electrostatic_potential = retain_electrostatic_potential
-        self.set_convergence_precision(ionic_energy_tolerance=ionic_energy_tolerance,
-                                       ionic_force_tolerance=ionic_force_tolerance)
+        if ionic_forces is not None:
+            warnings.warn(
+                "ionic_forces is deprecated as of vers. 0.3.0. It is not guaranteed to be in service in vers. 0.4.0. "
+                "Use ionic_force_tolerance instead.",
+                DeprecationWarning
+            )
+            if ionic_force_tolerance is None:
+                ionic_force_tolerance = ionic_forces
+        if ionic_energy is not None:
+            warnings.warn(
+                "ionic_energy is deprecated as of vers. 0.3.0. It is not guaranteed to be in service in vers. 0.4.0. "
+                "Use ionic_energy_tolerance instead.",
+                DeprecationWarning
+            )
+            if ionic_energy_tolerance is None:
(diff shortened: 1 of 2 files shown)


pyiron/vasp/base.py
```diff
@@ -942,8 +942,29 @@ def calc_minimize(
             self.write_charge_density = retain_charge_density
         if retain_electrostatic_potential:
             self.write_electrostatic_potential = retain_electrostatic_potential
-        self.set_convergence_precision(ionic_energy_tolerance=ionic_energy_tolerance,
-                                       ionic_force_tolerance=ionic_force_tolerance)
+        if ionic_forces is not None:
```
jan-janssen, line 945: Is there a specific reason why we do not include this part in `set_convergence_precision()` - I guess it is the same issue for all DFT codes.

sudarsan-surendralal: Because I realize that we overwrite the already set electronic convergence values if we call this.

samwaseda: I agree with @jan-janssen - I guess it's better to unify it in `set_convergence_precision` if it's anyway doing the same thing here. Why not replacing

```
self.input.incar["EDIFF"] = electronic_energy
```

by

```
if electronic_energy is not None:
    self.input.incar["EDIFF"] = electronic_energy
```

and call:

```
self.set_convergence_precision(ionic_energy_tolerance=ionic_energy_tolerance,
                                                      ionic_force_tolerance=ionic_force_tolerance,
                                                      electronic_energy=None)
```

sudarsan-surendralal: Thanks, this is definitely the more elegant solution.

pyiron/vasp/base.py
```diff
@@ -942,29 +942,10 @@ def calc_minimize(
             self.write_charge_density = retain_charge_density
         if retain_electrostatic_potential:
             self.write_electrostatic_potential = retain_electrostatic_potential
-        if ionic_forces is not None:
```
samwaseda: I guess this part should actually stay, because `ionic_forces` should be replaced by `ionic_force_tolerance` (even though personally I don't really care...)

sudarsan-surendralal: Sure but these lines are there in `set_convergence_precision`

samwaseda: So from what I understood in a recent discussion, we should replace all `ionic_forces` by `ionic_force_tolerance`. I guess it's just I missed them in `set_convergence_precision` when I did the conversion last time. But maybe @liamhuber can say more about it, because I'm personally fine with `ionic_forces` as well.

sudarsan-surendralal:
Can we merge this? I think this fixes a major bug and it's important tht we fix this quickly!

jan-janssen (approved):
Looks good to me

PR #936, opened 2020-09-03
Pipeline restart fix + non-generic master job name

SanderBorgmans:
There was a small bug in the pipeline restart functionality + it was not possible to run more than one pipeline in the project due to the non-unique name of the master job. This is fixed now by adding the template job name as a prefix.


jan-janssen (approved):
Looks good to me - feel free to merge it once the tests pass

PR #941, opened 2020-09-06
Numpy arrays

sudarsan-surendralal:


pyiron/atomistics/structure/atoms.py
```diff
@@ -1797,10 +1797,12 @@ def f_ind_ext(x):
             )
         self.min_nbr_number = min_nbr
         self.max_nbr_number = max_nbr
-        neighbor_obj.distances = self.neighbor_distance
-        neighbor_obj.vecs = self.neighbor_distance_vec
-        neighbor_obj.indices = self.neighbor_index
-        neighbor_obj.shells = self.neighbor_shellOrder
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always", np.VisibleDeprecationWarning)
```
samwaseda, line 1801: I'm not really sure if we're allowed to do this. This does not change the overall filter?

samwaseda:
I'm not even sure if it's important to `Neighbors` to check the type, since I don't think any user would use it in the first place

pyiron/atomistics/structure/atoms.py
sudarsan-surendralal: No, it only captures the warnings for the indended code section under `with ..`. Some sample code below

```
import warnings

def raise_warning():
    warnings.warn("Sample warning")

with warnings.catch_warnings(record=True) as w:
    #warnings.simplefilter("always", UserWarning)
    for _ in range(3):
        raise_warning()
```

and then in a separate cell if you type `raise_warning()`, it should raise the warning. Isn't this okay?

samwaseda: Wait I don't really understand why this example explains that the overall filter is not changed with `warnings.simplefilter` in the code above.

sudarsan-surendralal: Because calling `raise_warning` outside the with statement still raises the warning.

sudarsan-surendralal:
> 
> 
> I'm not even sure if it's important to `Neighbors` to check the type, since I don't think any user would use it in the first place

Not checking the type

> 
> 
> I'm not even sure if it's important to `Neighbors` to check the type, since I don't think any user would use it in the first place

A lot of us use it actually

pyiron/atomistics/structure/atoms.py
samwaseda: That means we don't need `warnings.simplefilter` in the end ... ? Sorry I'm probably not getting the point.

samwaseda:
> > I'm not even sure if it's important to `Neighbors` to check the type, since I don't think any user would use it in the first place
> 
> A lot of us use it actually

Really? I guess I use it only via `get_neighbors`. Maybe someone else should be reviewing your changes.

sudarsan-surendralal:
> 
> 
> > > I'm not even sure if it's important to `Neighbors` to check the type, since I don't think any user would use it in the first place
> > 
> > 
> > A lot of us use it actually
> 
> Really? I guess I use it only via `get_neighbors`. Maybe someone else should be reviewing your changes.

 Yes I too use it through `get_neighbors`. I guess you're the person who probably uses this as often as me. (So you're the best person to review this).

samwaseda:
> Yes I too use it through `get_neighbors`. I guess you're the person who probably uses this as often as me. (So you're the best person to review this).

ok but is there a particular concern that `get_neighbors` could pass illegal types?

pyiron/atomistics/structure/atoms.py
sudarsan-surendralal: This does not change any functionality. It's just that with the latest `numpy` version, every time we create an array with elements of unequal sizes, numpy complains. This is unavoidable when using `get_neighbors` which results in a lot of constant annoying warning messages (see image below). These changes capture them!

image

sudarsan-surendralal:
> 
> 
> > Yes I too use it through `get_neighbors`. I guess you're the person who probably uses this as often as me. (So you're the best person to review this).
> 
> ok but is there a particular concern that `get_neighbors` could pass illegal types?

No, but that's not the point of this pull request.  The main aim is to capture the warnings.

pyiron/atomistics/structure/atoms.py
samwaseda: No wait Sudarsan, my concern is different: Doesn't setting `warnings.simplefilter` change the global filter? @jan-janssen and I think pyiron shouldn't change the filter that user might be using.

samwaseda: For example, what happens if the user has a code like this:

```
import warnings

warnings.simplefilter('default') # filter changed to default

neighbours = structure.get_neighbors() # filter changed to always?
```

jan-janssen:
I am also against ignoring warnings, we should be able to resolve this issue.

pyiron/atomistics/structure/atoms.py
sudarsan-surendralal: > 
> 
> For example, what happens if the user has a code like this:
> 
> ```
> import warnings
> 
> warnings.simplefilter('default') # filter changed to default
> 
> neighbours = structure.get_neighbors() # filter changed to always?
> ```

The filter won't change. You can check by calling `print(warnings.filters)`. Anyway based on Jan's comment let's discuss this pull request during the pyiron meeting? It's not that critical anyway

samwaseda:
I guess with the changes made in #1001 we don't need this pull request anymore.

PR #943, opened 2020-09-07
:memo: Update water MD tutorial

niklassiemer:
The water MD notebook only used the last structure of the MD. I add short computation of a radial distribution function averaging more snapshots

### DIFF
--- a/notebooks/water_MD.ipynb
+++ b/notebooks/water_MD.ipynb
@@ -335,7 +335,7 @@
    "source": [
     "## Structure analysis\n",
     "\n",
-    "We will now use the `get_neighbors()` function to determine structural properties from the computed trajectories. We take advantage of the fact that the TIP3P water model is a rigid water model which means the neighbors of each molecule never change. Therefore they need to be indexed only once"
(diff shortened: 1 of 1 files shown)


notebooks/water_MD.ipynb
```diff
@@ -335,7 +335,7 @@
    "source": [
```
sudarsan-surendralal, line 335: <p>Thanks for extending this notebook! The only change I'd suggest is to not store the range of structures in a list as this can consume a lot of memory due to storing such objects in a list and because calling <code>get_structure()</code>  multiple times can slow down the notebook (because of multiple hdf5 access). So a better suggestion to go over the snapshots is: </p><p> </p><p><code>for pos in traj[::10]:</code></p><p><code>    struct.positions = pos</code></p><p><code>    neighbors = struct.get_neighbors(cutoff_radius=8)</code></p><p><code>    ..... # and so on</code></p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyiron/pyiron/pull/943/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

niklassiemer: Thanks a lot! Indeed, I was searching for a function to  avoid the looped get_structure() calls.

sudarsan-surendralal (approved):
Looks good to me!

PR #945, opened 2020-09-07
Vasprun fix

sudarsan-surendralal:
Vasprun parsing now works even if there are minor glitches in the vasprun output. Usually with the SCF energies


jan-janssen:
Is it possible to reduce the size of the output files?

sudarsan-surendralal:
Reduced file sizes! In fact I'd like to remove `vasprun_6.xml` since it's massive and although it represents a complex specialized simulation output (for Felix) , it is way too large (19 MB). Is this okay?

jan-janssen:
Is it possible to reduce the file with the specialised simulation output rather than removing it? Supporting magnetic calculations is I guess one of the essential features of pyiron, in particular supporting all the different kinds of magnetism VASP supports.

sudarsan-surendralal:
> 
> 
> Is it possible to reduce the file with the specialised simulation output rather than removing it? Supporting magnetic calculations is I guess one of the essential features of pyiron, in particular supporting all the different kinds of magnetism VASP supports.

It should be but I'm not aware of the specific details to generate such output. We have to wait until @FeLoch is back to do this. We can do this in a separate pull request IMO

jan-janssen:
Can not we just remove the part which is similar to the regular `vasprun.xml` files?

sudarsan-surendralal:
> 
> 
> Can not we just remove the part which is similar to the regular `vasprun.xml` files?

Not much of a difference in size dur to the regular part unfortunately (I tried).

sudarsan-surendralal:
Shall I merge this? I've opened a separate issue related to reducing file sizes: https://github.com/pyiron/pyiron/issues/953. Anyway the new files in this PR aren't that big now.

jan-janssen (approved):
Looks good to me

PR #952, opened 2020-09-09
Moving pyiron_base to a separate repository

jan-janssen:
The `pyiron.base` code is now available at https://github.com/pyiron/pyiron_base therefore we now have to import it as `pyiron_base`.


pmrv:
How are the docs build now?  I guess the current setup only builds pyiron atomistics after this is merged.  Does it complicate the deploy to readthedocs?

jan-janssen:
You are right, currently we only have documentation for pyiron not for pyiron_base. But my idea would be that we add a separate documentation for pyiron base - also based on read the docs and then link to it where it makes sense. This would allow us to build a general documentation and one focused on atomistics. Or we could also have a joined documentation.

pmrv:
Only thing I worry about is links between them, but I think there is an inter-sphinx plugin to do that.

jan-janssen:
I agree that we might face new challenges moving forward in this direction, still it is a necessary step to support an audience beyond atomistic simulation.

pmrv:
I read a bit on `namespace packages` ([here](https://packaging.python.org/guides/packaging-namespace-packages/) and [here](https://www.python.org/dev/peps/pep-0420/).  The idea is to have multiple distributions (that can live in different locations in your `PYTHONPATH`), that still contribute to the same package (and can be imported as such).  This would allow us to maintain `pyiron_base`, `pyiron_atomistic`, `pyiron_mpie` and `pyiron_contrib` as separate repos, but still import everything with just `import pyiron`, `import pyiron.base`, etc. 

Apparently it's as [easy](https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages) as dropping the topmost `__init__.py`, but I haven't had a detailed look.  Maybe we just discuss it tomorrow.

jan-janssen:
I briefly looked at their example repository https://github.com/pypa/sample-namespace-packages to me it seems it resolves the installation issue - not sure if this works with conda though - but I am not sure about the loading. As far as I remember our last discussion we want: 
```
from pyiron_atomistic import Project 
```
Gives a project with all atomistic tools while:
```
from pyiron_experimental import Project
```
Gives a project with all the tools used by experimentalists. I guess that is what we can have now. It is just `pyiron_atomistic` is currently still named `pyiron` and `pyiron_experimental` does not exist yet. But in principle `pyiron_experimental` would also be derived from `pyiron_base`.

PR #954, opened 2020-09-09
Travis fix

jan-janssen:

PR #955, opened 2020-09-09
Fixing tests

sudarsan-surendralal:
Fixing some failing tests (due to the presence of static files)


jan-janssen:
It is not the static files - I fixed it in https://github.com/pyiron/pyiron/pull/954 

It was originally related to https://github.com/pyiron/pyiron/pull/949

PR #961, opened 2020-09-10
Fix loading GenericParameters w/ to_object()

pmrv:
ProjectHDFio.to_object() calls from_hdf always with a defined
group_name, but this path was broken in GenericParameters, because the
old path expects 'data_dict' two layers below the root group (where
TYPE,NAME, etc are stored), i.e. /group_name/group_name/data_dict.

also update the HDF5 docs as to what is expected of from_hdf

### DIFF
--- a/pyiron/base/generic/parameters.py
+++ b/pyiron/base/generic/parameters.py
@@ -528,11 +528,7 @@ class GenericParameters(PyironObject):
             hdf (ProjectHDFio): HDF5 group object
             group_name (str): HDF5 subgroup name - optional
         """
-        if group_name:
-            with hdf.open(group_name) as hdf_group:
-                data = hdf_group[self.table_name]
-        else:
-            data = hdf[self.table_name]
+        data = hdf[group_name or self.table_name]
         if isinstance(data, dict):
             self._dataset = data
         else:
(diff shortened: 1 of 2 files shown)


pyiron/base/generic/parameters.py
```diff
@@ -528,11 +528,7 @@ def from_hdf(self, hdf, group_name=None):
             hdf (ProjectHDFio): HDF5 group object
             group_name (str): HDF5 subgroup name - optional
         """
-        if group_name:
-            with hdf.open(group_name) as hdf_group:
-                data = hdf_group[self.table_name]
-        else:
-            data = hdf[self.table_name]
+        data = hdf[group_name or self.table_name]
```
jan-janssen, line 531: Is this really the same? It should be `hdf[group_name + "/" + self.table_name]` or `hdf[self.table_name]`

pmrv: I'm a bit confused still how to uniformly use the group_name in to_hdf.  This (my) way is definitely broken for how it is used, but the old way doesn't work with `ProjectHDFio.to_object()`.  I hadn't had the time to think about it and it should be moved to pyiron_base anyway.

PR #964, opened 2020-09-11
Parsing bands

sudarsan-surendralal:
Now the Fermi-level, VBM, and CBM values are stored for every ionic step


sudarsan-surendralal:
Actually, I'd like to test this branch in production (to be done by @ibrsam) before we review this.

sudarsan-surendralal:
This works now in production runs!

PR #968, opened 2020-09-16
changed copy raw files option in import from path to ignore directories

Leimeroth:
importing didn't work when the directory contained other directories


jan-janssen (approved):
Looks good to me - feel free to merge it once the tests pass.

PR #970, opened 2020-09-17
Copy indices of second indices before modification

sudarsan-surendralal:


samwaseda (approved):
Perfect! Maybe we should add this case to the unit tests

PR #972, opened 2020-09-17
Append atoms

### LINKED ISSUE #951: Warning when atoms are added
I have an Fe structure to which I'd like to add one C atom:

```
job.structure = pr.create_structure('Fe', 'bcc', a_0).repeat(n_repeat)
job.structure += pr.create_atoms(positions=[[0, 0, 0.5*a_0]], elements=['C'])
```

Here, since `create_atoms` has an undefined cell (i.e. [0, 0, 0]), pyiron complains, because of the incompatibility of cell sizes, which is something we added recently. But in this case I find it not only bothering, but also potentially confusing, because adding atoms at least for me doesn't have a notion of cell.

My suggestion: if the volume of the second structure is 0 with periodic boundary conditions, we don't issue the warning. What do you think?

samwaseda:
Discussed in [this issue](https://github.com/pyiron/pyiron/issues/951). The warning is not raised if the cell is periodic with no volume and `append` is used.

Closes #951

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -2599,7 +2599,7 @@ class Atoms(ASEAtoms):
             return dist
 
     def append(self, atom):
-        if isinstance(atom, Atom) or isinstance(atom, ASEAtom):
+        if isinstance(atom, ASEAtom):
             super(Atoms, self).append(atom=atom)
         else:
             if atom.pbc.all() and np.isclose(atom.get_volume(), 0):
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -2598,6 +2598,14 @@ class Atoms(ASEAtoms):
         else:
             return dist
 
+    def append(self, atom):
+        if isinstance(atom, Atom) or isinstance(atom, ASEAtom):
+            super(Atoms, self).append(atom=atom)
+        else:
+            if atom.pbc.all() and np.isclose(atom.get_volume(), 0):
+                atom.cell = self.cell
+            self += atom
+
     def extend(self, other):
         """
         Extend atoms object by appending atoms from *other*. (Extending the ASE function)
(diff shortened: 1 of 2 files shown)


pyiron/atomistics/structure/atoms.py
```diff
@@ -2598,6 +2598,14 @@ def get_distances_array(self, a0=None, a1=None, mic=True, vector=False):
         else:
             return dist
 
+    def append(self, atom):
+        if isinstance(atom, Atom) or isinstance(atom, ASEAtom):
```
sudarsan-surendralal, line 2602: Since our `Atom` is derived from `ASEAtom`, `isinstance(atom, ASEAtom)` should be enough

pyiron/atomistics/structure/atoms.py
```diff
@@ -2598,6 +2598,14 @@ def get_distances_array(self, a0=None, a1=None, mic=True, vector=False):
...
+            super(Atoms, self).append(atom=atom)
+        else:
+            if atom.pbc.all() and np.isclose(atom.get_volume(), 0):
+                atom.cell = self.cell
```
sudarsan-surendralal, line 2606: Doesn't this change the cell of the second structure? Maybe we should copy the second structure and then change the cell

sudarsan-surendralal (changes requested):

pyiron/atomistics/structure/atoms.py
samwaseda: You mean `atom`? Changing the cell of that structure is what I want to achieve, so that's fine.

sudarsan-surendralal: > 
> 
> You mean `atom`? Changing the cell of that structure is what I want to achieve, so that's fine.

Yes but if I do `a.append(b)`, `b.cell` will now have the same value as `a.cell` when it should be 0. This shouldn't change during the operation right?

samwaseda: Very good point!

sudarsan-surendralal (approved):
Looks good!

PR #977, opened 2020-09-18
Changed kpoints per reciprocal angstrom

Leimeroth:
kpoints per reciprocal angstrom threw an error if the resulting kpoint mesh was 0 for an axis. I am not sure about the reason for this behavior, but for me it made more sense to set it to 1 kpoint in this case. If there was some special reason for throwing an error I think it would be good idea to make that behavior optional.
Also corrected the function documentation to be consistent with the actual function.

### DIFF
--- a/pyiron/dft/job/generic.py
+++ b/pyiron/dft/job/generic.py
@@ -91,7 +91,7 @@ class GenericDFTJob(AtomisticGenericJob):
         latlens = np.linalg.norm(cell, axis=-1)
         kmesh = np.rint( 2 * np.pi / latlens * kpoints_per_reciprocal_angstrom)
         if kmesh.min() <= 0:
-            print("Calculated kmesh was 0 for at least one axis, setting it to 1 instead")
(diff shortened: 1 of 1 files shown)


pyiron/dft/job/generic.py
```diff
@@ -91,7 +91,8 @@ def get_k_mesh_by_cell(self, kpoints_per_reciprocal_angstrom, cell=None):
         latlens = np.linalg.norm(cell, axis=-1)
         kmesh = np.rint( 2 * np.pi / latlens * kpoints_per_reciprocal_angstrom)
         if kmesh.min() <= 0:
-            raise AssertionError("kpoint per angstrom too low")
+            print("Calculated kmesh was 0 for at least one axis, setting it to 1 instead")
```
pmrv, line 94: I guess that should be a warning from `job.logger.warn`.

jan-janssen (approved):
Looks good to me - feel free to merge it

PR #978, opened 2020-09-18
phonopy notebook updated to replace scale function by pyiron function

samwaseda:


liamhuber (approved):

liamhuber:
Honestly, notebooks are the devil so I'm not confident I actually understand all the *changes* but the notebook as a whole is ok by me, and I did see the new scaling method and the frenglish fix.

One question actually, it looked like there is still a custom scaled method in the notebook -- did I misread and it was actually taken out? Or is it used somewhere other than structure scaling that I missed? If no to both, it should be cut. Otherwise I my happy.

notebooks/phonopy_example.ipynb
```diff
@@ -96,46 +95,8 @@
    "execution_count": 4,
    "metadata": {
     "ExecuteTime": {
-     "end_time": "2019-09-04T10:42:52.744061Z",
-     "start_time": "2019-09-04T10:42:52.733757Z"
-    }
-   },
-   "outputs": [],
-   "source": [
-    "def scale_structure(struct, scale):\n",
```
jan-janssen, line 105: @liamhuber Sam removed the scale_structure method

jan-janssen:
To simplify reviewing notebooks we recently added https://app.reviewnb.com/pyiron/pyiron/pull/978/files/

PR #983, opened 2020-09-22
change jobs names for child jobs in phonopy

### LINKED ISSUE #982: phonopy.py creates jobs with always the same name
https://github.com/pyiron/pyiron/blob/8e3e7678b50ceebd106bdd341e19d4990988f495/pyiron/atomistics/master/phonopy.py#L97

I created two phonopy jobs with different structures. These jobs have different names, but the child jobs are both called `supercell_phonon_0`, which makes it impossible to load by job name. Can I change it to something like `"supercell_{}_{}".format(self._job.job_name, ind)`?

samwaseda:
Closes #982

### DIFF
--- a/pyiron/testing/randomatomistic.py
+++ b/pyiron/testing/randomatomistic.py
@@ -7,6 +7,7 @@ import numpy as np
 import posixpath
 from pyiron_base import GenericParameters, GenericJob, Logstatus
 from pyiron.atomistics.job.interactive import GenericInteractive
+from pyiron.testing.executable import ExampleExecutable
 
 """
 Example Job class for testing the pyiron classes
@@ -120,7 +121,7 @@ class ExampleJob(GenericJob):
         self.__name__ = "ExampleJob"
         self.input = ExampleInput()
         self.executable = "python -m pyiron.testing.executable"
-        self._interactive_cache = {"alat": [], "count": [], "energy": []}
+        self.interactive_cache = {"alat": [], "count": [], "energy": []}
 
     def set_input_to_read_only(self):
         """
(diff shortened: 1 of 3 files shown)


pyiron/testing/randomatomistic.py
```diff
@@ -461,6 +461,9 @@ def from_hdf(self, hdf=None, group_name=None):
         super(AtomisticExampleJob, self).from_hdf(hdf=hdf, group_name=group_name)
         self._structure_from_hdf()
 
+    def run_static(self):
```
jan-janssen, line 464: I do not think this change belongs to this pull request.

samwaseda: That's true - now I'm going to open another pull request

samwaseda:
I don't understand why murnaghan tests started to fail...

jan-janssen:
> I don't understand why murnaghan tests started to fail...

I would assume it is related to your `run_static()` function.

samwaseda:
> I would assume it is related to your `run_static()` function.

I tested it with and without `run_static` but somehow it fails anyway...

samwaseda:
... and without `run_static`, `test_phonopy.py` fails.

jan-janssen:
The phonopy job needs forces, but the basic test job only has energy but no forces.

samwaseda:
Nice catch!!

jan-janssen:
It should be working now - but the performance is slow as phonopy struggles calculate the phonons from random forces

samwaseda:
> It should be working now - but the performance is slow as phonopy struggles calculate the phonons from random forces

Just remove `run`. I'm gonna replace it by the new Einstein class.

yury-lysogorskiy:
I mean to access subjobs, instead `project.load(sub_job_name)` do something like 
```python
j1=project['unique_master_job_name_1/supercell_phonon_0']
j2=project['unique_master_job_name_2/supercell_phonon_0']
```
in your case, that triggered this PR

samwaseda:
> I mean to access subjobs, instead `project.load(sub_job_name)` do something like
> 
> ```python
> j1=project['unique_master_job_name_1/supercell_phonon_0']
> j2=project['unique_master_job_name_2/supercell_phonon_0']
> ```
> 
> in your case, that triggered this PR

Sure, that's also possible. But that's rather a question of style. I presume you had a practical problem, right? Can we maybe talk about that?

Talking about the concept: we're anyway generally going away from the notion of Master and Child, because we start to see cases where it's not really clear what is the master/child of what & we realised that there's less flexibility with this hierarchy (like when we want to change the project location, because in this case not only you have to change the locations of jobs, but also we'd have to take care of their sub jobs). It's true that some of the oldest codes such as Murnaghan or Phonopy the job folders could be within other job folders, but it might be changed at some point. This being said, it should anyway be backward compatible, so feel free to post issues whenever it doesn't work.

PR #987, opened 2020-09-22
New example job

### LINKED ISSUE #984: Creating AtomisticExampleJob which acts more like a real job
Just a thought.

Currently `AtomisticExampleJob` does not give anything meaningful - we get always the same energy, random forces etc. This class cannot be used when we want to test for example a new energy minimizer class. How about using instead Einstein crystal? In this case we only have to calculate the displacements from the ideal positions, which means we hardly waste computational power, but still we get physically more meaningful values.

samwaseda:
Closes #984

### DIFF
--- a/pyiron/testing/randomatomistic.py
+++ b/pyiron/testing/randomatomistic.py
@@ -8,6 +8,7 @@ import posixpath
 from pyiron_base import GenericParameters, GenericJob, Logstatus
 from pyiron.atomistics.job.interactive import GenericInteractive
 from pyiron.testing.executable import ExampleExecutable
+from collections import defaultdict
 
 """
 Example Job class for testing the pyiron classes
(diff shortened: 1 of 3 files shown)


samwaseda:
A problem: The new class is not found when I do `pr.create_job('Example_job', 'job')`. How can I make it recognised?

samwaseda:
ok I see I have to update `__init__.py`

pyiron/testing/randomatomistic.py
```diff
@@ -461,6 +461,9 @@ def from_hdf(self, hdf=None, group_name=None):
         super(AtomisticExampleJob, self).from_hdf(hdf=hdf, group_name=group_name)
         self._structure_from_hdf()
 
+    def run_static(self):
```
jan-janssen, line 464: Please remove this part!

jan-janssen:
Can we use the existing Einstein Job ?
https://github.com/pyiron/pyiron_mpie/blob/master/pyiron_mpie/interactive/einstein.py

samwaseda:
> Can we use the existing Einstein Job ?
> https://github.com/pyiron/pyiron_mpie/blob/master/pyiron_mpie/interactive/einstein.py

Does this one work? I have no idea who developed this one ...

jan-janssen:
At least it has your name in it. I just want to suggest that we should have just one Einstein class. By the way I also implemented a Hessian job which takes the force constant matrix from phonopy as input https://github.com/pyiron/pyiron/blob/master/pyiron/thermodynamics/hessian.py

jan-janssen:
Is not this the same as the HessianJob just using a single force constant? I guess we could also change the set forces function: https://github.com/pyiron/pyiron/blob/master/pyiron/thermodynamics/hessian.py#L33

samwaseda:
> Is not this the same as the HessianJob just using a single force constant? I guess we could also change the set forces function: https://github.com/pyiron/pyiron/blob/master/pyiron/thermodynamics/hessian.py#L33

What does it do?

jan-janssen:
It is an harmonic oscillator just with multiple force constants. https://github.com/pyiron/pyiron/blob/master/pyiron/thermodynamics/hessian.py#L47

samwaseda:
Ok I see it's almost the same but it doesn't have pressure

jan-janssen:
That is why I suggest combining both.

jan-janssen:
Is not this already addressed in https://github.com/pyiron/pyiron/pull/989 ?

samwaseda:
> Is not this already addressed in #989 ?

The two pull requests ended up indeed being nearly the same. They just had different origins.

PR #989, opened 2020-09-22
flexible formats for force constants in HessianJob

samwaseda:
Following @jan-janssen 's comment, replace the core part of `EinstenExampleJob` by `HessianJob`

### DIFF
--- a/pyiron/thermodynamics/hessian.py
+++ b/pyiron/thermodynamics/hessian.py
@@ -22,11 +22,6 @@ class HessianJob(GenericInteractive):
         self.__version__ = "0.0.1"
         self.__name__ = "HessianJob"
         self.server.run_mode.interactive = True
-        # self.interactive_cache = {
-        #     "forces": [],
-        #     "positions": [],
-        #     "energy_pot": []
-        # }
         self._force_constants = None
         self._reference_structure = None
         self._next_positions = None
--- a/pyiron/thermodynamics/hessian.py
+++ b/pyiron/thermodynamics/hessian.py
@@ -18,7 +18,7 @@ __date__ = "Feb 20, 2020"
 
 class HessianJob(GenericInteractive):
     def __init__(self, project, job_name):
-        super(GenericInteractive, self).__init__(project, job_name)
+        super(HessianJob, self).__init__(project, job_name)
         self.__version__ = "0.0.1"
         self.__name__ = "HessianJob"
         self.server.run_mode.interactive = True
--- a/pyiron/thermodynamics/hessian.py
+++ b/pyiron/thermodynamics/hessian.py
@@ -28,6 +28,7 @@ class HessianJob(GenericInteractive):
         self._forces = np.zeros(3)
         self._pressure = np.zeros((3,3))
         self._stiffness_tensor = np.zeros((6,6))
+        self._pressure_times_volume = 0
 
     def set_elastic_moduli(self, bulk_modulus=0, shear_modulus=0):
         self._stiffness_tensor = np.zeros((6,6))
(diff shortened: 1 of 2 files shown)


jan-janssen:
I am not so much a fan of the if return pattern, should not it be just one if elif, .. else block, to be more readable.

samwaseda:
> I am not so much a fan of the if return pattern, should not it be just one if elif, .. else block, to be more readable.

I'm not either, but I couldn't come up with a better solution

samwaseda:
This class is now ready for review (sorry for setting review so early)

pyiron/thermodynamics/hessian.py
```diff
@@ -18,40 +18,117 @@
 
 class HessianJob(GenericInteractive):
     def __init__(self, project, job_name):
-        super(GenericInteractive, self).__init__(project, job_name)
+        super(HessianJob, self).__init__(project, job_name)
         self.__version__ = "0.0.1"
         self.__name__ = "HessianJob"
-        self.interactive_cache = {
```
jan-janssen, line 24: Why is this removed?

jan-janssen:
I have the feeling this pull request breaks the previous examples.

pyiron/thermodynamics/hessian.py
```diff
@@ -18,40 +18,117 @@
...
-            "forces": [],
-            "positions": [],
-            "energy_pot": []
-        }
+        self.server.run_mode.interactive = True
         self._force_constants = None
         self._reference_structure = None
-        self._next_positions = None
+        self._energy_pot = 0
```
jan-janssen, line 27: I do not understand this part.

pyiron/thermodynamics/hessian.py
samwaseda: Because since the introduction of `interactive_output_functions`, `interactive_cache` is not necessary anymore.

pyiron/thermodynamics/hessian.py
```diff
@@ -18,40 +18,117 @@
...
-        super(GenericInteractive, self).__init__(project, job_name)
+        super(HessianJob, self).__init__(project, job_name)
         self.__version__ = "0.0.1"
         self.__name__ = "HessianJob"
-        self.interactive_cache = {
-            "forces": [],
-            "positions": [],
-            "energy_pot": []
-        }
+        self.server.run_mode.interactive = True
```
jan-janssen, line 24: This part is also confusing for me.

pyiron/thermodynamics/hessian.py
```diff
@@ -18,40 +18,117 @@
...
+            self.structure = structure.copy()
+
+    def validate_ready_to_run(self):
+        super(HessianJob, self).validate_ready_to_run()
+        if self._force_constants is None:
+            raise AssertionError('set force constants by set_force_constants before run')
+        if self._reference_structure is None:
+            raise AssertionError('set reference structure by set_reference_structure before run')
+
+    def interactive_forces_getter(self):
```
jan-janssen, line 74: I do not think this is necessary if we use the interactive_cache.

samwaseda: That's true. Either way is possible. The slight advantage of defining getters is that the user can also remove some of them. But in this job probably this plays no significant role.

samwaseda:
> I have the feeling this pull request breaks the previous examples.

Which ones?

pyiron/thermodynamics/hessian.py
samwaseda: You need to put more info for me to understand what you don't understand.

pyiron/thermodynamics/hessian.py
samwaseda: Here as well.

pyiron/thermodynamics/hessian.py
jan-janssen: Why is the` interactive_cache` removed and replaced by individual variables?

pyiron/thermodynamics/hessian.py
jan-janssen: No from my perspective that is a general design decision to implement the cache for all jobs.

pyiron/thermodynamics/hessian.py
samwaseda: Because in principle we don't touch `interactive_cache` anymore to determine which quantities should be stored. Instead of creating individual variables I could also calculate the values each time, but to make it faster I did it this way.

jan-janssen:
https://github.com/pyiron/pyiron-meeting-mpie/blob/2a9c7423d4cf68f8ede6c5460e0058fce0e14ac0/meetings/2020-02-21-pyiron-meeting/jan-janssen/thermoint.ipynb

pyiron/thermodynamics/hessian.py
jan-janssen: As I said for me merging this pull request without discussing it in detail breaks examples and it breaks the design we implemented for all other interactive jobs. I am open to discuss both but then we should also adjust all other jobs.

samwaseda:
> https://github.com/pyiron/pyiron-meeting-mpie/blob/2a9c7423d4cf68f8ede6c5460e0058fce0e14ac0/meetings/2020-02-21-pyiron-meeting/jan-janssen/thermoint.ipynb

For me this already fails when I try to load `PhonopyMaster`

samwaseda:
ok it works if it's changed to `PhonopyJob`

pyiron/thermodynamics/hessian.py
samwaseda: Which design? `interactive_cache` is now not touched in any of the major codes (LAMMPS, SPHInX, VASP)

samwaseda: And I find it extremely unfair to talk about breaking examples if there's no unit test at all and no DocString explaining how the functions are supposed to be used.

PR #990, opened 2020-09-23
Discuss changes to the hessian class

jan-janssen:


jan-janssen:
Follow up to https://github.com/pyiron/pyiron/pull/989

samwaseda:
Now your example should work.

jan-janssen:
I still do not understand why you want to move away from interactive cache. I guess you made the change in the first place https://github.com/pyiron/pyiron/commit/6930fa6f8a9c2e799609de1351787ea6d27e383d

pyiron/thermodynamics/hessian.py
```diff
@@ -18,40 +18,115 @@
...
-        super(GenericInteractive, self).__init__(project, job_name)
+        super(HessianJob, self).__init__(project, job_name)
         self.__version__ = "0.0.1"
         self.__name__ = "HessianJob"
-        self.interactive_cache = {
-            "forces": [],
-            "positions": [],
-            "energy_pot": []
-        }
+        self.server.run_mode.interactive = True
```
jan-janssen, line 24: We should not enforce a given run_mode - the run_mode should be set by the user instead.

jan-janssen:
So maybe this was my mistake - I confused the old interactive cache with the newer implementation based on the default list. I adapted the example job accordingly https://github.com/pyiron/pyiron/pull/991 The only thing I do not like about this pull request is that the job is always interactive.

jan-janssen:
@samwaseda I made a couple of changes - Can you take another look? If it is fine for you we can merge the pull request once the tests pass.

PR #994, opened 2020-09-24
test murnaghan convergence

samwaseda:
With `AtomisticExampleJob`, volume convergence couldn't be seen, but this one will allow for testing convergence algorithm. 

In addition, the coverage will probably go above 60% with this pull request.


jan-janssen (approved):
Looks good to me

PR #997, opened 2020-09-24
Get global shells

samwaseda:
In the unit test I already included a Paradebeispiel (perfect example) for what `neigh.get_global_shells()` is useful for - For example, when a system is a perfect crystal with let's say one solute atom, the shell index 1 (and any other index) is not the same everywhere when we use `neigh.shells`. With `neigh.get_global_shells()`, you can be sure that wherever the same index can be found, the distance value is the same.

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -3700,7 +3700,7 @@ class Neighbors:
             decimals (int): decimals in np.round for rounding up distances
 
         Returns:
-            shells (ndarray): shell indices (cf. shells)
+            shells (numpy.ndarray): shell indices (cf. shells)
         """
         if self.distances is None:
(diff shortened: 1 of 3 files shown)


sudarsan-surendralal (commented):
While the changes look fine to me, let me try to understand the use of this function properly. Do we need this because `get_neighbors()` sets `self.shells` only considering the distance or neighbor cutoff while your function does it for the entire system?

PS: This can be a nice demo for the meeting tomorrow

samwaseda:
> While the changes look fine to me, let me try to understand the use of this function properly. Do we need this because `get_neighbors()` sets `self.shells` only considering the distance or neighbor cutoff while your function does it for the entire system?
> 
> PS: This can be a nice demo for the meeting tomorrow

So can I merge this one now?

sudarsan-surendralal (approved):

Issue #981, opened 2020-09-21
reading cell


sravankumarkb:
Hi all,

I came across the following error in one of my jobs. Upon checking the "error.out", I understand that the calculation finished and I believe the error has to do with pyiron reading the OUTCAR from VASP. More specifically, due to the lack of space between the numbers in OUTCAR. 

Error:
```
  File "/u/skanchar/software/pyiron/pyiron/vasp/outcar.py", line 842, in <listcomp>
    cell.append([float(l) for l in line.split()[0:3]])
ValueError: could not convert string to float: '0.000000000100.600000000'
```
Part of my OUTCAR: (please have a look at the z direct lattice vector)
```
  energy-cutoff  :      200.00
  volume of cell :      886.56
      direct lattice vectors                 reciprocal lattice vectors
     3.190000000  0.000000000  0.000000000     0.313479624  0.180987545  0.000000000
    -1.595000000  2.762621038  0.000000000     0.000000000  0.361975090  0.000000000
     0.000000000  0.000000000100.600000000     0.000000000  0.000000000  0.009940358
```
I think this problem will repeat if I try any calculations with big cells. Is there a way around this? Thank you.

sudarsan-surendralal:
How does the POSCAR file look line?

sravankumarkb:
beginning of the POSCAR:

```
Poscar file generated with pyiron
1.0
3.190000000000000 0.000000000000000 0.000000000000000
-1.595000000000000 2.762621038072359 0.000000000000000
0.000000000000000 -0.000000000000000 100.600000000000009
Mg
32
Selective dynamics
Cartesian
```

sravankumarkb:
If it helps, here is more verbose "error.out":

```
DAV:  12    -0.479726634030E+02   -0.40803E-07   -0.96812E-07  1976   0.342E-03
  33 F= -.47972663E+02 E0= -.47760127E+02  d E =-.623918E-03
 curvature: ****** expect dE=-0.437E-02 dE for cont linesearch -0.759E-06
 trial: gam= 0.34888 g(F)=  0.359E-04 g(S)=  0.000E+00 ort =-0.306E-05 (trialstep = 0.691E+01)
 search vector abs. value=  0.217E-03
 reached required accuracy - stopping structural energy minimisation

2020-09-21 14:54:34,661 - pyiron_log - INFO - run job: Mg_16 id: 8907771, status: collect
Traceback (most recent call last):
[...]

PR #998, opened 2020-09-24
:zap: Using scientific notation

sudarsan-surendralal:


Leimeroth:
This fixes the problem for me can we merge it?

jan-janssen:
I guess the issue is that it does not fix https://github.com/pyiron/pyiron/issues/981

Leimeroth:
Strange, I encountered exactly the same problem in some of my OUTCARs when I tried to import surface calculations

sudarsan-surendralal:
> 
> 
> Strange, I encountered exactly the same problem in some of my OUTCARs when I tried to import surface calculations

If the new branch is able to parse these calculations now then issue https://github.com/pyiron/pyiron/issues/981 is also probably fixed in my opinion. Can you confirm this?

Leimeroth:
The imported calculations had status aborted before, if I import them using your branch they have the status finished and the cell is saved correctly. My OUTCARs had the same problem with lines like this one: 0.000000000  0.000000000102.532989545

sudarsan-surendralal:
> 
> 
> The imported calculations had status aborted before, if I import them using your branch they have the status finished and the cell is saved correctly. My OUTCARs had the same problem with lines like this one: 0.000000000 0.000000000102.532989545

Ok thanks! Since this is the exact same problem @sravankumarkb had, I think we can now close this issue!

PR #1003, opened 2020-09-26
Extend get neighbors

samwaseda:
Summary:

- Distinction between global shells (based on all distances) and local shells (based on the distances of each atom) and `shells` now calls `get_local_shells` internally.
- `get_shell_matrix` now returns a list of sparse matrices (which hopefully will solve the memory issues Qais and I were encountering)
- `update_vectors` will update `vecs` and `distances` for the same indices without having to execute `get_neighbors` again. This works only if the largest distance is smaller than half the length in any direction (I'll explain why at the pyiron meeting)
- clustering methods we studied some time ago are now included in the class. They can be called as a child function from `get_local_shells` or `get_global_shells` or they can be called separately by `cluster_by_xxx` (the latter will allow for more fine adjustments of parameters). As the clustering methods could be very heavy, the results are stored and calling `get_xxx_shells` twice does not execute the algorithm again. In order to force the system to do so, you have to reset the results via `reset_clusters` or call `cluster_xxx` separately.

### DIFF
--- a/pyiron/atomistics/structure/neighbors.py
+++ b/pyiron/atomistics/structure/neighbors.py
@@ -89,18 +89,18 @@ class Neighbors(object):
         """
         if sort_by_distances:
             if self._cluster_dist is None:
-                self.sort_by_distances()
+                self.sort_shells_by_distances(use_vecs=sort_by_vecs)
             shells = [np.unique(np.round(dist, decimals=self._tolerance), return_inverse=True)[1]+1
-                         for dist in self._cluster_dist.cluster_centers_.flatten()
+                         for dist in self._cluster_dist.cluster_centers_[self._cluster_dist.labels_].flatten()
                      ]
             if isinstance(self.distances, np.ndarray):
                 return np.array(shells)
             return shells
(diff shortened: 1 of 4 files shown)


samwaseda:
Sorry this one because quite long. This class is ready now, but I will present it on Friday, so feel free to wait until then.

samwaseda:
Can I merge this now?

pyiron/atomistics/structure/neighbors.py
```diff
@@ -3,6 +3,8 @@
 # Distributed under the terms of "New BSD License", see the LICENSE file.
 
 import numpy as np
+from sklearn.cluster import MeanShift
+from scipy.sparse import coo_matrix
 from pyiron_base import Settings
 
 __author__ = "Joerg Neugebauer, Sudarsan Surendralal"
```
sudarsan-surendralal, line 10: @samwaseda Add your name here and remove mine (I didn't write any of this)

sudarsan-surendralal (approved):
To me, this looks good!

PR #1013, opened 2020-10-01
Use read-only attribute of InputList

pmrv:
This is just renaming locked -> read-only as this exists in the base class since pyiron-base=0.18


jan-janssen (approved):
Looks good to me

PR #1023, opened 2020-10-02
Sqsjob polish

liamhuber:
Done:
- `SQSJob` uses `InputList` instead of 
- Changed one input variable name (`output_structures` to `n_output_structures`)
- Cosmetic and documentation changes

**Not** done:
- Any investigation into what the `GenericInput`->`InputList` did to HDF5, outside of running and loading a job and verifying that all the output structures are there.'
- Backwards compatibility for `output_structures` (I don't think there's enough users using this to warrant it?)
- Very *helpful* docstrings for the `weights` or `objective` input parameters
- Adding unit tests


pmrv:
I'm looking at HDF5 right and likely have some other nitpicks as well.

pmrv:
There's two import guarded in a try/except, I guess this is because there are not on conda yet?

liamhuber:
> I'm looking at HDF5 right and likely have some other nitpicks as well.

Nits with these changes, or elsewhere in the sqs module? I left HDF5 as it was, and didn't notice any problems being introduced by switching over to `InputList`.

> There's two import guarded in a try/except, I guess this is because there are not on conda yet?

No, [they're](https://anaconda.org/conda-forge/pymatgen) [both](https://anaconda.org/conda-forge/sqsgenerator) in conda-forge. I'm just on my local machine and letting pycharm manage my env and (since a little while ago at least) I'm strictly on the conda-forge channel.

It sounds like your concerns might be related to the rest of the module rather than the commits in this PR, are you comfortable merging this and making a new issue(s) for other problems with SQS?

pmrv:
The only nit with these changes is the guarded imports.  If they are on conda, we should just add them as deps.

jan-janssen:
The issue is that the sqsgenerator is not available for windows, therefore it is currently no official dependence of pyiron.

jan-janssen:
The backwards compatibility would be very helpful for me, as I did a series of SQS calculation for the phase diagram calculation https://github.com/jan-janssen/NiCr-pyiron-example

pmrv:
> The backwards compatibility would be very helpful for me, as I did a series of SQS calculation for the phase diagram calculation https://github.com/jan-janssen/NiCr-pyiron-example

I will push it tomorrow, whether as part of this PR or separate doesn't matter to me.

jan-janssen:
Maybe we can add the backwards compatibility to the inputlist class directly then it should be easy to update all codes and we could raise a central depreciation warning.

pmrv:
> Maybe we can add the backwards compatibility to the inputlist class directly then it should be easy to update all codes and we could raise a central depreciation warning.

Should be possible, will cook something up, but probably not until early next week.

liamhuber:
> Maybe we can add the backwards compatibility to the inputlist class directly then it should be easy to update all codes and we could raise a central depreciation warning.

Do I understand correctly that we should thus hold off merging this until [the relevant pyiron_base](https://github.com/pyiron/pyiron_base/pull/34) PR is merged?

jan-janssen:
From my perspective implementing the backwards compatibility should not be too hard, as @pmrv illustrated here https://github.com/pyiron/pyiron_base/blob/cb03bda39f133edcb3ea71e1935812e77b139184/pyiron_base/job/script.py#L193 then we can still update it once the general backwards compatibility is available.

pyiron/atomistics/job/sqs.py
```diff
@@ -148,10 +150,30 @@ def from_hdf(self, hdf=None, group_name=None):
...
+            version = "0.1.0"
+
+        if version == "0.1.0":
+            with self.project_hdf5.open("input") as hdf5_input:
+                try:
+                    gp = GenericParameters(table_name="custom_dict")
+                    gp.from_hdf(hdf5_input)
+                    for k in gp.keys():
+                        self.input[k] = gp[k]
+                except TypeError:
```
pmrv, line 173: This is only in the ScriptJob because of the script_path member there, so we don't need it here.

liamhuber: "this" here means the `try/except` right?

samwaseda:
From the discussion above, I had sort of already guessed that the test would fail, but can you do something about it to make it? I'm now extremely unwilling to allow a file with no unit test...

jan-janssen:
The test work in general - they just fail on windows as the sqs generator is not available on windows.

pyiron/atomistics/job/sqs.py
```diff
@@ -33,30 +33,79 @@ def pymatgen_to_pyiron(structure):
...
         super(SQSJob, self).__init__(project, job_name)
-        self.input = GenericParameters(table_name="input")
-        self.input['mole_fractions'] = dict()
-        self.input['weights'] = None
-        self.input['objective'] = 0.0 
-        self.input['iterations'] = 1e6 
-        self.input['output_structures'] = 1
+        # self.input = InputList(table_name='input')
+        self.input = InputList(table_name='custom_dict')
+        self.input.mole_fractions = dict()
```
samwaseda, line 87: How about getting the mole fractions automatically if it remains undefined? Maybe you can include it in `run_static`. And then maybe throw an error if there's only one element.

liamhuber: > How about getting the mole fractions automatically if it remains undefined? Maybe you can include it in `run_static`. And then maybe throw an error if there's only one element.

What is the use case for having them automatically determined? I can only picture the following:

```python
ref = pr.create_job(pr.job_type.Vasp, 'ref')
ref.structure = pr.create_ase_bulk('Fe', cubic=True).repeat(3)
ref.structure[:7] = 'Ni'

sqs = pr.create_job(pr.job_type.SQSJob, 'sqs')
sqs.ref_job = ref
sqs.run()
```

Instead of

```python
ref = pr.create_job(pr.job_type.Vasp, 'ref')
ref.structure = pr.create_ase_bulk('Fe', cubic=True).repeat(3)

sqs = pr.create_job(pr.job_type.SQSJob, 'sqs')
sqs.ref_job = ref
sqs.input.mole_fractions = {'Fe': 0.88, 'Ni': 0.12}
sqs.run()
```

Is this enough of an improvement to warrant writing and maintaining the extra interface? I can see in some edge cases that it might be more convenient, but it's also less explicit. Overall I lean towards leaving `mole_fractions` as required input...

samwaseda: I might want to have the chemical potential of 2,3,4,.., solute atoms in solid solution for example.

tests/atomistics/job/test_sqs.py
```diff
@@ -19,9 +19,14 @@ def tearDownClass(cls):
         cls.project.remove(enable=True, enforce=True)
 
     def test_run(self):
-        self.project.create_job(
-            'SQSJob', "job_test"
-        )
+        try:  # Only test if the machine has access to sqsgenerator -- at time of writing Windows doesn't
```
jan-janssen, line 23: I do not think this is necessary unless you plan to actually run the calculation. But for running the calculation we would have to also install the SQSgenerator which is currently not done.

pmrv: Since sqsgenerator only runs on linux, I'd suggest wrapping in
```python
import sys
if sys.platform == 'linux':
  ...
```
That makes the intent clearest.

jan-janssen: No, it runs on Linux and Mac OS X

pyiron/atomistics/job/sqs.py
```diff
@@ -94,14 +94,13 @@ def __init__(self, project, job_name):
...
-    @staticmethod
-    def _fail_early_if_imports_missing():
+    def _fail_early_if_imports_missing(self):
         """
         Just a temporary measure as long as the imports are wrapped in a try/pass instead of being on the dependencies
         list.
         """
         try:
-            ParallelSqsIterator()
+            from sqsgenerator.core.sqs import ParallelSqsIterator
```
jan-janssen, line 103: Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants. https://www.python.org/dev/peps/pep-0008/#imports

tests/atomistics/job/test_sqs.py
liamhuber: > I do not think this is necessary unless you plan to actually run the calculation. 

I was actually not super happy that it didn't fail until runtime, [so I changed it to fail earlier](https://github.com/pyiron/pyiron/pull/1023/commits/5997fb705e9e8f005b4f312d23ae0fdc88bf1f2e). Thus the test needs it when the job is created.

> But for running the calculation we would have to also install the SQSgenerator which is currently not done.

You mean we don't have `sqsgenerator` on our Linux and Mac test environments? I mean, I guess I'll find out in <50m when the test passes or fails... I would like to have them though, since we expect `SQSJob` to be running on Linux and Mac environments and are testing it...

> That makes the intent clearest.

Howso? I think we don't care what system we're on, but that whatever the system is it can import the module. In this sense I find Jan's original try-except clear. (Ok, we pass the failure, which is a little unclear...I will add a comment clarifying intent!)

pyiron/atomistics/job/sqs.py
liamhuber: I think functionality trumps PEP8. I would rather import somewhere weird in order to fail early (i.e. when the job is instantiated) than wait until I try to run it.

Feel free to import another class from the same module at the top and just instantiate that class, but the actual class we import looks expensive to instantiate to me.

jan-janssen: You could also set `ParallelSqsIterator` to None in the except of the import and then check if it is None.

pyiron/atomistics/job/sqs.py
liamhuber: > I might want to have the chemical potential of 2,3,4,.., solute atoms in solid solution for example.

I'm not confident I understand -- Do you mean you want run a set of jobs in iteration? In this case I see how it's easy to set up a loop over structures, but I don't think it's so bad to do it with the dicts either. Again, I'm not trying to veto this idea, I'm just still not 100% sure I want to be eternally responsible for maintaining the interface in perpetuity 😝 

e.g.
```python
structures = []
for i in np.arange(5):
    struct[:i] = 'Ni'
    structures.append(struct.copy())
```

vs

```python
fracs = [{'Fe':(1-(i/len(struct)), 'Ni':i/len(struct)} for i in np.arange(5)]
```

pyiron/atomistics/job/sqs.py
liamhuber: Nice! I like it. Done.

pyiron/atomistics/job/sqs.py
samwaseda: For example, if I want to calculate the energy of formation for a cluster of atoms, I compare it with that of solid solution. In this case I'd compare the same number of atoms inside the cluster and in the solid solution.

samwaseda: Or I can also put it this way: If I already have a structure, I'd naïvely expect the job to figure out what to do, i.e. I'd expect something like this to work:

```
job = pr.create_job('SQSJob', 'sqs')
job.structure = structure
job.run()
```

liamhuber: No objection on my end, but I will leave it for another enhancement PR.

liamhuber: Fair enough! Any chance you want to implement it? 😉

liamhuber:
Everything passes. Either @samwaseda or I will implement his requested interface and add a test for it, then I'll merge. So if there's anything more else you want speak now or forever hold your peace :wink:

pyiron/atomistics/job/sqs.py
```diff
@@ -33,30 +36,76 @@ def pymatgen_to_pyiron(structure):
...
+    if ParallelSqsIterator is None:
+        raise ImportError("SQSJob relies on sqsgenerator.core.sqs.ParallelSqsIterator, but this is unavailable.")
+
+
 class SQSJob(AtomisticGenericJob):
+    """
+    Produces special quasirandom structures designed to duplicate truly random chemical distributions as well as
+    possible while using finite periodic cells.
+
+    A pyiron wrapper for the [SQS code of Dominik Gehringer](https://git.unileoben.ac.at/p1655622/sqsgenerator).
```
jan-janssen, line 68: I thought the official repository is https://github.com/dgehringer/sqsgenerator

pyiron/atomistics/job/sqs.py
```diff
@@ -33,30 +36,76 @@ def pymatgen_to_pyiron(structure):
...
+        - weights (list/numpy.ndarray): Specifies the weights of the individual shells. (Default is None, which uses the
+            inverse of the shell number for the weight, i.e.
+            [1, 0.5, 0.3333333333, 0.25, 0.2, 0.166666667, 0.1428571429].)
+        - objective (float): Specifies the value the objective functions. The program tries to reach the specified
+            objective function. (Default is 0.)
+        - iterations (int): How many iterations to make searching for the most special quasirandom structure using a
+            random shuffling procedure. (Default is 1e6)
+        - n_output_structures (int): How many different SQS structures to return (in decreasing special
+            quasirandomness). (Default is 1.)
+    """
```
jan-janssen, line 93: As we had the discussion about the mole_fraction it might make sense to add an example here, like Sudarsan does it for VASP: https://github.com/pyiron/pyiron/blob/master/pyiron/vasp/base.py#L55

pyiron/atomistics/job/sqs.py
```diff
@@ -33,30 +36,76 @@ def pymatgen_to_pyiron(structure):
...
+    A pyiron wrapper for the [SQS code of Dominik Gehringer](https://git.unileoben.ac.at/p1655622/sqsgenerator).
+
+    'Structural models used in calculations of properties of substitutionally random $A_{1-x}B_x$ alloys are usually
+    constructed by randomly occupying each of the N sites of a periodic cell by A or B. We show that it is possible to
+    design ‘‘special quasirandom structures’’ (SQS’s) that mimic for small N (even N=8) the first few, physically most
+    relevant radial correlation functions of a perfectly random structure far better than the standard technique does.
+    We demonstrate the usefulness of these SQS’s by calculating optical and thermodynamic properties of a number of
+    semiconductor alloys in the local-density formalism.'
+    From the abstract of Zunger, Wei, Ferreira, and Bernard, Phys. Rev. Lett. 65 (1990) 353,
+    DOI:https://doi.org/10.1103/PhysRevLett.65.353
```
jan-janssen, line 77: We have the ability to automatically collect publications which should be cited when a given module is use. Again here is an example how it is done in VASP https://github.com/pyiron/pyiron/blob/master/pyiron/vasp/base.py#L85

pyiron/atomistics/job/sqs.py
```diff
@@ -33,30 +36,76 @@ def pymatgen_to_pyiron(structure):
...
+        self.input.mole_fractions = dict()
+        self.input.weights = None
+        self.input.objective = 0.0
+        self.input.iterations = 1e6
+        self.input.n_output_structures = 1
         self._python_only_job = True
         self._lst_of_struct = []
-    
+        _fail_if_imports_missing()
+        self.__hdf_version__ = "0.2.0"
```
jan-janssen, line 107: In addition to the `__hdf_version__` we should also define a version for the SQS code wrapper https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/job/generic.py#L157 and the supported versions of the sqsgenerator https://github.com/dgehringer/sqsgenerator/tags

pyiron/atomistics/job/sqs.py
samwaseda: > Fair enough! Any chance you want to implement it? 😉

Sure :)

samwaseda:
While running the code I realized that there's nothing like energy output (at least in `job['output']`, where I could see how the penalty value decreased. Does @dgehringer's code return something like that?

pyiron/atomistics/job/sqs.py
samwaseda: done

pyiron/atomistics/job/sqs.py
dgehringer: Yeah the actual repository is the one on GitHub

dgehringer: Sorry for the ambiguity

dgehringer:
> While running the code I realized that there's nothing like energy output (at least in `job['output']`, where I could see how the penalty value decreased. Does @dgehringer's code return something like that?
actually there in `get_sqs_structures` function the `decmp` variable (which is returned) is an array of the SROs. This is maybe something you're looking for @samwaseda

pyiron/atomistics/job/sqs.py
liamhuber: Done. I am leaving the abstract also in the docstring, as it is already a better description of the physics captured in the class than I can write.

pyiron/atomistics/job/sqs.py
liamhuber: Shouldn't this be handled by the conda requirements?

liamhuber: I get that that there is no Windows-compliant version so we're not currently including this in the conda requirements, but do we want some more general way to handle this sort of situation? Tbh, it is not clear to me from the two links you gave what exactly you are looking for in this case.

liamhuber:
Thanks for getting the test all set up, @samwaseda  😄 (and other changes, but I'm particularly jazzed about the test)

liamhuber:
> > While running the code I realized that there's nothing like energy output (at least in `job['output']`, where I could see how the penalty value decreased. Does @dgehringer's code return something like that?
> 
> actually there in `get_sqs_structures` function the `decmp` variable (which is returned) is an array of the SROs. This is maybe something you're looking for @samwaseda

I guess in the future we should add some manner of output field to the docstring, but I have no objection to leaving that for a future PR -- this one has already been hanging out for almost a week

samwaseda:
> actually there in `get_sqs_structures` function the `decmp` variable (which is returned) is an array of the SROs. This is maybe something you're looking for @samwaseda

I also thought this was what I was looking for, but then I wasn't really sure how to understand it. And probably they are only the final values aren't they? It would be helpful to have them over the course of the calculation. This being said, I agree with @liamhuber that we can do something about it in a different PR.

samwaseda:
Can we actually run a job within `try:`? For this kind of class (which largely depends on external programmes), the mere fact that it runs is already quite helpful.

liamhuber:
> Can we actually run a job within `try:` in the unit tests? For this kind of class (which largely depends on external programmes), the mere fact that it runs is already quite helpful.

@samwaseda I guess we don't usually run jobs in the tests due to speed, but having an operational test somewhere (maybe in 'integration tests' or a notebook) sounds good. In fact, there is maybe already a notebook for this class kicking around?

pyiron/atomistics/job/sqs.py
jan-janssen: I mean the version that is stored in the Database, currently for python only jobs this is 0.0.1 from GenericJob and for jobs with executables the executable version is stored. Still it would be nice to use the database to filer for all old jobs so users can update them, that is why I suggested setting the version rather than just the hdf_version but I agree currently it is not consistent.

PR #1032, opened 2020-10-03
Implement S/PHI/nX potential setter

jan-janssen:
```
from pyiron import Project
pr = Project("test")
structure = pr.create_ase_bulk("Al", cubic=True)
job = pr.create_job(pr.job_type.Sphinx, "sphinx")
structure[0] = "Ca"
print(job.list_potentials())
job.potential.Al = "Al_sv_GW"
job.potential.Ca = "Ca_sv_GW"
job.run()
```

### DIFF
--- a/pyiron/sphinx/interactive.py
+++ b/pyiron/sphinx/interactive.py
@@ -10,6 +10,7 @@ import warnings
 import time
 from pyiron.sphinx.base import SphinxBase, Group
 from pyiron.atomistics.job.interactive import GenericInteractive, GenericInteractiveOutput
+from pyiron.vasp.potential import VaspPotentialSetter
 
 BOHR_TO_ANGSTROM = (
     scipy.constants.physical_constants["Bohr radius"][0] / scipy.constants.angstrom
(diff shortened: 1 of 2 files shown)


pyiron/sphinx/interactive.py
```diff
@@ -45,6 +46,10 @@ def structure(self):
     @structure.setter
     def structure(self, structure):
         GenericInteractive.structure.fset(self, structure)
+        if structure is not None:
```
samwaseda, line 49: Do we have to do this twice (in base and interactive)?

jan-janssen: The issue is that SphinxInteractive is derived from SphinxBase and GenericInteractive so we could remove it from SphinxBase as in principle nobody should use just SphinxBase but for now I did not want to break it and preferred to implement it twice.

jan-janssen:
Test it on [Binder](https://mybinder.org/v2/gh/pyiron/pyiron/sphinx_potential_setter)

PR #1050, opened 2020-10-12
Vasp parse elastic constants

Leimeroth:
This parses the total elastic constants calculated to the hdf if they were calculated using IBRION= 6 . Maybe it would be interesting to divide between clamped elastic constants and ionic contributions but I don't know how to do that in a nice style.


jan-janssen:
To give everybody the chance to give some feedback on a pull request it is recommended to keep it open for at least 24 hours.

PR #1052, opened 2020-10-12
activation relaxation technique added

samwaseda:
from pyiron_mpie


samwaseda:
Now this class is ready - as soon as the tests pass, I'll merge it.

PR #1056, opened 2020-10-14
lowered flush frequency

samwaseda:
I got a few SPHInX simulations which didn't converge within the run time limit. Then I realized that the general flush frequency was raised to 10,000 before (did I do it?) and all information stored in `interactive_cache` was lost. Now the default is set to 1 inside SPHInX.

@sudarsan-surendralal : I'm sure that the same thing should be done in VASP


jan-janssen:
I agree and I made the same changes for VASP

PR #1057, opened 2020-10-14
Get displacement

samwaseda:
Initially I just wanted to add periodic boundary conditions, but I ended up making the code a bit more compact and probably more efficient.

### DIFF
--- a/pyiron/atomistics/job/atomistic.py
+++ b/pyiron/atomistics/job/atomistic.py
@@ -865,7 +865,7 @@ class GenericOutput(object):
 
         """
         # Check if the cell changes in any snapshot
-        c = cell.reshape(-1, 9)
+        c = cells.reshape(-1, 9)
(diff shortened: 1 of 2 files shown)


samwaseda:
A few comments:

- Since the numpy shape for `cells` in `get_displacements` is anyway the same whether it contains the same information or not, I put the criterion within `get_displacements`
- Volume check was replaced by cell check, since the cell might change without changing the volume
- Since I don't think `get_displacements` has to be an extremely efficient function, I do the matrix inversion for all cells individually, which reduces the number of lines quite significantly.
- I heavily use `np.einsum` because I find very easy to understand and apparently it's also the [fastest function available ](https://stackoverflow.com/questions/18365073/why-is-numpys-einsum-faster-than-numpys-built-in-functions)

sudarsan-surendralal (approved):
The changes look good! The reason I had `varying_cell` as a parameter is to avoid checking the cells for large trajectories. But I think this is still quite quick!

PR #1058, opened 2020-10-15
Scipy minimiser updated

samwaseda:


samwaseda:
I merge it now because 1, there are only technical changes; 2, no one is probably using the class; 3, the changes are super local.

pyiron/interactive/scipy_minimizer.py
```diff
@@ -52,13 +40,6 @@ def set_input_to_read_only(self):
     def write_input(self):
         pass
 
-    def interactive_close(self):
```
jan-janssen, line 55: Why is this part removed?

samwaseda: It didn't change anything with or without this part.

jan-janssen: Is the output in the HDF5 file the same? in particular in the generic section?

samwaseda: yes

samwaseda: Actually there was anyway nothing in `generic`

jan-janssen: Ok, it seems like `interactive_is_activated()` is not even defined so this was never working ... Still the intention was that the minimiser running in interactive mode could minimize multiple structures which are given to it one by one. But I guess that is a separate pull request.

PR #1060, opened 2020-10-15
plotly added for structure plot

samwaseda:
Since NGLView doesn't give coordinates or color map, I added plotly.

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1277,6 +1277,7 @@ class Atoms(ASEAtoms):
         opacity=1,
     ):
         """
+        Make a 3D plot of the atomic structure.
 
         Args:
             camera (str): 'perspective' or 'orthographic'. (Default is 'perspective'.)
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1243,8 +1243,7 @@ class Atoms(ASEAtoms):
             return np.eye(3)  # view_plane = [0,0,1] is the default view of NGLview, so we do not modify it
         return np.roll(rotation_matrix / np.linalg.norm(rotation_matrix, axis=-1)[:, np.newaxis], 2, axis=0).T
 
-    @staticmethod
-    def _get_flattened_orientation(view_plane, distance_from_camera):
+    def _get_flattened_orientation(self, view_plane, distance_from_camera):
         """
         A helper method to plot3d, which generates a rotation matrix from the input `view_plane`, and returns a
         flattened list of len = 16. This flattened list becomes the input argument to `view.contol.orient`.
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1275,6 +1275,26 @@ class Atoms(ASEAtoms):
         view_plane=np.array([1, 1, 1]),
         distance_from_camera=1.25,
     ):
+        """
+
+        Args:
+            camera (str): 'perspective' or 'orthographic'. (Default is 'perspective'.)
+            particle_size (float): Size of the particles. (Default is 1.)
+            scalar_field (numpy.ndarray): Color each atom according to the array value (Default is None, use coloring
+                scheme.)
+            view_plane (numpy.ndarray): A Nx3-array (N = 1,2,3); the first 3d-component of the array specifies
+                which plane of the system to view (for example, [1, 0, 0], [1, 1, 0] or the [1, 1, 1] planes), the
+                second 3d-component (if specified, otherwise [1, 0, 0]) gives the horizontal direction, and the third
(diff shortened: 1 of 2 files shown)


jan-janssen:
I remember plotly was a paid service some time ago but now it seems to be free https://plotly.com/python/is-plotly-free/

jan-janssen:
Is this a draft or should we already start reviewing this pull request? Typically I do not look at drafts as I assume the author still wants to work on those before I take a look.

pmrv:
Is it possible to add the orientation of the camera according to hkl as in normal plot3d?  I assume Raynol's code is mostly independent of Nglview when it comes the geometry stuff.

liamhuber (commented):
I like it. Really liked how it looked in the meeting today too. Might have one or two small nits, but I'm pretty happy!

pyiron/atomistics/structure/atoms.py
```diff
@@ -1227,30 +1227,102 @@ def _get_flattened_orientation(view_plane, distance_from_camera):
...
+        elements = parent_basis.get_chemical_symbols()
+        atomic_numbers = parent_basis.get_atomic_numbers()
+        positions = self.positions
+        if scalar_field is None:
+            scalar_field = elements
+        fig = px.scatter_3d(x=positions[:,0],
+                            y=positions[:,1],
+                            z=positions[:,2],
+                            color=scalar_field,
+                            opacity=1,
```
liamhuber, line 1313: How does it wind up looking with opacity<1? This seems like a good candidate for a function argument. Also, is it reasonable to shoot for a per-atom opacity map?

pyiron/atomistics/structure/atoms.py
```diff
@@ -1227,30 +1227,102 @@ def _get_flattened_orientation(view_plane, distance_from_camera):
...
+        """
+        try:
+            import plotly.express as px
+        except ImportError:
+            print("plotly not installed")
+            return
+        parent_basis = self.get_parent_basis()
+        elements = parent_basis.get_chemical_symbols()
+        atomic_numbers = parent_basis.get_atomic_numbers()
+        positions = self.positions
```
liamhuber, line 1306: As far as I see, positions are only needed for the x,y,z arguments. Since these are already each on their own line, length is not critical. Rather than introducing a variable, what about just using `self.positions` down there?

pyiron/atomistics/structure/atoms.py
```diff
@@ -1227,30 +1227,102 @@ def _get_flattened_orientation(view_plane, distance_from_camera):
...
+        if scalar_field is None:
+            scalar_field = elements
+        fig = px.scatter_3d(x=positions[:,0],
```
liamhuber, line 1309: If x, y, and z are the first three function arguments, you should be able to instead use

```python
fig = px.scatter_3d(
    *self.positions.T,
    color=scalar_field,
    opacity=1,
    size=that_long_thing
)
```

Personally I like this better, but if I'm in the minority I don't particularly object to how it is now. This format goes well with my earlier comment about not defining a `positions` variable though, because then we're only calling `self.positions` once anyhow.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1227,30 +1227,102 @@ def _get_flattened_orientation(view_plane, distance_from_camera):
...
+    def plot3d_plotly(
+        self,
+        scalar_field=None,
+        particle_size=1.0,
+        camera="orthographic",
+        view_plane=np.array([1, 1, 1]),
+        distance_from_camera=1.25,
+    ):
+        """
+
```
liamhuber, line 1279: Docstring before args needs some short description like "Make a 3D plot of the atomic structure."

pyiron/atomistics/structure/atoms.py
samwaseda: If it's for example inside a notebook, I agree. But if it's inside pyiron, I prefer to explicitly write which argument was meant, because they might change the order on their side.

samwaseda: Ah and they are anyway not the first arguments

liamhuber: I haven't tried it out, but I wonder if comma assignment works inside functions, eg f(x,y,z=*vector.T). Just on my phone now, but I'll experiment later. Sadly, I suspect it doesn't.

liamhuber:
P.s. as a culture thing I don't want to set too bad an example by reviewing prs on a Friday night when all y'all should not feel like you're expected to be working--- my computer is just taking forever loading the next Star Wars:Squadrons mission so I'm killing time as productivity as possible on my phone 😝 have a good (and not all work) weekend!

pyiron/atomistics/structure/atoms.py
samwaseda: I tried, but it unfortunately didn't work

PR #1063, opened 2020-10-16
Attempt to fix issue related to using ASE atom

Leimeroth:
This fixes an issue when trying to run a gpaw job in interactive mode, that occurs because the pyiron atoms object is converted to an ASE atoms object at some point. This feels like an ugly workaround and there is probably a better fix for this so any suggestions are very welcome.

### DIFF
--- a/pyiron/atomistics/job/interactivewrapper.py
+++ b/pyiron/atomistics/job/interactivewrapper.py
@@ -7,6 +7,7 @@ import warnings
 from pyiron_base import GenericParameters, GenericJob, GenericMaster
 from pyiron.atomistics.structure.atoms import ase_to_pyiron
 from ase import Atoms as ASEAtoms
+from pyiron.atomistics.structure.atoms import Atoms as PAtoms
 
 __author__ = "Osamu Waseda, Jan Janssen"
 __copyright__ = (
@@ -164,11 +165,11 @@ class InteractiveWrapper(GenericMaster):
         """
         db_dict = super(InteractiveWrapper, self).db_entry()
         if self.structure:
-            if isinstance(self.structure, ASEAtoms):
+            if isinstance(self.structure, PAtoms):
+                parent_structure = self.structure.get_parent_basis()
+            else:
                 parent_structure = ase_to_pyiron(self.structure)
                 parent_structure = parent_structure.get_parent_basis()
-            else:
-                parent_structure = self.structure.get_parent_basis()
             db_dict["ChemicalFormula"] = parent_structure.get_chemical_formula()
         return db_dict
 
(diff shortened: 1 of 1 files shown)


jan-janssen:
As this is an ASE specific change, I would prefer to place it in one of the ASE specific classes, but I also do not see how to do it directly.

pyiron/atomistics/job/interactivewrapper.py
```diff
@@ -162,7 +164,11 @@ def db_entry(self):
         """
         db_dict = super(InteractiveWrapper, self).db_entry()
         if self.structure:
-            parent_structure = self.structure.get_parent_basis()
+            if type(self.structure) == ASEAtoms:
```
jan-janssen, line 167: Can you do the comparison like we did it here? 
https://github.com/pyiron/pyiron/blob/master/pyiron/gpaw/gpaw.py#L28

It is recommended to use `isinstance()` rather than `type()` so we check if it is an instance of our atoms class which is by now derived from the ASE atoms class.

jan-janssen:
This was confusing, I meant` isinstance` and then compare to the pyiron atoms. 

```
from pyiron.atomistics.structure.atoms import Atoms as PAtoms
....
   @structure.setter
    def structure(self, basis):
        if isinstance(basis, PAtoms):
            basis = pyiron_to_ase(basis)
        self._structure = basis
```

pyiron/atomistics/job/interactivewrapper.py
```diff
@@ -5,6 +5,9 @@
 from datetime import datetime
 import warnings
 from pyiron_base import GenericParameters, GenericJob, GenericMaster
+from pyiron.atomistics.structure.atoms import ase_to_pyiron
+from ase import Atoms as ASEAtoms
```
jan-janssen, line 9:
```suggestion
```

pyiron/atomistics/job/interactivewrapper.py
```diff
@@ -162,7 +164,11 @@ def db_entry(self):
         """
         db_dict = super(InteractiveWrapper, self).db_entry()
         if self.structure:
-            parent_structure = self.structure.get_parent_basis()
+            if isinstance(self.structure, PAtoms):
+                parent_structure = self.structure.get_parent_basis()
+            else:
+                parent_structure = ase_to_pyiron(self.structure)
```
jan-janssen, line 170: But that is just personal preference: 
```suggestion
                parent_structure = ase_to_pyiron(self.structure).get_parent_basis()
```

pyiron/atomistics/job/interactivewrapper.py
```diff
@@ -162,7 +164,11 @@ def db_entry(self):
...
+                parent_structure = parent_structure.get_parent_basis()
```
jan-janssen, line 171:
```suggestion
```

jan-janssen (approved):
Looks good to me, feel free to merge it once the tests passed.

PR #1065, opened 2020-10-16
Art name changed

samwaseda:
Following today's meeting I changed `ARTInteractive` to `ART`.


samwaseda:
Actually I don't really understand why the tests failed but with this fix it should be fine. Also the new name `ARTInteractive` for the full python class is probably more appropriate because it can really be used only as an interactive class.

PR #1072, opened 2020-10-19
Scipy

samwaseda:
Pressure minimisation


jan-janssen:
Can you share a quick example? Or even add a unit test?

samwaseda:
> Can you share a quick example? Or even add a unit test?

I'm working on the unit tests and DocString will follow

samwaseda:
Now I check the convergence on the pyiron side (instead of making `scipy` do it). Frankly, I don't know what's the best way to transfer the position data to `scipy` when the cell size is changed. Now I took the scaled positions, but we might see unexpected behaviour related to this.

PR #1073, opened 2020-10-20
Separate plot

samwaseda:
As `atoms.py` was becoming longer and longer, I separated all plot functionalities.


pyiron/atomistics/structure/atoms.py
```diff
@@ -1276,54 +1029,15 @@ def plot3d_plotly(
...
+            particle_size=particle_size,
+            camera=camera,
+            view_plane=view_plane,
+            distance_from_camera=distance_from_camera,
+            opacity=opacity,
         )
-        fig.update_layout(scene_camera=angle)
-        fig.update_traces(marker=dict(line=dict(width=0.1, color='DarkSlateGrey')))
-        return fig
+    plot3d_plotly.__doc__ = Visualization.plot3d_plotly.__doc__
```
liamhuber, line 1040: +1 for easier doc maintenance :)

pyiron/atomistics/structure/visualization.py
```diff
@@ -0,0 +1,624 @@
...
+__version__ = "1.0"
+__maintainer__ = "Sudarsan Surendralal"
+__email__ = "surendralal@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
+
+s = Settings()
+
+
+class Visualization(object):
```
liamhuber, line 25: From the github diff it's not easy to see -- is this just all the old methods newly encapsulated as the Visualization object, or do some of the methods also have changed functionality? Actually, I'll take a quick peek at the commit messages to see if that clarifies things, but I might still want confirmation.

liamhuber (commented):
I really like the idea of spinning off the visualization functionality to a separate module; I 100% agree some of our modules are getting long enough to be unwieldy. I also think the plan was cleanly executed here and can get behind merging it.

However, I want to step back for one minute and think whether there is an even better way to set up the architecture.

In my understanding, modules are for grouping related classes and methods, and classes are for grouping related methods and data. So when I look at the Visualization class I get a yellow flag when I see that it only has one piece of data and the vast majority of its methods are `@staticmethod`.

To me, it would make more sense to leave all the stuff in `visualization.py` as functions (e.g. which take as a first argument an `Atoms` instance where necessary), and then *bind* the relevant methods to the `Atoms` class at instantiation.

Below is a toy example of what I mean. For a user who has instantiated an `Atoms` object, it will look identical, but there are changes behind the scenes. In my opinion this alternative setup is shorter, cleaner, and makes more sense organizationally, but I'm actively soliciting a rebuttal here to see if anyone smells any problems!

```python
# Current
## Visualization module
class Visualization:
    def __init__(self, structure):
        self.ref_structure = structure
        
    def plotA(self):
        """PlotA docstring"""
        self._subroutine("Plotting {} in A way".format(self.ref_structure))
        
    def plotB(self):
        """PlotB docstring"""
        self._subroutine("Plotting {} in A way".format(self.ref_structure))
        
    @staticmethod
    def _subroutine(msg):
        print(msg)

## Atoms module
class Atoms:
    def __init__(self):
        self.visualization = Visualization(self)
        
    def plotA(self):
        self.visualization.plotA()
    plotA.__doc__ = Visualization.plotA.__doc__
    
    def plotB(self):
        self.visualization.plotA()
    plotB.__doc__ = Visualization.plotB.__doc__
    
    def __repr__(self):
        return "I'm a structure"
```

```python
# Operationally equivalent alternative
## Visualization module
def plotA(atoms):
    """PlotA docstring"""
    _subroutine("Plotting {} in A way".format(atoms))

def plotB(atoms):
    """PlotB docstring"""
    _subroutine("Plotting {} in A way".format(atoms))

def _subroutine(msg):
    print(msg)
  
## Atoms module
from types import MethodType
class Atoms:
    def __init__(self):
        self.plotA = MethodType(plotA, self)
        self.plotB = MethodType(plotB, self)
        
    def __repr__(self):
        return "I'm a structure"
```

The only thing I can see is that if you were trying to look at the docstrings from the class level (`Atoms.plotA?` as opposed to `atoms_instance.plotA?`) you'd run into trouble. But I don't think I've ever done this in a notebook. If it's a sufficiently rare behaviour, then the shortcoming is not important.

pyiron/atomistics/structure/visualization.py
samwaseda: Except for `ref_structure`, there shouldn't be any difference

samwaseda:
> I really like the idea of spinning off the visualization functionality to a separate module; I 100% agree some of our modules are getting long enough to be unwieldy. I also think the plan was cleanly executed here and can get behind merging it.
> 
> However, I want to step back for one minute and think whether there is an even better way to set up the architecture.
> 
> In my understanding, modules are for grouping related classes and methods, and classes are for grouping related methods and data. So when I look at the Visualization class I get a yellow flag when I see that it only has one piece of data and the vast majority of its methods are `@staticmethod`.
> 
> To me, it would make more sense to leave all the stuff in `visualization.py` as functions (e.g. which take as a first argument an `Atoms` instance where necessary), and then _bind_ the relevant methods to the `Atoms` class at instantiation.
> 
> Below is a toy example of what I mean. For a user who has instantiated an `Atoms` object, it will look identical, but there are changes behind the scenes. In my opinion this alternative setup is shorter, cleaner, and makes more sense organizationally, but I'm actively soliciting a rebuttal here to see if anyone smells any problems!
> 
> ```python
> # Current
> ## Visualization module
> class Visualization:
>     def __init__(self, structure):
>         self.ref_structure = structure
>         
>     def plotA(self):
>         """PlotA docstring"""
>         self._subroutine("Plotting {} in A way".format(self.ref_structure))
>         
>     def plotB(self):
>         """PlotB docstring"""
>         self._subroutine("Plotting {} in A way".format(self.ref_structure))
>         
>     @staticmethod
>     def _subroutine(msg):
>         print(msg)
> 
> ## Atoms module
> class Atoms:
>     def __init__(self):
>         self.visualization = Visualization(self)
>         
>     def plotA(self):
>         self.visualization.plotA()
>     plotA.__doc__ = Visualization.plotA.__doc__
>     
>     def plotB(self):
>         self.visualization.plotA()
>     plotB.__doc__ = Visualization.plotB.__doc__
>     
>     def __repr__(self):
>         return "I'm a structure"
> ```
> 
> ```python
> # Operationally equivalent alternative
> ## Visualization module
> def plotA(atoms):
>     """PlotA docstring"""
>     _subroutine("Plotting {} in A way".format(atoms))
> 
> def plotB(atoms):
>     """PlotB docstring"""
>     _subroutine("Plotting {} in A way".format(atoms))
> 
> def _subroutine(msg):
>     print(msg)
>   
> ## Atoms module
> from types import MethodType
> class Atoms:
>     def __init__(self):
>         self.plotA = MethodType(plotA, self)
>         self.plotB = MethodType(plotB, self)
>         
>     def __repr__(self):
>         return "I'm a structure"
> ```
> 
> The only thing I can see is that if you were trying to look at the docstrings from the class level (`Atoms.plotA?` as opposed to `atoms_instance.plotA?`) you'd run into trouble. But I don't think I've ever done this in a notebook. If it's a sufficiently rare behaviour, then the shortcoming is not important.

ok let me see if I can manage that

samwaseda:
> To me, it would make more sense to leave all the stuff in `visualization.py` as functions (e.g. which take as a first argument an `Atoms` instance where necessary), and then _bind_ the relevant methods to the `Atoms` class at instantiation.
> 
> Below is a toy example of what I mean. For a user who has instantiated an `Atoms` object, it will look identical, but there are changes behind the scenes. In my opinion this alternative setup is shorter, cleaner, and makes more sense organizationally, but I'm actively soliciting a rebuttal here to see if anyone smells any problems!
> 

Done!

> The only thing I can see is that if you were trying to look at the docstrings from the class level (`Atoms.plotA?` as opposed to `atoms_instance.plotA?`) you'd run into trouble. But I don't think I've ever done this in a notebook. If it's a sufficiently rare behaviour, then the shortcoming is not important.

I'd say that would be a really really rare case. In that case the user must be able to load `plotA` directly to see the DocString.

liamhuber:
> ok let me see if I can manage that

Perfect thanks! I guess that means you have no negative feedback for the idea ;)

I guess it's as easy as deleting the `Visualization` class, tabbing back its methods, replacing `self` with `atoms` in signatures and `self.ref_structure` with `atoms` in the function bodies, then replacing the re-declaration of the methods in `Atoms` with the method binding in `Atoms.__init__`.

But we could maybe wait to see if anyone has negative feedback on the whole idea before you put too much time into it...

samwaseda:
> > ok let me see if I can manage that
> 
> Perfect thanks! I guess that means you have no negative feedback for the idea ;)
> 
> I guess it's as easy as deleting the `Visualization` class, tabbing back its methods, replacing `self` with `atoms` in signatures and `self.ref_structure` with `atoms` in the function bodies, then replacing the re-declaration of the methods in `Atoms` with the method binding in `Atoms.__init__`.
> 
> But we could maybe wait to see if anyone has negative feedback on the whole idea before you put too much time into it...

It's actually already done :D

liamhuber:
> It's actually already done :D

Oops! I saw the Codacy commit and peeked at it, but totally missed the new commits above that! I somehow also totally missed an entire comment from you, I guess I took too long drafting my comment ;)

Anyhow, I guess it's now an architecture fait accompli, so we've got the high ground if anyone complains, haha.

pyiron/atomistics/structure/visualization.py
```diff
@@ -0,0 +1,604 @@
...
+        (list): Flattened list of len = 16, which is the input argument to `view.contol.orient`
+    """
+    if distance_from_camera <= 0:
+        raise ValueError("´distance_from_camera´ must be a positive float!")
+    flattened_orientation = np.eye(4)
+    flattened_orientation[:3, :3] = _get_orientation(view_plane)
+
+    return (distance_from_camera * flattened_orientation).ravel().tolist()
+
+def plot3d_plotly(
```
liamhuber, line 269: I would only suggest now to slide the `plot3d`, `plot3d_plotly` and `plot3d_ase` to the head of the module file so they're easier to find for someone who's browsing around.

liamhuber (changes requested):
Just a readability nit. Thanks for this improvement! We should consider other places this sort of idea* can be applied elsewhere.

*encapsulating functionality from too-long classes somewhere else. Maybe using method binding as here, but of course also maybe with new classes as you originally had it.

pyiron/atomistics/structure/visualization.py
samwaseda: Oh true

pyiron/atomistics/structure/visualization.py
```diff
@@ -0,0 +1,604 @@
...
+        opacity (float): opacity
+
+    Returns:
+        (plotly.express): The NGLView widget itself, which can be operated on further or viewed as-is.
+
+    """
+    try:
+        import plotly.express as px
+    except ImportError:
+        print("plotly not installed")
```
sudarsan-surendralal, line 59: Maybe a `ModuleNotFoundError` with a error message recommending `plot3d()` would be better than a print statement

pyiron/atomistics/structure/visualization.py
```diff
@@ -0,0 +1,604 @@
...
+    camera="orthographic",
+    view_plane=np.array([1, 1, 1]),
+    distance_from_camera=1.25,
+    opacity=1,
+):
+    """
+    Make a 3D plot of the atomic structure.
+
+    Args:
+        camera (str): 'perspective' or 'orthographic'. (Default is 'perspective'.)
```
sudarsan-surendralal, line 38: I think you are missing the `atoms` argument in the docstring. (Also found elsewhere)

sudarsan-surendralal (approved):
Looks good to me. Just a few minor suggestions

pyiron/atomistics/structure/visualization.py
liamhuber: This is a bit intentional though, as the `atoms` argument is just `self` when the method gets bound to the `Atoms` class instance. I thought there was a nice middle ground that the plain english sentence at the head of the docstring makes the context of the `atoms` argument pretty clear. We maybe need something even better than this though... Perhaps something at the module level documentation of `visualization.py` giving a generic definition of the `atoms` argument and explaining the intent of binding these methods?

liamhuber (approved):

pyiron/atomistics/structure/visualization.py
samwaseda: Oh is `ModuleNotFoundError` raised when it's not found?

sudarsan-surendralal: Yes. `ImportError` is raised if that package is there but the module you want to import isn't. Here if plotly is not there a `ModuleNotFoundError` is raised.

liamhuber: Nice, I learned something new!

pyiron/atomistics/structure/visualization.py
samwaseda: I've been thinking about it all the time and I still don't know what's the best way. I'm right now slightly inclined to make the function private and not to put `atoms` in the DocString. At least in the current format, @sudarsan-surendralal is right that as long as the function is independently functional, there must be an entry for `atoms`, even though we know that it will be confusing in most of the cases since it will be mostly used from `Atoms`.

liamhuber: I think making the functions private is bad style, since we use it elsewhere, but I am also in agreement with Sudarsan's fundamental complaint. What happens if we make the whole visualization module private? Do we get style complaints when a sibling module inherits it with a relative path import call? I am happy to check, but I will have to wait until little mister is asleep to focus that deeply.

samwaseda: > What happens if we make the whole visualization module private?

What do you mean by that?

liamhuber: > > What happens if we make the whole visualization module private?
> 
> What do you mean by that?

@samwaseda yes, it works as I hoped!

In `pyiron/atomistics/structure` I added a module `_foo.py` containing a single method `bar`.

Then in `atoms.py` I can write `from . _foo import bar` and pycharm is perfectly happy. But over in, say, a random file in pyiron_contrib, if I write `from pyiron.atomistics.structure._foo import bar` the IDE gives me the warning that I am using "Access to a protected member _foo of a module".

Basically, by flagging the entire sub-module with an underscore we're making it private from outside the scope of its parent module. It's python, so consenting adults can access it if they want, but then I feel comfortable that we're no longer responsible for the fact that the docstrings don't make sense outside the context they were intended for.

@sudarsan-surendralal how do you feel about this?

samwaseda: I'm fine with the idea, but I don't really understand why it's bad to put `def _plot3d`.

liamhuber: If you put 'def _plot3d' in one module, then import it in another module, any decent IDE will complain that you're accessing a private variable. The leading underscore should indicate that the function (/class/module/method/variable) is only intended to be used within the current scope (i.e. the same module/module/parent module/class/module for function/class/module/method/variable, respectively). 

It will break absolutely nothing to use `def _plot3d`, it's just not conventional. Well, I guess it might break the codacy check, but whatever.

If we're going to use an underscore though, I'd rather underscore the module, because it's better form.
cf. the top answer in [this SO question](https://stackoverflow.com/questions/3602110/python-private-module-in-a-package), which backs up what my PyCharm 'experiment' showed me.

liamhuber: Actually, the answer from that SO question that directs to the [relevant passage of PEP8: public and internal interfaces](https://www.python.org/dev/peps/pep-0008/?#id50) is even better

samwaseda: Ah ok I didn't know that.

sudarsan-surendralal: I agree with @liamhuber, it's better to avoid using private functions in this case. This is because private functions are meant to be used only within the module (in this case: `visualization.py`) and therefore it's not good practice to import these private functions elsewhere (eg. `atoms.py`). I like the private module option that Liam mentioned (`_foo.py`). I haven't used this before but this is exactly what we want, public functions that are accessible to developers to be used elsewhere in the codebase but not to the users. 

Regarding the docstrings, I think it's nice to have `atoms` described as an argument even though it might seem redundant just for the sake of consistency (and to stop IDEs like pycharm from complaining!).

samwaseda: So for me the best options are:

- `visualize.py` and docstring entry for `atoms`
- `_visualize.py` and **no** docstring entry for `atoms`

Frankly I'm fine with both options but I'd be rather unwilling to do anything other than those.

liamhuber: Agreed. Since Sudarsan is on board with the private module, and it's a nifty part of python we haven't used yet, let's go with that ;)

liamhuber: I do think the module-level docstring in `_visualize.py` should have sentence to make clear the intention that those methods get bound to an Atoms object though.

samwaseda: Wait, if there's a DocString entry for `atoms`, then there's no reason to make `visualize.py` private, right?

samwaseda: Oh sorry so do I understand correctly that @sudarsan-surendralal wants a DocString entry for `atoms` and @liamhuber doesn't?

liamhuber: I propose your second solution: `_visualize.py`, where the public functions therein don't have a docstring element for `atoms`. However, I would additionally like to add to the head of `_visualize.py` a module-level docstring like 

"""
This module encapsulates visualization techniques for the `Atoms` class. The public facing methods here are intended to be bound to that class, and thus the first argument of these is always of types `Atoms` and should not appear in the docstrings.
"""

This fits between the imports at the `__author__` declaration for consistency with other modules.

samwaseda: Yes and that's where I see a difference between you two, since from what I understand @sudarsan-surendralal wants `atoms` explicitly in the DocString. I personally support your opinions equally, so you guys can decide what's better.

samwaseda: I just committed what could be a compromise -> I put `plot3d` etc. in the class `Visualize(object)`, which does nothing, but the first argument is now `self`. The rest is essentially the same (i.e. instead of `MethodType(plot3d, self)` I put `MethodType(Visualize.plot3d, self)` etc.) What do you think?

samwaseda: And since this class cannot be used standalone I indeed made a private module. Let me know what you think.

liamhuber: This comes back a little bit to my complaint that a class without data probably shouldn't be a class.

What if we just rename the `atoms` argument as `self`? I think this makes the intention maximally clear.

liamhuber: Maybe I'm barking up the wrong tree. Let's google around for a minute what common solutions are to too-long classes.

samwaseda: I'd claim it is a class with data, it's just this class itself does not provide data.

I guess pycharm will anyway complain whether it's called `atoms` or `self`. And I agree with @sudarsan-surendralal that as long as it's a function, all arguments should be listed in the DocString.

liamhuber: @samwaseda Sorry for causing so much trouble when you're just trying to clean things up!!! I think this sort of cleanup is needed in many places right now, so my motivation was just to give you a really hard time this once so we get a good transferable solution worked out ;)

liamhuber: > I'd claim it is a class with data, it's just this class itself does not provide data.
Agreed. I just thought with only a single piece of data it might be better as a module rather than a class. I definitely agree that we wind up running into docstring issues by going down that path though.


I couldn't find anything on the internet, but in Fowler's 'Refactoring', breaking off a new class to handle part of the class that originally had too much responsibility is indeed the suggested solution.

In the Fowler-recommended solution there's no rebinding of `object.sub_object.method` to `object.method`, but I guess for us `structure.plot3d` *must* be rebound for backwards compatibility.

So, frustratingly, at the end of the day I come back to most favouring your very original formulation. I still am not 100% happy with it, but it's still very clearly forward progress -- I think all the solutions we've discussed are, but the original one is probably the best/safest...

samwaseda: > @samwaseda Sorry for causing so much trouble when you're just trying to clean things up!!! I think this sort of cleanup is needed in many places right now, so my motivation was just to give you a really hard time this once so we get a good transferable solution worked out ;)

No worries - this one is an important conceptual question we should discuss to create a common ground. I find it particularly crucial in the absence of personal discussions, because now we are vulnerable to the danger that we keep doing things according to our own individual tastes, which will certainly make pyiron fall apart at some point. If we disagree on certain issues, we should definitely discuss, and I'd like to strongly encourage everyone to follow your example.

samwaseda: > So, frustratingly, at the end of the day I come back to most favouring your very original formulation. I still am not 100% happy with it, but it's still very clearly forward progress -- I think all the solutions we've discussed are, but the original one is probably the best/safest...

The original formulation being the current version, correct?

samwaseda: One thing I could do is to add

```
def __init__(self):
    raise ValueError('This class cannot be used standalone')
```

With this, `Visualize.plot3d(structure)` will still work, but it won't be possible to do things like `view = Visualize()`.

liamhuber: > No worries - ...

Cool, yes I entirely agree! I just wanted to be super clear that I am grateful for the effort you're putting in and am not simply trying to be a pain in the ass ;) I also think you're spot on on making sure that we have (at least a moderately) shared vision for how these things should look!

liamhuber: > The original formulation being the current version, correct?

No, rather

```python
# _visualization.py
class Visualize:
    def __init__(self, atoms):
        self._atoms = atoms
    
#atoms.py 
class Atoms(ASEAtoms):
    def __init__(self):
        self.visualize = Visualize(self)
```

liamhuber:
Could we also use this opportunity to discuss deep vs. wide?

For example, right now usage looks like `structure.plot3d_plotly(*args, **kwargs)` and `structure.analyze_ovito_centrosymmetry(*args, **kwargs)`.

I really like the tab-completion of pyiron, but I definitely find that many objects have a tab list so long I can't find what I'm looking for. How would you feel about `structure.visualization.plotly(*args, **kwargs)` and `structure.analysis.get_ovito_centrosymmetry(*args, **kwargs)` or even `structure.analysis.ovito.get_centrosymmetry(*args, **kwargs)`?

In the short term we'd still need to bind the aliases, but we could put a deprecation warning on them.

pyiron/atomistics/structure/visualization.py
samwaseda: Really? This will create 100 duplicate lines. In terms of reducing the size of `atoms.py`, that's quite inefficient, especially if it only boils down to the DocString problem. And from what I understand, Fowler is talking about class functions, which **must** retain a class structure. But in this case, as you correctly pointed out, the functionalities are largely independent, so the solution we have here does not go against what he describes (I'll gladly accept criticism though :D)

samwaseda:
> Could we also use this opportunity to discuss deep vs. wide?
> 
> For example, right now usage looks like `structure.plot3d_plotly(*args, **kwargs)` and `structure.analyze_ovito_centrosymmetry(*args, **kwargs)`.
> 
> I really like the tab-completion of pyiron, but I definitely find that many objects have a tab list so long I can't find what I'm looking for. How would you feel about `structure.visualization.plotly(*args, **kwargs)` and `structure.analysis.get_ovito_centrosymmetry(*args, **kwargs)` or even `structure.analysis.ovito.get_centrosymmetry(*args, **kwargs)`?
> 
> In the short term we'd still need to bind the aliases, but we could put a deprecation warning on them.

I agree. I also suggested to separate structure creation and structure analysis. Part of this effort was the recent split of `Neighbors`.

Regarding structure plot I'm not really sure if I agree with your idea, because I don't think it goes far beyond what we have right now. What requires more visualisation will probably be the `Analysis` class. And we usually want to see the plot quickly, which means I'd rather want to have to write only `structure.plot3d()`. I can imagine we could simply put all the plot functions together and do something like `structure.plot3d(mode='plotly')`, because the arguments should be mostly the same.

pmrv:
I take Liam's side on the deep vs. wide issue, but it's probably best left to a separate PR (for the analysis stuff).  Also like Sam's suggestion of `plot3d(mode='...')`.  Non-developer users likely won't care about the underlying plotting framework and it saves us from updating a bunch of notebooks, when we inevitably decide to use plotly by default instead of nglview.

pyiron/atomistics/structure/_visualize.py
```diff
@@ -0,0 +1,609 @@
...
+            opacity (float): opacity
+
+        Returns:
+            (plotly.express): The NGLView widget itself, which can be operated on further or viewed as-is.
+
+        """
+        try:
+            import plotly.express as px
+        except ModuleNotFoundError:
+            print("plotly not installed - use plot3d instead")
```
pmrv, line 64: I haven't followed the full discussion above, but `plot3d()` raises an error when `nglview` is not installed.  I'm not partial to either printing/warning or an exception, but it should be symmetric.

pyiron/atomistics/structure/_visualize.py
```diff
@@ -0,0 +1,609 @@
...
+        view.camera = camera
+        view.background = background
+
+        orientation = _get_flattened_orientation(view_plane=view_plane,
+                                                 distance_from_camera=distance_from_camera)
+        view.control.orient(orientation)
+
+        return view
+
+    def plot3d_ase(
```
pmrv, line 316: I don't quite understand what the difference between `plot3d()` and `plot3d_ase()` is, but since this was already this way previously, it can be a separate issue.  I'd want to discuss about this on Friday though.

pyiron/atomistics/structure/_visualize.py
```diff
@@ -0,0 +1,609 @@
...
+)
+__version__ = "1.0"
+__maintainer__ = "Sudarsan Surendralal"
+__email__ = "surendralal@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
+
+s = Settings()
+
+class Visualize(object):
```
pmrv, line 24: `class Visualize:` is enough in python3 and we dropped python2 support, so it's not needed to derive from `object`.

pyiron/atomistics/structure/atoms.py
```diff
@@ -192,6 +193,9 @@ def __init__(
             self.dimension = len(self.positions[0])
         else:
             self.dimension = 0
+        self.plot3d = MethodType(Visualize.plot3d, self)
```
pmrv, line 196: I think it's a bit weird to define the `Visualize` object and then never instantiate it and instead pick the methods like this.  I'd prefer simple functions in the `_visualize` module, but @samwaseda  and @liamhuber discussed this at length already, so I won't block this and follow what they agreed on.

pmrv (approved):
Only thing I'd like to see is the unification of the three plot functions via `plot3d(mode='...')` as suggested by @samwaseda. But doing this in a short follow-up PR is totally fine and more clear anyway, I'd think.

pyiron/atomistics/structure/visualization.py
liamhuber: Gah! No! I don't want 100 duplicate lines ;)
I must not have communicated clearly. I mean that I now have come full circle to embrace your original version where all the plotting methods are pulled out of `Atoms` and put into `visualization.Visualization`. However, unlike the current setup where the type of `Visualization` needs to implicitly be `Atoms`, I'm proposing that we go back to the old setup where it's just a regular class with a data field of type `Atoms`, then inside `Atoms` we instantiate it (`self.visuazation = Visualization(self)`, and wrap it's plot functions `def plot3d(self, *args, **kwargs): self.visualization.plot3d(*args, **kwargs)`.

I guess then we have a handful of dull lines where we wrap the plotting functions, but since we are only making quick aliases for the few public functions of `Visualize` it should be pretty short.

In fact, reading the further conversation it might be nice if `Visualize` had separate methods like `_plot3d_ase` and `_plot3d_plotly` (or whatever!), then a *single* public method `plot3d` which also takes the `method` kwarg. Then we'd only need to alias a *single* method in `Atoms`, and we could change the default plotting behaviour to use `plotly` instead of NGLView without borking the demo notebooks.

pyiron/atomistics/structure/atoms.py
liamhuber: Yes, I also find this problematic. I think the core of the issue here is that `Visualize` is just `object`-type (which Marvin rightly points out no longer needs to be explicit), but then its public methods assume that the `self` argument should be of type `Atoms`. The original formulation of this PR (which I fussed at...) really instantiated the `Visualization` class and passed it an `Atoms` instance as data.

This smells like a messed up inheritance tree to me. The closest thing I could see would be having something like `_AtomsBase(ASEAtoms)`, `_AtomsVisualization(AtomsBase)`, `_AtomsAnalysis(AtomsBase)`, etc., then finally making our public-facing class incorporate all the functionality using multiple inheritance: `Atoms(_AtomsVisualization, _AtomsAnalysis)`. 

However, if we chose to go this multiple inheritance route I think it's something that could be built *on top of* this PR, rather than stopping this PR to do it that way.

pyiron/atomistics/structure/_visualize.py
samwaseda: I can't really remember exactly why we had to create two functions, but I remember that for some time we were not so sure about our `plot3d`. Maybe that was the problem. To be honest, I don't think anyone still uses `plot3d_ase`, so maybe it's good to have it only in `plot3d(mode='ase')`.

pyiron/atomistics/structure/visualization.py
samwaseda: ah ok it's a good idea with `*args`. I'll do it this way.

I guess for `plotly` to be the default there are still quite some changes to make (`vector_field`, plotting frame etc.), so I'll leave this point for a later PR.

samwaseda: I guess now it's done.

samwaseda:
Is there a way to make the arguments available to tab completion without having to explicitly put all arguments in `plot3d` (inside `Visualize` and not `Atoms`)? The default values are extremely different and I also don't want to increase the number of lines unnecessarily.

liamhuber:
> Is there a way to make the arguments available to tab completion without having to explicitly put all arguments in `plot3d` (inside `Visualize` and not `Atoms`)? The default values are extremely different and I also don't want to increase the number of lines unnecessarily.

Not that I'm aware of. I guess you need to duplicate the signature when you're making the alias.

Another reason that, in the long run, we should use some fancier OOP solution like multiple inheritance ;)

pyiron/atomistics/structure/_visualize.py
```diff
@@ -0,0 +1,669 @@
...
+__status__ = "production"
+__date__ = "Sep 1, 2017"
+
+s = Settings()
+
+class Visualize:
+    def __init__(self, job):
+        self._job = job
+
+    def plot3d(self, mode='NGLview', **kwargs):
```
liamhuber, line 28: I guess this docstring is not valid until all the modes take the same arguments. I don't have strong feelings about how to handle this.

samwaseda:
Since I find it unacceptable that tab completion doesn't work, I put back this super ugly version. To be honest, I still find it nicer to do it with `MethodTypes` without initialising `Visualize`, but I'm okay with the current version as well.

liamhuber:
> Since I find it unacceptable that tab completion doesn't work, I put back this super ugly version. To be honest, I still find it nicer to do it with `MethodTypes` without initialising `Visualize`, but I'm okay with the current version as well.

Agreed, behind the scenes ugliness is worth it for forward-facing tab completion.

I agree we still have more work to do here, but I think the current state is a really nice step forward from before you started this 😃

samwaseda:
Ok then with this I'm going to merge this PR.

As a concluding remark, I want to thank everyone for your effort to make the best possible implementation and for the discussion which was both enlightening and inspiring. I'd like to let everyone know that I appreciate all contributions, however long the PR could potentially become.

And finally, I find it really funny that the whole discussion was ignited only by a single line in the DocString that was missing :)

liamhuber:
> Ok then with this I'm going to merge this PR.
> 
> As a concluding remark, I want to thank everyone for your effort to make the best possible implementation and for the discussion which was both enlightening and inspiring. I'd like to let everyone know that I appreciate all contributions, however long the PR could potentially become.

Huzzah!

Yes, I also appreciated the discussion. I found it very enlightening. I am still sorry that your small cleanup PR took so long to get merged, @samwaseda 😅 But I'm glad you anyhow found the process acceptable.

> And finally, I find it really funny that the whole discussion was ignited only by a single line in the DocString that was missing :)

Haha, indeed!!!

PR #1079, opened 2020-10-21
Elastic tensor class

samwaseda:
Following our recent discussion. DocString+Tests will follow.


samwaseda:
I guess now it's ready for review.

Btw. here's the representative line of DocString which characterises this class:

*This class is still under construction and there's no guarantee that it does what it looks like it does.*

pmrv:
Without having a detailed look, @liamhuber has a similar class [here](https://github.com/pyiron/pyiron_contrib/pull/28).  It looks like there could be opportunities to share code.  Generally I feel that if you expect this class to change a lot from now and this is not quite stable yet, it would fit better in pyiron_contrib.

samwaseda:
> Without having a detailed look, @liamhuber has a similar class [here](https://github.com/pyiron/pyiron_contrib/pull/28). It looks like there could be opportunities to share code. Generally I feel that if you expect this class to change a lot from now and this is not quite stable yet, it would fit better in pyiron_contrib.

So last week at the pyiron meeting we happened to have contributions related to elastic tensor and we found out that Liam's code relied on @yury-lysogorskiy 's code which had a license issue, and my code was doing what it was supposed to do. In fact I was not really planning to include this in pyiron, because I have the feeling that the user should always know how the elastic tensor is calculated, but the others apparently liked it and wanted me to implement it in pyiron. So in the end, it's true that it does what essentially @yury-lysogorskiy 's code did, but there's no political conflict in the new code and it's much shorter.

And it's true that I should have put it in pyiron_contrib - I've been just too lazy to clone the depository so far :D (and I also don't really understand where this kind of code should be deposited in pyiron_contrib). But this code does what it should do, it's just the input values (number of measurements etc.) are probably not well chosen. I guess this requires a thorough convergence analysis, but then we'll anyway find out that this is material specific as well.

pmrv:
Ah, didn't know about the licensing issues and previous discussion, that's fine by me then.

There's `pyiron_contrib.atomistics`, if we find in a few weeks that the input parameters, etc. are stable as they are, moving it here is very quick and doesn't incur a lot of update cost (just drop `import pyiron_contrib` from the notebook in question, or don't).

I'll do a proper review in a bit.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,223 @@
...
+__email__ = "janssen@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
+
+
+eV_div_A3_to_GPa = (
+    1e21 / scipy.constants.physical_constants["joule-electron volt relationship"][0]
+)
+
+def calc_elastic_tensor(strain, stress=None, energy=None, volume=None, rotations=None):
```
liamhuber, line 30: Nit: This is long and hard to read. It would be helpful to have it refactored into smaller chunks so that `calc_elastic_tensor` read more like

```python
if len(strain)==0:
         raise ValueError('Not enough points')

rotations = _build_all_rotations(rotations)
strain = _expand_strain_list_using_rotations(strain)
if stress is not None and len(stress)*len(rotations)==len(strain):
    coeff = _fit_coeffs_with_stresses(stress, strain, rotations)
elif energy is not None and volume is not None and len(energy)*len(rotations)==len(strain) and len(energy)==len(volume):  
    coeff = _fit_coeffs_with_energies(energy, volume, strain, rotations)
else:
         raise ValueError('Problem with fitting data')
return coeff
```

These functions might need to be further broken down. E.g. in the block I pulled out as `_fit_coeffs_with_energies`, the four-consecutive lines starting with `C=` or the four with `coeff=` are also yellow flags for pulling out a well-named function describing all the math.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+    Calculate elastic constants from the elastic tensor.
+
+    For anistropic material (i.e. zener_ratio!=1), the values may or may not make
+    sense -> don't trust the results straightforwardly
+
+    Args:
+        elastic_tensor (numpy.ndarray): 6x6 tensor
+    """
+    output = {}
```
liamhuber, line 88: Should this be an `InputList` instead of a dict? I'm not sure what the current big picture is, but we should start looking for these opportunities.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,223 @@
...
+    Class to calculate the elastic tensor and isotropic elastic constants
+
+    Example:
+
+    >>> job = pr.create_job('SomeAtomisticJob', 'atomistic')
+    >>> job.structure = pr.create_structure('Fe', 'bcc', 2.83)
+    >>> elastic = job.create_job('ElasticTensor', 'elastic')
+    >>> elastic.run()
+
+    This class is still under construction and there's no guarantee that it does
```
liamhuber, line 134: This is definitely a red flag that it should be over in `pyiron_contrib`....

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,223 @@
...
+    what it looks like it does.
+    """
+    def __init__(self, project, job_name="elastic"):
+        """
+
+        Args:
+            project:
```
liamhuber, line 141: Boooo missing docstrings! ;)

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,223 @@
...
+            job_name:
+        """
+        super().__init__(project, job_name)
+        self.__name__ = "ElasticTensor"
+        self.__version__ = "0.1.0"
+
+        # print ("h5_path: ", self.project_hdf5._h5_path)
+
+        # define default input
+        self.input["min_num_measurements"] = (11, "minimum number of samples to be taken")
```
liamhuber, line 151: `self.input` should be an `InputList`. This doesn't allow for giving the docstring in the param definition, but you can look over in the PR Marvin mentioned for my proposed solution to that. Briefly: at the class level in addition to `Attributes` I propose to add `Input` and `Output` lists giving the keys and values of these `InputList`s (-->`IOList`` one day)

liamhuber: If you do this you'll probably need to add `self.input.to_hdf(); self.output.to_hdf()` to the `to_hdf` method (and similar for `from_hdf`) as the parent job classes should but don't yet handle `InputList` being used this way.

tests/atomistics/master/test_elastic.py
```diff
@@ -0,0 +1,67 @@
...
+        cls.project.remove_jobs_silently(recursive=True)
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.file_location = os.path.dirname(os.path.abspath(__file__))
+        cls.project.remove(enable=True, enforce=True)
+
+    def test_validate_ready_to_run(self):
+        job = self.project.create_job(
+            'HessianJob', "job_test"
```
liamhuber, line 30: Nit: Inconsistent quotations

liamhuber (changes requested):
Very nice! I have a handful of small nits, and a request to refactor. I liked the structure though, and love that it's coming in with unit tests from day 1. Didn't see any functional problems.

As Marvin said, this relates very well to my SQSElastic code. That PR is still draft, so in the medium-term I'll switch it over to rely on your module instead of the private pyiron_mpie one currently in its stack.

pyiron/atomistics/master/elastic.py
samwaseda: > `self.input` should be an `InputList`.

DEFINITELY. The problem is right now it's derived from `AtomisticParallelMaster` (or `ParallelMaster`) and I have little control over the original input class...

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,223 @@
...
+        strain (numpy.ndarray): nx3x3 strain tensors
+        stress (numpy.ndarray): nx3x3 stress tensors
+        energy (numpy.ndarray): n energy values
+        volume (numpy.ndarray): n volume values
+        rotations (numpy.ndarray): mx3x3 rotation matrices
+    """
+    if len(strain)==0:
+        raise ValueError('Not enough points')
+    if rotations is not None:
+        rotations = np.append(np.eye(3), rotations).reshape(-1, 3, 3)
```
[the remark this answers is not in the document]
samwaseda: My concern is that the user gives an empty list, which in principle shouldn't happen as long as phonopy is used, but I guess it's a bit more robust. I'm easily overruled if my explanation is not sufficient though :)

pyiron/atomistics/master/elastic.py
samwaseda: ok that's true let me see

pyiron/atomistics/master/elastic.py
jan-janssen: > > `self.input` should be an `InputList`.
> 
> DEFINITELY. The problem is right now it's derived from `AtomisticParallelMaster` (or `ParallelMaster`) and I have little control over the original input class...

You can overwrite the input class.

pyiron/atomistics/master/elastic.py
samwaseda: uuuuuuuuh but do we need a docstring in `__init__`?

pyiron/atomistics/master/elastic.py
samwaseda: I changed the phrase (which definitely doesn't change anything but now it might sound better)

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,271 @@
...
+__author__ = "Joerg Neugebauer, Jan Janssen"
+__copyright__ = (
+    "Copyright 2020, Max-Planck-Institut für Eisenforschung GmbH - "
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Jan Janssen"
+__email__ = "janssen@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
```
pmrv, line 23: We'll probably want to update that to at least this year. ;)

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,271 @@
...
+    orientation = np.einsum('ij,i->ij', orientation, 1/np.linalg.norm(orientation, axis=-1))
+    if not np.isclose(np.linalg.det(orientation), 1):
+        raise ValueError('orientation must be an orthogonal 3x3 tensor')
+    C = np.zeros((3,3,3,3))
+    for i in range(3):
+        for j in range(3):
+            for k in range(3):
+                for l in range(3):
+                    C[i,j,k,l] = elastic_tensor[i+(i!=j)*(6-2*i-j), k+(k!=l)*(6-2*k-l)]
+    C = np.einsum('Ii,Jj,ijkl,Kk,Ll->IJKL', orientation, orientation, C, orientation, orientation)
```
pmrv, line 50: I'm fine with einsum, but a comment giving the equivalent in Matrix notation would be nice. I assume this is a rotation?

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,271 @@
...
+    else:
+        rotations = np.eye(3)
+    strain = np.einsum('nik,mkl,njl->nmij', rotations, strain, rotations).reshape(-1, 3, 3)
+    strain = np.stack((strain[:,0,0], strain[:,1,1], strain[:,2,2], 2*strain[:,1,2], 2*strain[:,0,2], 2*strain[:,0,1]), axis=-1)
+    if stress is not None and len(stress)*len(rotations)==len(strain):
+        coeff, score = _fit_coeffs_with_stress(strain=strain, stress=stress, rotations=rotations)
+    elif energy is not None and volume is not None and len(energy)*len(rotations)==len(strain) and len(energy)==len(volume):
+        coeff, score = _fit_coeffs_with_energies(strain=strain, energy=energy, volume=volume, rotations=rotations)
+    else:
+        raise ValueError('Problem with fitting data')
```
pmrv, line 115: Message should be more specific, like "should have either stress or energy and volume".

pyiron/atomistics/master/elastic.py
pmrv: Also, if this is only a dict does it end up in the HDF (not that an InputList would help in this regard yet)?

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def job_name(self, parameter):
+        return "{}_{}".format(self._job.job_name, parameter[0]).replace(".", "_")
+
+    @staticmethod
+    def modify_job(job, parameter):
+        job.structure = parameter[1]
+        return job
+
+# ToDo: not all abstract methods implemented
```
pmrv, line 164: Which are missing?

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,271 @@
...
+    >>> elastic = job.create_job('ElasticTensor', 'elastic')
+    >>> elastic.run()
+
+    The input parameters might not be chosen adequately. If you have a large
+    computation power, increase `min_num_measurements`. At the same time, make
+    sure to choose an orientation which maximizes the number symmetry operations.
+    Also if the child job does not support pressure, better increase the number
+    of measurements.
+    """
+    def __init__(self, project, job_name="elastic"):
```
pmrv, line 182: Major nit but still, do we usually give default job names?

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,271 @@
...
+
+        Args:
+            project:
+            job_name:
+        """
+        super().__init__(project, job_name)
+        self.__name__ = "ElasticTensor"
+        self.__version__ = "0.1.0"
+
+        self.input["min_num_measurements"] = (11, "minimum number of samples to be taken")
```
pmrv, line 193: Having the input parameters comments also/only in the class docstring would be nice, when you browse readthedocs.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,271 @@
...
+            _, indices = np.unique(np.round(rotations, 6), axis=0, return_inverse=True)
+            rotations = rotations[np.unique(indices)]
+            self.input['rotations'] = rotations.tolist()
+        if len(self.input['strain_matrices'])==0:
+            eps_lst = np.random.random((int(self._number_of_measurements), 3, 3))-0.5
+            eps_lst *= self.input['max_strain']
+            eps_lst += np.einsum('nij->nji', eps_lst)
+            self.input['strain_matrices'] = eps_lst.tolist()
+
+    def list_structures(self):
```
pmrv, line 227: Since this is potentially called by users a very brief docstring would be nice.  Just so it's more visible on readthedocs.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,271 @@
...
+            return_score = True,
+        )
+        self._output['fit_score'] = score
+        for k,v in calc_elastic_constants(elastic_tensor).items():
+            self._output[k] = v
+        with self.project_hdf5.open("output") as hdf5_out:
+            for key, val in self._output.items():
+                hdf5_out[key] = val
+
+    def get_C_by_orientation(self, orientation):
```
pmrv, line 268: Same as `list_structures`.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,271 @@
...
+    Class to calculate the elastic tensor and isotropic elastic constants
+
+    Example:
+
+    >>> job = pr.create_job('SomeAtomisticJob', 'atomistic')
+    >>> job.structure = pr.create_structure('Fe', 'bcc', 2.83)
+    >>> elastic = job.create_job('ElasticTensor', 'elastic')
+    >>> elastic.run()
+
+    The input parameters might not be chosen adequately. If you have a large
```
pmrv, line 176: If the naming/setup of input parameters is your only concern, I'm happy to merge directly into pyiron even without this sentence.  If turns out we want to tweak them we can always do that before bumping the minor version.

pmrv (approved):
I have mostly some nits, but generally happy with this.

pyiron/atomistics/master/elastic.py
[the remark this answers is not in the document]
liamhuber: Right, fair enough. Look at this again though, I guess these two functions should be made private since they're not intended to be used outside elastic.py. In this case I'm less worried about a user passing empty rotations because it would be a developer who would have to do it rather than a user, but even so your point is very reasonable.

pyiron/atomistics/master/elastic.py
liamhuber: I guess not, but then let's clean up the stub. Same for the job generator. If there's no change relative to the parent I don't see a need for a docstring, but what's there now is just clutter.

pyiron/atomistics/master/elastic.py
liamhuber: The only thing I get afraid of there is someone using this in a notebook and then demanding backwards compatibility. Or someone demanding backwards compatibility whether anyone's using it or not. 

...In case your detecting a theme from my github activity this week, backwards compatibility is something that keeps me awake at night 😆

pyiron/atomistics/master/elastic.py
liamhuber: Agreed, looking at this fresh that also surprised me.

pyiron/atomistics/master/elastic.py
liamhuber: Actually, I'd hold off here. IMO this should not be a public method. We need to go back into the base classes and tweak some stuff. I'm pretty sure this only gets called internally by the job itself when it's run with status=collect.

liamhuber (commented):
I didn't mean to start a fresh review, but now I'm worried the 'cancel review' button will kill my three comments 😛

pyiron/atomistics/master/elastic.py
pmrv: Not sure if this is a good idea, as the `to_hdf`/`from_hdf` of the base class might expect `input` to be a `GenericParameter` object.  I haven't checked the source though.

pyiron/atomistics/master/elastic.py
samwaseda: It's a 4d tensor. As long as python is written on the 2d screen, I don't think there's a practical way to write it clearly without using indices.

pyiron/atomistics/master/elastic.py
samwaseda: Yeah that's true. It already disturbed me during debugging :D

pyiron/atomistics/master/elastic.py
pmrv: I feel it's reasonable to assume nothing enters public API between version bumps.   Then this class would only become visible after the next version bump, this is what the last part of my previous comment refers to.

pyiron/atomistics/master/elastic.py
samwaseda: Nothing, only my attention was missing 😎

pyiron/atomistics/master/elastic.py
samwaseda: No we shouldn't. I guess it happened when I copied it from Murnaghan and didn't think about it.

pyiron/atomistics/master/elastic.py
samwaseda: Actually this one comes from `AtomisticParallelMaster` and I guess the function name must be like this. Maybe @jan-janssen can say more about it.

pyiron/atomistics/master/elastic.py
samwaseda: True, I just fixed it

pyiron/atomistics/master/elastic.py
samwaseda: oh I didn't realise that it was not 2017 anymore :D

pyiron/atomistics/master/elastic.py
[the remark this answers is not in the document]
samwaseda: Ah! That's actually exactly the opposite of what I'd like the user to do. The thing is, these fit functions are perfect and I'd probably also use them just as they are. The problem is the data collection - there I'd like the user to set up the simulations themselves and measure strain tensors etc. as well, which is not difficult at all. The mathematically difficult part is the fitting, and for this you can use these functions.

pyiron/atomistics/master/elastic.py
samwaseda: No it doesn't, it's just a function which returns output values. I guess it has a solid utility since you might be having an elastic tensor from somewhere and you might want to have elastic constants. For this, it makes more sense to have an independent function.

pyiron/atomistics/master/elastic.py
samwaseda: In fact this can only be used by the user - it's not used internally. Here's how you'd use it: If you have a box in which a dislocation is meant to be inserted, you might have an orientation like [[1, 1, 1], [-1, 0, 1], [1, -2, 1]]. In which case, you can either create a box from the beginning with this orientation, or you rotate the elastic tensor. Since creating a box with a weird orientation both increases the number of atoms (and hence the computation time) and decreases the number of symmetries phonopy finds, usually it makes more sense to rotate the elastic tensor. And those who are familiar with the anisotropic elasticity theory (I can't remember who did the work) might know that a full elastic tensor is required to parametrise the strain field around a dislocation for the given orientation. So this function totally makes sense for such a case.

samwaseda: But it's true that the name is not quite comprehensible so I change it now.

pyiron/atomistics/master/elastic.py
pmrv: Mh, I guess on second glance it's actually clear what this one does.  My bad, then.

pyiron/atomistics/master/elastic.py
samwaseda: Sorry I was wrong - it does nothing. I just deleted it

samwaseda: And I guess it's a useful function, so I put it back with a DocString :D

samwaseda: ... and I just realized that it cannot work in the current framework so in the end I removed it again :D

samwaseda:
I just started wondering if there's a method to better distribute the measurement points, i.e. in principle we have 6 independent strain components epsilon_0 to epsilon_5. And I could start saying I'd like to distribute these points as homogeneously as possible within the maximum strain. Do you guys know if there's a good clear method which can do something like this? This must be somehow related to ML techniques as well.

samwaseda:
> I just started wondering if there's a method to better distribute the measurement points, i.e. in principle we have 6 independent strain components epsilon_0 to epsilon_5. And I could start saying I'd like to distribute these points as homogeneously as possible within the maximum strain. Do you guys know if there's a good clear method which can do something like this? This must be somehow related to ML techniques as well.

ok here's what I got: I define the quality of epsilons as the sum of the distances to the nearest neighbouring epsilon for all epsilons, i.e.

Q = sum_i min_j || epsilon_i - epsilon_j ||

1. Propose a set of epsilons
2. Get the current quality
3. Change one epsilon
4. Calculate the new quality
5. If the new quality if lower than the old one, revert the change
6. Back to step 2

samwaseda:
> 
> ok here's what I got: I define the quality of epsilons as the sum of the distances to the nearest neighbouring epsilon for all epsilons, i.e.
> 
> Q = sum_i min_j || epsilon_i - epsilon_j ||
> 
> 1. Propose a set of epsilons
> 2. Get the current quality
> 3. Change one epsilon
> 4. Calculate the new quality
> 5. If the new quality if lower than the old one, revert the change
> 6. Back to step 2

ok this brings nothing. Forget about it.

pmrv:
> I just started wondering if there's a method to better distribute the measurement points, i.e. in principle we have 6 independent strain components epsilon_0 to epsilon_5. And I could start saying I'd like to distribute these points as homogeneously as possible within the maximum strain. Do you guys know if there's a good clear method which can do something like this? This must be somehow related to ML techniques as well.

I found this once, but it's probably overkill https://github.com/python-adaptive/adaptive

samwaseda:
I just got a fascinating brilliant idea after today's meeting. It'll take some time to implement my idea so I change this one to draft for now.

samwaseda:
So the first version of what @JNmpi wanted and I can offer is now here. Here's the idea: for each strain tensor, the algorithm defines an internal axis, along which stress+energy are measured. Afterwards, coefficients are fitted along these axes where only the lowest order (which corresponds to the elastic tensor) is decomposed into the Cartesian coordinates. This strategy allows us to keep the number of coefficients very low, while anharmonic contributions will be properly taken into account. I will explain the physics behind this in detail at the next pyiron meeting, so you don't need to try to understand how it works for now.

samwaseda:
Now the class is entirely done. And since the number of lines was increasing endlessly, I decided to open a [new issue](https://github.com/pyiron/pyiron/pull/1129) for extra functionalities (e.g. rotations, elastic constants).

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+    strain_lst = np.random.random((n_set, 3, 3))-0.5
+    strain_lst += np.einsum('nij->nji', strain_lst)
+    if normalize:
+        strain_lst = np.einsum('nij,n->nij', strain_lst, 1/np.linalg.norm(strain_lst.reshape(-1, 9), axis=-1))
+    strain_lst *= max_strain
+    m = np.linspace(-1, 1, int(2*polynomial_order+2*additional_points-1))
+    strain_lst = np.einsum('k,nij->nkij', m, strain_lst).reshape(-1, 9)
+    return np.unique(strain_lst, axis=0).reshape(-1, 3, 3)
+
+class ElasticJobGenerator(JobGenerator):
```
liamhuber, line 237: Nit: IMO, these generator classes (except the ABC `JobGenerator`) should be private.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+        return "{}_{}".format(self._job.job_name, parameter[0]).replace(".", "_")
+
+    @staticmethod
+    def modify_job(job, parameter):
+        job.structure = parameter[1]
+        return job
+
+class ElasticTensor(AtomisticParallelMaster):
+    """
+    Class to calculate the elastic tensor and isotropic elastic constants
```
liamhuber, line 257: Nit: periods at the end of sentences through docstrings. (Only if this makes it more consistent with the rest of the code base. I'm pretty sure it does, but I only spot checked a couple methods...)

pyiron/atomistics/master/elastic.py
liamhuber: This is part of a broader push I'd like to make to start privatizing stuff that never gets used outside its scope even by the developers (and isn't expected to ever be used outside that scope).

pyiron/atomistics/master/elastic.py
liamhuber: No, it should be fine. You'll need to explicitly handle the hdf operations for input and output here (unless @niklassiemer shifts his changes in [pyiron_base PR 50] up the inheritance tree like I hope he will 😉). Otherwise it's worked fine for me.

I still want to see this!

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+        self.input["min_num_measurements"] = (11, "minimum number of samples to be taken")
+        self.input["min_num_points"] = (105, "minimum number of data points"
+            + "(number of measurements will be min_num_points/len(rotations))")
+        self.input["max_strain"] = (
+            0.01,
+            "relative volume variation around volume defined by ref_ham",
+        )
+        self.input['polynomial_order'] = 2
+        self.input['additional_points'] = 0
+        self.input['strain_matrices'] = []
```
liamhuber, line 297: Nit: Please add attributes in the same order as they appear in the docstring. This makes it easier to see when a docstring element is missing.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+            times number of symmetry operations)
+        strain_matrices (numpy.ndarray): Strain tensors to be applied on simulation boxes (optional)
+        rotations (numpy.ndarray): Rotation matrices for box symmetry
+        use_symmetry (bool): Whether or not exploit box symmetry (ignored if `rotations` already specified)
+        use_elements (bool): Whether or not respect chemical elements for box symmetry (ignored if `rotations`
+            already specified)
+
+    The default input parameters might not be chosen adequately. If you have a large
+    computation power, increase `min_num_measurements`. At the same time, make
+    sure to choose an orientation which maximizes the number symmetry operations.
```
liamhuber, line 279: I really like this added context! I would go even further and quickly explain *why* not supporting pressure is a problem, i.e. something like "--without pressure the constants must be fit to total cell energies, which contain only a fraction of the data that the pressure matrix does. This lack can be compensated by sampling more rotations." (or  however you want to word it.)

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+        eps_lst *= self.input['max_strain']
+        eps_lst += np.einsum('nij->nji', eps_lst)
+        self.input['strain_matrices'] = get_strain(max_strain=self.input['max_strain'],
+                                                   n_set=self._number_of_measurements,
+                                                   polynomial_order=self.input['polynomial_order'],
+                                                   additional_points=self.input['additional_points'],
+                                                   normalize=self.input['normalize_magnitude']).tolist()
+
+    def validate_ready_to_run(self):
+        super().validate_ready_to_run()
```
liamhuber, line 332: I'm a little torn whether we should be doing these in setters rather than all the way at `validate_ready_to_run` -- I don't think you need to change anything here! It's just a general question I have.

I really like all the checks and find the error messages clear :)

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+                    else:
+                        output_dict[key] = []
+                output_dict['id'].append(job_id)
+            for k,v in output_dict.items():
+                self._output[k] = np.array(v)
+        energy = self._output['energy_tot']
+        if len(self._output['energy_pot'])==len(self._output['volume']):
+            energy = self._output['energy_pot']
+        elastic_tensor, score = calc_elastic_tensor(
+            strain = self.input['strain_matrices'],
```
liamhuber, line 370: PEP8: No spaces around `=` in function arguments.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+            warnings.warn('Setting additional points for anharmonic calculations only increases the number of calculations')
+        if self.input['polynomial_order']==2 and self.input['normalize_magnitude']:
+            warnings.warn('Magnitude normalization could reduce accuracy in harmonic calculations')
+        if self.input['polynomial_order']>2 and not self.input['normalize_magnitude']:
+            warnings.warn('Not normalizing magnitude could destabilise fit procedure')
+        if self.input['fit_first_order']:
+            warnings.warn('first order fit does not correct the second order coefficients in the current implementation')
+
+    def collect_output(self):
+        if self.ref_job.server.run_mode.interactive:
```
liamhuber, line 347: Nit: This `if/else` looks like it needs a refactor. I'm just writing in the comment and not testing it so take this with a grain of salt, but I think the following should be equivalent:

```python
for key in ['energy_tot', 'energy_pot', 'pressures', 'volume']:
    if self.ref_job.server.run_mode.interactive:
        self.output[key] = self._read_child_output(self.project_hdf5.inspect(self.child_ids[0]))
    else:
        self.output[key] = [self._read_child_output(self.project_hdf5.inspect(job_id)) for job_id in self.child_ids]
```

where

```python
def _read_child_output(hdf, key):
    if key in hdf["output/generic"].list_nodes():
        return hdf["output/generic/{}".format(key)][-1])
    else:
        return []
```

This cuts 19 lines down to 10 and (IMO) improves readability.

liamhuber: Note: I also sneakily use `self.output` instead of `self._output` so I have an excuse to point it out and remind you that `__init__` should have `self._output = InputList(table_name='output')` and there should be an `output` property and setter wrapping assignments as `InputList` ;) 

(You might need to fiddle with the table name...for `input` there's definitely no conflict, for output I'm not as sure.)

tests/atomistics/master/test_elastic.py
```diff
@@ -0,0 +1,68 @@
...
+
+import os
+import unittest
+from pyiron.atomistics.structure.atoms import CrystalStructure
+from pyiron_base import Project
+from pyiron.atomistics.master.elastic import calc_elastic_tensor
+import numpy as np
+
+
+class TestMurnaghan(unittest.TestCase):
```
liamhuber, line 13: Naming error. Shouldn't effect functionality, but definitely confusing for the reader 😉

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+        self.input["max_strain"] = (
+            0.01,
+            "relative volume variation around volume defined by ref_ham",
+        )
+        self.input['polynomial_order'] = 2
+        self.input['additional_points'] = 0
+        self.input['strain_matrices'] = []
+        self.input['use_symmetry'] = True
+        self.input['rotations'] = []
+        self.input['normalize_magnitude'] = False
```
liamhuber, line 300: Missing docstring entry

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+        self.input['use_elements'] = (True, 'whether or not consider chemical elements for '
+                                            + 'the symmetry analysis. Could be useful for SQS')
+        self.input['fit_first_order'] = False
```
liamhuber, line 303: Missing docstring entry. Might need a new name too as without the docstring it's not immediately obvious what it does. (Take with a grain of salt though, I don't do this type of calculation and it might be perfectly sensible to the practitioner.)

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+    else:
+        raise ValueError('Provide either strain and stress, or strain, energy '
+                         + 'and volume of same length.')
+    if return_score:
+        return np.array(coeff)[:,:6], score
+    else:
+        return np.array(coeff)[:,:6]
+
+def get_strain(max_strain=0.05, n_set=10, polynomial_order=3, additional_points=0, normalize=False):
+    strain_lst = np.random.random((n_set, 3, 3))-0.5
```
liamhuber, line 228: This code also appears down in `ElasticTensor._create_strain_matrices`. I think it would be worthwhile factoring out a method

```python
_random_symmetric_matrices(n):
    matrices = np.random.random((n, 3, 3))-0.5
    return matrices + np.einsum('nij->nji', matrices)
```

Here and below you rescale, but here you also (potentially) normalize. I think it's fine to just return random on (-0.5, 0.5) and rescale similar to you have it. Although later you can do something like `eps = self.scale * _random_symmetric_matrices(self.n_whatevers)`. (But with all the right names, obviously.)

liamhuber: p.s. I think I might be learning einsum by osmosis 😝

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+                       1/(norm+np.isclose(norm, 0))), decimals=8)
+    indices = np.unique(indices, axis=0, return_inverse=True)[1]
+    na = np.newaxis
+    indices = np.unique(indices[~np.isclose(norm, 0)])[:,na]==indices[na,:]
+    # 0-tensor gets a special treatment, since it's linearly dependent on all
+    # terms (even though it virtually changes nothing)
+    indices[:,np.isclose(norm, 0)] = True
+    return indices
+
+def _get_higher_order_terms(
```
liamhuber, line 119: As a non-practitioner, the nomenclature is `higher_order_terms` is not clear, but the docstring opens with a circular reference...

liamhuber, line 119: This function is difficult to follow. The fact it's private but has a docstring, and also includes inline `# comments` leads me to believe you're already aware of this 😛 Can clarity be added by breaking it down into smaller, well-named functions? I'd be willing to pay the cost of a few extra lines in exchange for the readability if you see nice places to break it apart.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+    s = np.array(strain_lst).reshape(-1, 9)
+    norm = np.linalg.norm(s, axis=-1)
+    indices = np.round(np.einsum('ni,n,n->ni', s, np.sign(np.sum(s, axis=-1)),
+                       1/(norm+np.isclose(norm, 0))), decimals=8)
+    indices = np.unique(indices, axis=0, return_inverse=True)[1]
+    na = np.newaxis
+    indices = np.unique(indices[~np.isclose(norm, 0)])[:,na]==indices[na,:]
+    # 0-tensor gets a special treatment, since it's linearly dependent on all
+    # terms (even though it virtually changes nothing)
+    indices[:,np.isclose(norm, 0)] = True
```
liamhuber, line 116: PEP8: space --- wait, actually it might not be strictly PEP8 to put a space after a comma in the `[]`. But I think it would be more consistent with your style elsewhere.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+        max_polynomial_order=None,
+        fit_first_order=False,
+    ):
+    higher_terms = _get_higher_order_terms(
+        strain,
+        max_polynomial_order=max_polynomial_order,
+        rotations=rotations,
+        derivative=True,
+    )
+    strain = np.einsum('nik,mkl,njl->nmij',
```
liamhuber, line 45: Twice here you take an einsum and then re-order the output with a magic map `[0, 4, 8, 5, 2, 1]`. I think it would benefit from being factored out as a function with a good name, because right now it seems very magical.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+    stress = np.einsum('nik,mkl,njl->nmij',
+                       rotations, stress, rotations).reshape(-1, 9)
+    stress = stress[:, [0, 4, 8, 5, 2, 1]]
+    score = []
+    coeff = []
+    for ii in range(6):
+        strain_tmp = strain.copy()
+        strain_tmp[:, 6:] *= strain[:, ii, np.newaxis]
+        reg = LinearRegression(
+            fit_intercept=fit_first_order).fit(strain_tmp, stress[:,ii])
```
liamhuber, line 60: Nit: line break is hard to read. Do you really go over the limit here? If you need to break it I'd rather break it a second time so the argument is on its own, but I'm open to alternative style interpretations.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+    ):
+    higher_terms = _get_higher_order_terms(
+        strain,
+        max_polynomial_order=max_polynomial_order,
+        rotations=rotations,
+        derivative=False,
+    )
+    strain_voigt = np.einsum('nik,mkl,njl->nmij',
+                             rotations, strain, rotations).reshape(-1, 9)
+    strain_voigt = strain_voigt[:, [0, 4, 8, 5, 2, 1]]
```
liamhuber, line 81: third use of A^T.B.A+weird reordering (or whatever the linalg is). I am now implacable without extracting it to a method.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+    strain_voigt[:,3:] *= 2
+    energy = np.tile(energy, len(rotations))
+    volume = np.tile(volume, len(rotations))
+    strain = 0.5*np.einsum('ni,nj->nij', strain_voigt, strain_voigt)
+    strain = np.triu(strain).reshape(-1, 36)
+    strain = strain[:,np.sum(strain, axis=0)!=0]
```
liamhuber, line 87: Three consecutive strain lines also smells like it needs a well-named function to summarize the operation that is being made

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+        strain = np.concatenate((strain, higher_terms), axis=-1)
+    if fit_first_order:
+        strain = np.concatenate((strain, strain_voigt), axis=-1)
+    strain = np.einsum('n,ni->ni', volume, strain)
+    reg = LinearRegression().fit(strain, energy)
+    score = reg.score(strain, energy)
+    coeff = np.triu(np.ones((6,6))).flatten()
+    coeff[coeff!=0] *= reg.coef_[:21]*eV_div_A3_to_GPa
+    coeff = coeff.reshape(6,6)
+    coeff = 0.5*(coeff+coeff.T)
```
liamhuber, line 98: ditto four consecutive `coeff` lines

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+import numpy as np
+import scipy.constants
+from pyiron.atomistics.master.parallel import AtomisticParallelMaster
+from pyiron_base import JobGenerator
+from sklearn.linear_model import LinearRegression
+from collections import defaultdict
+import warnings
+import itertools
+
+__author__ = "Joerg Neugebauer, Jan Janssen"
```
liamhuber, line 16: Shouldn't this just be `(O)Sam(u) Waseda`? I don't have any objection to putting them here after you, to be clear, but looking at the commit history I think the first (if not only) author is unambiguously you, @samwaseda

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,385 @@
...
+__copyright__ = (
+    "Copyright 2020, Max-Planck-Institut für Eisenforschung GmbH - "
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Jan Janssen"
```
liamhuber, line 22: Hmm, I might need to steal this idea myself... write some code and then slap Jan down as the maintainer 😝

liamhuber (changes requested):
Cool stuff, @samwaseda! I covered it in red ink with nits and requests to refactor for readability, but don't let that distract from my excitement to get this in the code base. It's a nice piece of work!

The pyiron classes themselves were super close to top-notch, but I had a lot of trouble reading the functions at the top. In anticipation of someone other than you maintaining this one day, I think it's prudent to put some TLC into getting these more readable now while it's all fresh in your mind.

samwaseda:
By the way, I put a notebook in the pyiron meeting folder for next Friday. This was the one that I would have used to explain the new module today, and I have to admit it's not quite a self-explanatory notebook, but maybe you can take a look into it. Anyway I'll talk about it next week.

pyiron/atomistics/master/elastic.py
samwaseda: Good point

pyiron/atomistics/master/elastic.py
samwaseda: Actually I copied it from `Murnaghan` without thinking much about it, but I guess the reason why it's done this way is because it performs one `inspect` for each job, and probably that makes the parsing faster.

samwaseda: > Note: I also sneakily use `self.output` instead of `self._output` so I have an excuse to point it out and remind you that `__init__` should have `self._output = InputList(table_name='output')` and there should be an `output` property and setter wrapping assignments as `InputList` ;)
> 
> (You might need to fiddle with the table name...for `input` there's definitely no conflict, for output I'm not as sure.)

And I still don't know how to use `InputList` properly XD. Do you think if I simply add `self._output = InputList(table_name='output')` it's gonna start working straightaway?

pyiron/atomistics/master/elastic.py
samwaseda: yessss I'm horribly aware of the problem. I'm gonna try to make it somewhat easier to understand. When I first created the function I myself didn't really know what I needed, but now after I got better understanding, I can maybe try to make it clearer...

pyiron/atomistics/master/elastic.py
samwaseda: I guess the order was the same but there were simply quite a few attributes/keys missing.

pyiron/atomistics/master/elastic.py
samwaseda: Yeah that's true. It's just this `input` is inherited from somewhere deep down and I didn't really want to look into it too much. But it's true that for example in the case of `polynomial_order` it makes more sense to do it in the setter. When we switch to `InputList` I'll do it at the same time.

pyiron/atomistics/master/elastic.py
samwaseda: Done!

pyiron/atomistics/master/elastic.py
samwaseda: Hui this one's gonna be difficult to change. Can you elaborate on your comment?

samwaseda: oh sorry now I just saw what ditto refers to

pyiron/atomistics/master/elastic.py
samwaseda: Here I remove the lower triangle, which is the same as the upper triangle to stabilise the linear regression (what I do below with `coeff` is the opposite transformation). I add a few comments for now because I don't use the same functional elsewhere and I can't come up with a fancy function name makes it super clear.

pyiron/atomistics/master/elastic.py
samwaseda: I changed it to `higher_order_strains` but I guess the fundamental problem is the one that you state below...

pyiron/atomistics/master/elastic.py
liamhuber: > Actually I copied it from `Murnaghan` without thinking much about it, but I guess the reason why it's done this way is because it performs one `inspect` for each job, and probably that makes the parsing faster.

Ah, good point. I still find the code messy though. What about something like this:

```python
if self.ref_job.server.run_mode.interactive:
    ids = [self.child_ids[0]]
else:
    ids = [job_id for job_id in self.child_ids]
self._output = self._add_energy_pressure_volume_output(self._output, ids)
```

```python
def _add_energy_pressure_volume_output(self, output, ids):
    for id_ in ids:
        hdf = self.project_hdf5.inspect(id_)
        for key in ['energy_tot', 'energy_pot', 'pressures', 'volume']:
            val = _read_output_key_or_empty(hdf, key)
            try:
                output[key].append(val)
            except KeyError:
                output[key] = [val]
    # I'm not sure about the formatting here -- 
    #    I'm wrapping the interactive case as a list, and never converting the regular case to np.array
    #    But there must be *some* way to do this nicely without as much duplicate code as we have now

@staticmethod
def _read_output_key_or_empty(hdf, key):
    # I suspect this is useful all over the codebase atm
    if key in hdf["output/generic"].list_nodes():
        return hdf["output/generic/{}".format(key)][-1])
    else:
        return []
```

liamhuber: > And I still don't know how to use `InputList` properly XD. Do you think if I simply add `self._output = InputList(table_name='output')` it's gonna start working straightaway?

Yes, should do. You'll need to add `self._output.to_hdf(self._hdf5)` and `self._output.from_hdf(self._hdf5)` (or extremely similar?) in the `to_` and `from_hdf` methods. Other than that you should be good to go.

pyiron/atomistics/master/elastic.py
liamhuber: That's totally fair. I just wanted to do some prodding to encourage you to follow through on your good intentions 😝

pyiron/atomistics/master/elastic.py
liamhuber: I am in favour of extracting it out as a method for clean reading, but if you're more comfortable just adding a comment (as in the discussion above) I don't mind that either.

pyiron/atomistics/master/elastic.py
```diff
@@ -0,0 +1,493 @@
...
+    must be fit to total cell energies, which contain only a fraction of the
+    data that the pressure matrix does. This lack can be compensated by
+    sampling more rotations.
+    """
+    def __init__(self, project, job_name):
+        super().__init__(project, job_name)
+        self.__name__ = "ElasticTensor"
+        self.__version__ = "0.1.0"
+
+        self.input["min_num_measurements"] = (
```
liamhuber, line 363: First, still old input class? I feel like you had a reason for this, but given the [rapid progress of InputList](https://github.com/pyiron/pyiron_base/pull/70#issuecomment-725628115) I think we should switch over...

Second, the long set of default values here plus all the validity checking later leads me to think this should be a subclass of `InputList` that uses properties to make sure the values are all correct:

```python
_ElasticTensorInput(InputList):
    __init__(self, init=None, table_name=None):
        super().__init__(init=init, table_name=table_name)
        #blah blah
        self._polynomial_order = 2
        #blah blah

    @property 
    def polynomial_order(self):
        return self._polynomial_order

    @polynomial_order.setter
    def polynomial_order(self, order):
        if order < 2:
            raise ValueError("polynomial_order should be an integer >=2 but got {}".format(order))
        self._polynomial_order = order
```

And then in `ElasticTensor.__init__`: `self.input = _ElaticTensorInput(table_name='input')`

liamhuber (changes requested):
I would still like to see a transition over to `InputList` and offer some justification for that.

It could also be updated to use 120 char lines as we discussed elsewhere.

Otherwise looking awesome!

pyiron/atomistics/master/elastic.py
samwaseda: The most important reason is because it's derived from `AtomisticParallelMaster`, which still uses `GenericParameters` and most of the internal routines are already defined there.

liamhuber: > The most important reason is because it's derived from `AtomisticParallelMaster`, which still uses `GenericParameters` and most of the internal routines are already defined there.

Fine.

Fine

😝 but seriously, for the purpose of this PR I grudgingly accept.

From a management perspective, I think `InputList` adoption is something we should probably prioritize more. The more derived classes we make that don't use it, the more of a pain in the butt the transition becomes. In the meantime, on the usage side we wind up with more and more notebooks with mixed formatting like this:

```python
joba.input.foo = 'foo'
joba.input.bar  = 42
jobb.input['typedkey'] = 'value'
jobb.input['notabcompletion'] = -1
```

I mean, sure, it doesn't crash anything...but it's hella ugly 😉

liamhuber (approved):
I suspect the failing tests are related to #1202. If so, you can do like Jan did over in #1199 and just force the pyscal version to be "=" instead of ">=" and this should resolve it.

Whenever the tests are passing I'm happy to see this merged 👍

PR #1080, opened 2020-10-21
Nbands convergence check (vasp)

sravankumarkb:
Hi all,

Here is a function I would like to have in vasp base class. This function allows me to check if the calculation is converged with respect to the nbands. It would be great if you could let me know your feedback before merging into the master. Thanks.


pyiron/vasp/base.py
```diff
@@ -400,6 +400,12 @@ def collect_output(self):
                 exclude_nodes=self._exclude_nodes_hdf,
             )
 
+    def nbands_convergence_check(self):
+        if (max(self["output/electronic_structure/occ_matrix"][:,-1])>0):
```
samwaseda, line 404: I see only two indices but shouldn't there be three? I presume the first is k-points and the second is bands. Where's spin?

samwaseda: And I guess we can make it shorter:

```
def nbands_convergence_check(self):
    return np.all(np.isclose(self["output/electronic_structure/occ_matrix"][:,-1], 0))
```

samwaseda: Oh wait there should be even four indices since there should be also ionic steps.

sravankumarkb: Thanks Sam for the comments, I ran a couple of calculations with spin and more than one ionic steps, the shape of the occupancy matrix does not change and remains the same as (kpoints, nbands).  I have made the function shorter as you suggested and added a docstring. 
If this satisfies your comments, you can resolve the conversation. Thanks.

samwaseda: Ok I created a new issue and hopefully @sudarsan-surendralal will reply soon. Feel free to merge it now or after Sudarsan corrects the bug.

sudarsan-surendralal: The shape in `output/electronic_structure` should be `n_spin x n_kpoints x n_bands`. This is the electronic structure of the final step only. The spin not being there is a bug!

samwaseda: I agree that the most interesting one is the final occupancy, but shouldn't we also check the occupancies of all steps? This being said, I admit it's possible that the amount of data to store might become too large if there are too many steps...

sudarsan-surendralal:
@sravankumarkb Can you now merge the master branch into this branch and try Sam's suggestions?

sravankumarkb:
> @sravankumarkb Can you now merge the master branch into this branch and try Sam's suggestions?

Thank you @sudarsan-surendralal for updating the electronic structure shape in the output to include the spin. I have merged the master branch and updated the nbands_convergence_check function to reflect the suggestions of @samwaseda 

Now the nbands_convergence_check function works for both spin polarized and non spin polarized calcs. 
If you guys are happy with the current function, maybe we can consider merging it into the master? 
Thanks.

pyiron/vasp/base.py
```diff
@@ -403,6 +403,16 @@ def collect_output(self):
...
+        """
+        Function to check the convergence of nbands in the Vasp calculation
+
+        Returns:
+
+            bool : True if the highest band is unoccupied, False if the highest band is occupied
+        """
+        return np.all(np.isclose(self["output/electronic_structure/occ_matrix"][:,:,-1], 0)) #shape is n_spin x n_kpoints x n_bands
+
     def convergence_check(self):
```
sudarsan-surendralal, line 416: Nice implementation. Could you also call `nbands_convergence_check` inside `convergence_check` and then set the status to `not_converged` if the bands aren't converged?

sravankumarkb: Thanks @sudarsan-surendralal  for the suggestion. I will start working on it

sudarsan-surendralal: Actually, can you move this function to `GenericDFTJob`? Since this is useful for both VASP and SPHINX

jan-janssen (changes requested):
Please add a unit test and currently the function is not used by default is this intended? Or should the function be used to check the result is converged? Meaning if it returns true the status should be finished and if it returns false the status should be not converged?

samwaseda:
I added a test

sravankumarkb:
> I added a test

Thank you @samwaseda :)

sravankumarkb:
> Please add a unit test and currently the function is not used by default is this intended? Or should the function be used to check the result is converged? Meaning if it returns true the status should be finished and if it returns false the status should be not converged?

Thanks @jan-janssen for the suggestion. I have not thought about making the function default and therefore setting the job status, could you please let me know what I need to add to make it default? Thanks.

samwaseda:
> Thanks @jan-janssen for the suggestion. I have not thought about making the function default and therefore setting the job status, could you please let me know what I need to add to make it default? Thanks.

Since we don't have a proper error/warning handling, I would say it's okay for now that this function is not called automatically for now. In addition, I think it's too severe if we consider it as not converged only because there's no empty band left.

sravankumarkb:
> > Thanks @jan-janssen for the suggestion. I have not thought about making the function default and therefore setting the job status, could you please let me know what I need to add to make it default? Thanks.
> 
> Since we don't have a proper error/warning handling, I would say it's okay for now that this function is not called automatically for now. In addition, I think it's too severe if we consider it as not converged only because there's no empty band left.

Thanks @samwaseda for the clarification. Does that mean that we can merge this or is there anything else I need to do to get this merged? Thanks.

samwaseda (approved):

jan-janssen:
Please transfer the pull request to https://github.com/pyiron/pyiron_atomistics

Issue #1074, opened 2020-10-20
Pressure setting in LAMMPS MD fails with "None" entries in list


t-brink:
I set up a LAMMPS job like this:

```
job = pr.create_job(pr.job_type.Lammps, "test1")
job.structure = ...
job.potential = ...
job.calc_md(
    temperature=1000.0,
    temperature_damping_timescale=100.0,
    pressure=[0.0, 0.0, 0.0, None, None, None], # <-------------------
    pressure_damping_timescale=500.0,
    time_step=1.0,
    n_ionic_steps=100000,
)
```

In the `control.inp` file this is erroneously translated to the following fix:

```
fix ensemble all npt temp 1000.0 1000.0 0.1 x 0.0 0.0 0.5 y 0.0 0.0 0.5 z 0.0 0.0 0.5 xy nan nan 0.5 xz nan nan 0.5 yz nan nan 0.5
```

while it should omit the `xy`, `xz`, and `yz` components.

I tried with the newest git version from the main branch.

t-brink:
I dug into this and the problem is that `pyiron.lammps.control.LammpsControl.pressure_to_lammps()` contains the following conversion from the pressure input list:

```
pressure = np.array(all_pressures, dtype=float)
```

The `dtype=float` will convert all `None`s to `nan`s and in `pyiron.lammps.control.LammpsControl.calc_md()` in [line 516](https://github.com/pyiron/pyiron/blob/595bbc0a53ab1e841e826142e4fcbc2e3bc6691d/pyiron/lammps/control.py#L516), there is a check `if value is not None:`, which will evaluate to `True` for `nan`. The same might be true for `calc_minimize()`, but I didn't encounter this, yet. Probably `pressure_to_lammps()` needs to return a list with the `None` values intact? Needs careful checking how that interacts with the rotation, so I hesitate to provide a naive patch.

pmrv:
Relevant previous commits:
https://github.com/pyiron/pyiron/commit/f95e3b086262c470520123cdad0f144f9843b97d
https://github.com/pyiron/pyiron/commit/75f1bc9455b3cb96e9db569dae35324094777c0a

liamhuber:
I think @samwaseda is our rotations master?

t-brink:
I discusses with @pmrv and I will propose a pull request soon. It seems trivial to fix without disturbing anything.

samwaseda:
[...]

PR #1017, opened 2020-10-01
Fix Lammps pressure

jan-janssen:

Issue #1016, opened 2020-10-01
Full relaxation with LAMMPS for triclinic cell


yury-lysogorskiy:
I want to do the full relaxation of triclinic cell (primitive FCC) with LAMMPS:
```
    relaxation_job = subpr.create_job(pr.job_type.Lammps, "opt")
    relaxation_job.structure = initial_structure
    relaxation_job.potential=potential        
    relaxation_job.calc_minimize(pressure=0.0)
```
But I got the error:
``` 
ValueError: Cells which are not orthorhombic or rectangular are incompatible with Lammps constant pressure calculations unless the entire pressure tensor is defined.The reason is that Lammps demands such cells be represented with an upper-triangular unit cell, thus a rotation between Lammps and pyiron coordinateframes is required; it is not possible to rotate the pressure tensor if any of its components is None.
```

How should I specify the `pressure` parameter in order to have **full relaxation**?

PR #1083, opened 2020-10-21
Fix LAMMPS pressure

t-brink:
As discussed in bug #1074, a LAMMPS job converted the `pressure` input to a numpy array of type float, thereby converting `None` values to NaN, which was not checked correctly by later parts of the code.

Furthermore, hydrostatic pressures were handled incorrectly, by converting a scalar pressure p to a stress tensor [p,p,p,p,p,p] instead of [p,p,p,0,0,0] or [p,p,p,None,None,None] in some circumstances.

To be discussed: I believe that hydrostatic pressures can be applied correctly in all cases (even with rotation) by using [p,p,p,None,None,None], i.e., keeping the shear components fixed. I am not 100% sure it works when rotating the box, so if there is no consensus I would propose to disallow even hydrostatic pressure when the box has to be rotated and require a full stress tensor. The previous code was inconsistent in the different cases, which is worse.

### DIFF
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -208,48 +208,44 @@ class LammpsControl(GenericParameters):
             rotation_matrix (numpy.ndarray): The 3x3 matrix rotating from the pyiron to Lammps coordinate frame.
 
         Returns:
-            (numpy.ndarray): pxx, pyy, pzz, pxy, pxz, and pyz to be passed to Lammps.
+            (list): pxx, pyy, pzz, pxy, pxz, and pyz to be passed to Lammps.
         """
 
         # Get something with six elements
         if not hasattr(pressure, "__len__"):
-            if not np.isclose(np.matrix.trace(rotation_matrix), 3) and pressure == 0.0:
-                pressure = np.append(pressure * np.ones(3), pressure * np.ones(3))
-            else:
-                pressure = np.append(pressure * np.ones(3), 3 * [None])
+            # This is a hydrostatic stress. It is invariant under rotation, so no transformation needed.
+            pressure = [float(pressure), float(pressure), float(pressure), None, None, None]
         else:
             if len(pressure) > 6:
                 raise ValueError("Pressure can have a maximum of 6 values, (x, y, z, xy, xz, and yz), but got " +
                                  "{}".format(len(pressure)))
-            all_pressures = 6 * [None]
-            all_pressures[:len(pressure)] = pressure
-            pressure = np.array(all_pressures, dtype=float)
+            pressure = list(pressure) + (6 - len(pressure)) * [None]
 
-        # Cell degrees of freedom can be kept fixed using None for the pressure, check where that's done.
-        not_none_mask = (pressure!=None)
-        if not np.any(not_none_mask):
-            raise ValueError("Pressure cannot have a length but all be None")
+            if all(p is None for p in pressure):
+                raise ValueError("Pressure cannot have a length but all be None")
 
-        # If necessary, rotate the pressure tensor to the Lammps coordinate frame
-        if not np.isclose(np.matrix.trace(rotation_matrix), 3):
-            if not np.all(not_none_mask):
-                raise ValueError("Cells which are not orthorhombic or rectangular are incompatible with Lammps "
-                                 "constant pressure calculations unless the entire pressure tensor is defined."
-                                 "The reason is that Lammps demands such cells be represented with an "
-                                 "upper-triangular unit cell, thus a rotation between Lammps and pyiron coordinate"
-                                 "frames is required; it is not possible to rotate the pressure tensor if any of "
-                                 "its components is None.")
-            pxx, pyy, pzz, pxy, pxz, pyz = pressure
-            pressure_tensor = np.array(
-                [[pxx, pxy, pxz],
-                 [pxy, pyy, pyz],
-                 [pxz, pyz, pzz]]
-            )
-            lammps_pressure_tensor = rotation_matrix.T @ pressure_tensor @ rotation_matrix
-            pressure = lammps_pressure_tensor[[0, 1, 2, 0, 0, 1], [0, 1, 2, 1, 2, 2]]
+            # If necessary, rotate the pressure tensor to the Lammps coordinate frame
+            if not np.isclose(np.matrix.trace(rotation_matrix), 3):
+                if any(p is None for p in pressure):
+                    raise ValueError("Cells which are not orthorhombic or an upper-triangular cell are incompatible with Lammps "
+                                     "constant pressure calculations unless the entire pressure tensor is defined."
+                                     "The reason is that Lammps demands such cells be represented with an "
+                                     "upper-triangular unit cell, thus a rotation between Lammps and pyiron coordinate"
+                                     "frames is required; it is not possible to rotate the pressure tensor if any of "
+                                     "its components is None.")
+                pxx, pyy, pzz, pxy, pxz, pyz = pressure
+                pressure_tensor = np.array(
+                    [[pxx, pxy, pxz],
+                     [pxy, pyy, pyz],
+                     [pxz, pyz, pzz]]
+                )
+                lammps_pressure_tensor = rotation_matrix.T @ pressure_tensor @ rotation_matrix
+                pressure = list(lammps_pressure_tensor[[0, 1, 2, 0, 0, 1], [0, 1, 2, 1, 2, 2]])
 
-        pressure[not_none_mask] *= LAMMPS_UNIT_CONVERSIONS[self["units"]]["pressure"]
-        return pressure
+        return [(p * LAMMPS_UNIT_CONVERSIONS[self["units"]]["pressure"]
+                 if p is not None
+                 else p)
+                for p in pressure]
 
     def calc_minimize(
         self,
@@ -300,18 +296,12 @@ class LammpsControl(GenericParameters):
         ionic_force_tolerance *= force_units
 
         if pressure is not None:
-            if None in np.array([pressure]).flatten():
-                if rotation_matrix is not None and not np.isclose(np.linalg.det(rotation_matrix), 1):
-                    raise AssertionError('Pressure cannot contain None if upper triangle in cell is defined')
-            elif rotation_matrix is not None:
-                pressure = self.pressure_to_lammps(pressure, rotation_matrix)
+            pressure = self.pressure_to_lammps(pressure, rotation_matrix)
             str_press = ""
             for press, str_axis in zip(pressure, [" x ", " y ", " z ", " xy ", " xz ", " yz "]):
                 if press is not None:
                     str_press += str_axis + str(press*pressure_units )
-            if len(str_press) == 0:
-                raise ValueError("Pressure values cannot all be None")
-            elif len(str_press) > 1:
+            if len(str_press) > 1:
                 str_press += " couple none"
             self.set(fix___ensemble=r"all box/relax" + str_press)
         self.remove_keys(["fix___nve"])
@@ -498,18 +488,10 @@ class LammpsControl(GenericParameters):
             
         # Set thermodynamic ensemble
         if pressure is not None:  # NPT
-
-            if np.sum(np.array([pressure]).flatten()!=None)==0:
-                raise ValueError("At least one component of pressure must be other than None")
-
-            if hasattr(pressure, "__len__") and len(pressure) > 6:
-                raise ValueError("Pressure must be a float or a vector with length <= 6")
-
             if temperature is None or temperature.min() <= 0:
                 raise ValueError("Target temperature for fix nvt/npt/nph cannot be 0 or negative")
 
-            if rotation_matrix is not None:
-                pressure = self.pressure_to_lammps(pressure, rotation_matrix)
+            pressure = self.pressure_to_lammps(pressure, rotation_matrix)
 
             pressure_string = ""
             for coord, value in zip(["x", "y", "z", "xy", "xz", "yz"], pressure):
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -315,7 +315,6 @@ class LammpsControl(GenericParameters):
             raise NotImplementedError
         energy_units = LAMMPS_UNIT_CONVERSIONS[self["units"]]["energy"]
         force_units = LAMMPS_UNIT_CONVERSIONS[self["units"]]["force"]
-        pressure_units = LAMMPS_UNIT_CONVERSIONS[self["units"]]["pressure"]
 
         ionic_energy_tolerance *= energy_units
         ionic_force_tolerance *= force_units
@@ -328,7 +327,7 @@ class LammpsControl(GenericParameters):
                 str_press = ""
                 for press, str_axis in zip(pressure, [" x ", " y ", " z ", " xy ", " xz ", " yz "]):
                     if press is not None:
-                        str_press += str_axis + str(press*pressure_units )
+                        str_press += str_axis + str(press)
                 if len(str_press) > 1:
                     str_press += " couple none"
             self.set(fix___ensemble=r"all box/relax" + str_press)
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -242,7 +242,7 @@ class LammpsControl(GenericParameters):
                  [pxz, pyz, pzz]]
             )
             lammps_pressure_tensor = rotation_matrix.T @ pressure_tensor @ rotation_matrix
-                pressure = list(lammps_pressure_tensor[[0, 1, 2, 0, 0, 1], [0, 1, 2, 1, 2, 2]])
+            pressure = list(lammps_pressure_tensor[[0, 1, 2, 0, 0, 1], [0, 1, 2, 1, 2, 2]])
 
         return [(p * LAMMPS_UNIT_CONVERSIONS[self["units"]]["pressure"]
                  if p is not None
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -222,11 +222,13 @@ class LammpsControl(GenericParameters):
         if np.isscalar(pressure):
             return float(pressure) * LAMMPS_UNIT_CONVERSIONS[self["units"]]["pressure"]
 
-        # Normalize pressure to a list of 6 entries.
+        # Normalize pressure to a list of 6 entries of either float or NoneType type.
         if len(pressure) > 6:
             raise ValueError("Pressure can have a maximum of 6 values, (x, y, z, xy, xz, and yz), but got " +
                              "{}".format(len(pressure)))
-        pressure = list(pressure) + (6 - len(pressure)) * [None]
+        pressure = [float(p) if p is not None else None
+                    for p in pressure]
+        pressure += (6 - len(pressure)) * [None]
 
         if all(p is None for p in pressure):
             raise ValueError("Pressure cannot have a length but all be None")
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -201,6 +201,9 @@ class LammpsControl(GenericParameters):
         Lammps handles complex cells in a particular way, namely by using an upper triangular cell. This means we may
         need to convert our pressure tensor to a new coordinate frame. We also handle that transformation here.
 
+        I case of a single pressure value, it is again returned as a single pressure value, to be used with the "iso"
+        option (i.e., coupled deformation in x, y, and z).
+
         Finally, we also ensure that the units are converted from pyiron's GPa to whatever Lammps needs.
 
         Args:
@@ -209,30 +212,46 @@ class LammpsControl(GenericParameters):
 
         Returns:
             (list): pxx, pyy, pzz, pxy, pxz, and pyz to be passed to Lammps.
+
+            or
+
+            (float): a single, isotropic pressure to be used with the "iso" option
         """
 
-        # Get something with six elements
-        if not hasattr(pressure, "__len__"):
-            if np.isclose(pressure, 0.0):
-                pressure = 6*[0.0]
-            else:
-                pressure = [float(pressure), float(pressure), float(pressure), None, None, None]
-        else:
-            if len(pressure) > 6:
-                raise ValueError("Pressure can have a maximum of 6 values, (x, y, z, xy, xz, and yz), but got " +
-                                 "{}".format(len(pressure)))
-            pressure = list(pressure) + (6 - len(pressure)) * [None]
+        # If pressure is a scalar, only unit conversion is needed.
+        if np.isscalar(pressure):
+            return float(pressure) * LAMMPS_UNIT_CONVERSIONS[self["units"]]["pressure"]
+
+        # Normalize pressure to a list of 6 entries.
+        if len(pressure) > 6:
+            raise ValueError("Pressure can have a maximum of 6 values, (x, y, z, xy, xz, and yz), but got " +
+                             "{}".format(len(pressure)))
+        pressure = list(pressure) + (6 - len(pressure)) * [None]
 
         if all(p is None for p in pressure):
             raise ValueError("Pressure cannot have a length but all be None")
 
-        # If necessary, rotate the pressure tensor to the Lammps coordinate frame
-        if not np.isclose(np.matrix.trace(rotation_matrix), 3):
+        # If necessary, rotate the pressure tensor to the Lammps coordinate frame.
+        # Isotropic, hydrostatic pressures are rotation invariant.
+        is_isotropic_hydrostatic = (
+            # All diagonal components must be equal and not None.
+            pressure[0] is not None and pressure[1] is not None and pressure[2] is not None
+            and np.isclose(pressure[0], pressure[1]) and np.isclose(pressure[1], pressure[2])
+            # Either all shear components must be None, or all shear components must be zero.
+            and (
+                (pressure[3] is None and pressure[4] is None and pressure[5] is None)
+                or
+                (pressure[3] is not None and np.isclose(pressure[3], 0.0) and
+                 pressure[4] is not None and np.isclose(pressure[4], 0.0) and
+                 pressure[5] is not None and np.isclose(pressure[5], 0.0))
+            )
+        )
+        if not np.isclose(np.matrix.trace(rotation_matrix), 3) and not is_isotropic_hydrostatic:
             if any(p is None for p in pressure):
                 raise ValueError("Cells which are not orthorhombic or an upper-triangular cell are incompatible with Lammps "
-                                 "constant pressure calculations unless the entire pressure tensor is defined."
+                                 "constant pressure calculations unless the entire pressure tensor is defined. "
                                  "The reason is that Lammps demands such cells be represented with an "
-                                 "upper-triangular unit cell, thus a rotation between Lammps and pyiron coordinate"
+                                 "upper-triangular unit cell, thus a rotation between Lammps and pyiron coordinate "
                                  "frames is required; it is not possible to rotate the pressure tensor if any of "
                                  "its components is None.")
             pxx, pyy, pzz, pxy, pxz, pyz = pressure
@@ -273,9 +292,11 @@ class LammpsControl(GenericParameters):
                 `max_iter` steps, terminate at the converged step. If the minimisation does not converge up to
                 `max_iter` steps, terminate at the `max_iter` step. (Default is 100000.)
             pressure (None/float/numpy.ndarray/list): Target pressure. If set to None, an isochoric (constant V)
-                calculation is performed. A list of up to length 6 can be given to specify xx, yy, zz, xy, xz, and yz
-                components of the pressure tensor, respectively. These values can mix floats and `None` to allow only
-                certain degrees of cell freedom to change. (Default is None, run isochorically.)
+                calculation is performed. If set to a scalar, the shear of the cell and the ratio of the x, y, and
+                z components is kept constant, while an isotropic, hydrostatic pressure is applied. A list of up to
+                length 6 can be given to specify xx, yy, zz, xy, xz, and yz components of the pressure tensor,
+                respectively. These values can mix floats and `None` to allow only certain degrees of cell freedom
+                to change. (Default is None, run isochorically.)
             n_print (int): Write (dump or print) to the output file every n steps (Default: 100)
             style ('cg'/'sd'/other values from Lammps docs): The style of the numeric minimization, either conjugate
                 gradient, steepest descent, or other keys permissible from the Lammps docs on 'min_style'. (Default
@@ -299,12 +320,15 @@ class LammpsControl(GenericParameters):
 
         if pressure is not None:
             pressure = self.pressure_to_lammps(pressure, rotation_matrix)
-            str_press = ""
-            for press, str_axis in zip(pressure, [" x ", " y ", " z ", " xy ", " xz ", " yz "]):
-                if press is not None:
-                    str_press += str_axis + str(press*pressure_units )
-            if len(str_press) > 1:
-                str_press += " couple none"
+            if np.isscalar(pressure):
+                str_press = " iso {}".format(pressure)
+            else:
+                str_press = ""
+                for press, str_axis in zip(pressure, [" x ", " y ", " z ", " xy ", " xz ", " yz "]):
+                    if press is not None:
+                        str_press += str_axis + str(press*pressure_units )
+                if len(str_press) > 1:
+                    str_press += " couple none"
             self.set(fix___ensemble=r"all box/relax" + str_press)
         self.remove_keys(["fix___nve"])
         self.set(min_style=style)
@@ -411,9 +435,11 @@ class LammpsControl(GenericParameters):
                                            temperature and the final target temperature (in between the target value
                                            is varied linearly).
             pressure (None/float/numpy.ndarray/list): Target pressure. If set to None, an NVE or an NVT calculation is
-                performed. A list of up to length 6 can be given to specify xx, yy, zz, xy, xz, and yz components of
-                the pressure tensor, respectively. These values can mix floats and `None` to allow only certain degrees
-                of cell freedom to change. (Default is None, run isochorically.)
+                performed. If set to a scalar, the shear of the cell and the ratio of the x, y, and z components is kept
+                constant, while an isotropic, hydrostatic pressure is applied. A list of up to length 6 can be given to
+                specify xx, yy, zz, xy, xz, and yz components of the pressure tensor, respectively. These values can mix
+                floats and `None` to allow only certain degrees of cell freedom to change. (Default is None, run
+                isochorically.)
             n_ionic_steps (int): Number of ionic steps
             time_step (float): Step size in fs between two steps.
             n_print (int):  Print frequency
@@ -495,12 +521,15 @@ class LammpsControl(GenericParameters):
 
             pressure = self.pressure_to_lammps(pressure, rotation_matrix)
 
-            pressure_string = ""
-            for coord, value in zip(["x", "y", "z", "xy", "xz", "yz"], pressure):
-                if value is not None:
-                    pressure_string += " {0} {1} {1} {2}".format(
-                        coord, str(value), str(pressure_damping_timescale)
-                    )
+            if np.isscalar(pressure):
+                pressure_string = " iso {0} {0} {1}".format(pressure, pressure_damping_timescale)
+            else:
+                pressure_string = ""
+                for coord, value in zip(["x", "y", "z", "xy", "xz", "yz"], pressure):
+                    if value is not None:
+                        pressure_string += " {0} {1} {1} {2}".format(
+                            coord, value, pressure_damping_timescale
+                        )
 
             if langevin:  # NPT(Langevin)
                 fix_ensemble_str = "all nph" + pressure_string
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -213,33 +213,35 @@ class LammpsControl(GenericParameters):
 
         # Get something with six elements
         if not hasattr(pressure, "__len__"):
-            # This is a hydrostatic stress. It is invariant under rotation, so no transformation needed.
-            pressure = [float(pressure), float(pressure), float(pressure), None, None, None]
+            if np.isclose(pressure, 0.0):
+                pressure = 6*[0.0]
+            else:
+                pressure = [float(pressure), float(pressure), float(pressure), None, None, None]
         else:
             if len(pressure) > 6:
                 raise ValueError("Pressure can have a maximum of 6 values, (x, y, z, xy, xz, and yz), but got " +
                                  "{}".format(len(pressure)))
             pressure = list(pressure) + (6 - len(pressure)) * [None]
 
-            if all(p is None for p in pressure):
-                raise ValueError("Pressure cannot have a length but all be None")
+        if all(p is None for p in pressure):
+            raise ValueError("Pressure cannot have a length but all be None")
 
-            # If necessary, rotate the pressure tensor to the Lammps coordinate frame
-            if not np.isclose(np.matrix.trace(rotation_matrix), 3):
-                if any(p is None for p in pressure):
-                    raise ValueError("Cells which are not orthorhombic or an upper-triangular cell are incompatible with Lammps "
-                                     "constant pressure calculations unless the entire pressure tensor is defined."
-                                     "The reason is that Lammps demands such cells be represented with an "
-                                     "upper-triangular unit cell, thus a rotation between Lammps and pyiron coordinate"
-                                     "frames is required; it is not possible to rotate the pressure tensor if any of "
-                                     "its components is None.")
-                pxx, pyy, pzz, pxy, pxz, pyz = pressure
-                pressure_tensor = np.array(
-                    [[pxx, pxy, pxz],
-                     [pxy, pyy, pyz],
-                     [pxz, pyz, pzz]]
-                )
-                lammps_pressure_tensor = rotation_matrix.T @ pressure_tensor @ rotation_matrix
+        # If necessary, rotate the pressure tensor to the Lammps coordinate frame
+        if not np.isclose(np.matrix.trace(rotation_matrix), 3):
+            if any(p is None for p in pressure):
+                raise ValueError("Cells which are not orthorhombic or an upper-triangular cell are incompatible with Lammps "
+                                 "constant pressure calculations unless the entire pressure tensor is defined."
+                                 "The reason is that Lammps demands such cells be represented with an "
+                                 "upper-triangular unit cell, thus a rotation between Lammps and pyiron coordinate"
+                                 "frames is required; it is not possible to rotate the pressure tensor if any of "
+                                 "its components is None.")
+            pxx, pyy, pzz, pxy, pxz, pyz = pressure
+            pressure_tensor = np.array(
+                [[pxx, pxy, pxz],
+                 [pxy, pyy, pyz],
+                 [pxz, pyz, pzz]]
+            )
+            lammps_pressure_tensor = rotation_matrix.T @ pressure_tensor @ rotation_matrix
                 pressure = list(lammps_pressure_tensor[[0, 1, 2, 0, 0, 1], [0, 1, 2, 1, 2, 2]])
 
         return [(p * LAMMPS_UNIT_CONVERSIONS[self["units"]]["pressure"]
--- a/tests/lammps/test_control.py
+++ b/tests/lammps/test_control.py
@@ -48,12 +48,12 @@ class TestLammps(unittest.TestCase):
 
     def test_pressure_to_lammps(self):
         lc = LammpsControl()
-        # Correct normalization without rotation.
+        # Correct normalization without rotation. Note that we convert from GPa to bar for LAMMPS.
         no_rot = np.identity(3)
-        self.assertEqual(lc.pressure_to_lammps(1.0, no_rot), [1.0, 1.0, 1.0, None, None, None])
-        self.assertEqual(lc.pressure_to_lammps([1.0, 2.0, 3.0], no_rot), [1.0, 2.0, 3.0, None, None, None])
-        self.assertEqual(lc.pressure_to_lammps([1.0, 2.0, 3.0, None, None, None], no_rot), [1.0, 2.0, 3.0, None, None, None])
-        self.assertEqual(lc.pressure_to_lammps([None, None, None, None, None, 2.0], no_rot), [None, None, None, None, None, 2.0])        
+        self.assertEqual(lc.pressure_to_lammps(1.0, no_rot), [10000.0, 10000.0, 10000.0, None, None, None])
+        self.assertEqual(lc.pressure_to_lammps([1.0, 2.0, 3.0], no_rot), [10000.0, 20000.0, 30000.0, None, None, None])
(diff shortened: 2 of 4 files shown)


samwaseda (approved):
And @t-brink was right. This is correct :) (actually I was not aware of the `ValueError` that was added in line 230)

pyiron/lammps/control.py
```diff
@@ -208,48 +208,44 @@ def pressure_to_lammps(self, pressure, rotation_matrix):
...
         """
 
         # Get something with six elements
         if not hasattr(pressure, "__len__"):
-            if not np.isclose(np.matrix.trace(rotation_matrix), 3) and pressure == 0.0:
-                pressure = np.append(pressure * np.ones(3), pressure * np.ones(3))
-            else:
-                pressure = np.append(pressure * np.ones(3), 3 * [None])
+            # This is a hydrostatic stress. It is invariant under rotation, so no transformation needed.
+            pressure = [float(pressure), float(pressure), float(pressure), None, None, None]
```
samwaseda, line 217: Maybe it's more like a style issue, but I might rather put `3*[float(pressure)]+3*[None]`

jan-janssen (changes requested):
This reverts https://github.com/pyiron/pyiron/pull/1017 so it reopens https://github.com/pyiron/pyiron/issues/1016

t-brink:
> This reverts #1017 so it reopens #1016

The old commit does not have an explanation, can you explain the rationale of `pressure = np.append(pressure * np.ones(3), pressure * np.ones(3))` to me? Why are the shear components set to the hydrostatic stress value? Or is a single scalar not meant to represent a hydrostatic stress? Am I getting sleepy and missing something?

samwaseda:
> The old commit does not have an explanation, can you explain the rationale of `pressure = np.append(pressure * np.ones(3), pressure * np.ones(3))` to me? Why are the shear components set to the hydrostatic stress value? Or is a single scalar not meant to represent a hydrostatic stress? Am I getting sleepy and missing something?

That's mostly relevant for pressure=0, because no one really does a finite pressure calculation. I guess we have to discuss whether we'd like to include the shear component as well.

pyiron/lammps/control.py
```diff
@@ -208,48 +208,44 @@ def pressure_to_lammps(self, pressure, rotation_matrix):
...
+            (list): pxx, pyy, pzz, pxy, pxz, and pyz to be passed to Lammps.
         """
 
         # Get something with six elements
         if not hasattr(pressure, "__len__"):
-            if not np.isclose(np.matrix.trace(rotation_matrix), 3) and pressure == 0.0:
-                pressure = np.append(pressure * np.ones(3), pressure * np.ones(3))
-            else:
-                pressure = np.append(pressure * np.ones(3), 3 * [None])
+            # This is a hydrostatic stress. It is invariant under rotation, so no transformation needed.
```
samwaseda, line 216: This comment is wrong.

t-brink:
> > The old commit does not have an explanation, can you explain the rationale of `pressure = np.append(pressure * np.ones(3), pressure * np.ones(3))` to me? Why are the shear components set to the hydrostatic stress value? Or is a single scalar not meant to represent a hydrostatic stress? Am I getting sleepy and missing something?
> 
> That's mostly relevant for pressure=0, because no one really does a finite pressure calculation. I guess we have to discuss whether we'd like to include the shear component as well.

But then the correct code is `pressure = np.append(pressure * np.ones(3), np.zeros(3))` or something along that line. The old version is wrong for all but pressure=0.0. Also, it kind of mixed both versions of doing things and sometimes used `None` for the shear.

The desired behavior can be discussed, of course. The documentation should be updated to be clear what this means. I am fine with either default, one can always override it.

BTW, bug #1016 should not triggered by the new code, I tested this quickly and the exception in the bug report is not raised.

jan-janssen:
To my understanding with changes the following code does not work: 

```
job = pr.create_job(pr.job_type.Lammps, "lmp")
job.structure = pr.create_ase_bulk("Al")
job.calc_minimize(pressure=0.0)
job.run()
```

pyiron/lammps/control.py
t-brink: This can be, I am also not sure about it. Can you elaborate?

samwaseda: It's just if the box has to be rotated the pressure has to be rotated as well...

t-brink:
> To my understanding with changes the following code does not work:
> 
> ```
> job = pr.create_job(pr.job_type.Lammps, "lmp")
> job.structure = pr.create_ase_bulk("Al")
> job.calc_minimize(pressure=0.0)
> job.run()
> ```

Please try, it works for me. To my understanding the old code ran without raising an exception, but the stress tensor is not hydorstatic ([[p,0,0],[0,p,0],[0,0,p]]) but instead [[p,p,p],[p,p,p],[p,p,p]] which makes no sense to me.

samwaseda:
> > To my understanding with changes the following code does not work:
> > ```
> > job = pr.create_job(pr.job_type.Lammps, "lmp")
> > job.structure = pr.create_ase_bulk("Al")
> > job.calc_minimize(pressure=0.0)
> > job.run()
> > ```
> 
> Please try, it works for me. To my understanding the old code ran without raising an exception, but the stress tensor is not hydorstatic ([[p,0,0],[0,p,0],[0,0,p]]) but instead [[p,p,p],[p,p,p],[p,p,p]] which makes no sense to me.

No it shouldn't work, because the cell has components in the upper triangle matrix, but I thought this was exactly the behaviour you wanted.

jan-janssen:
We need this code to work for the database in http://atomistictools.org previously it was implemented based on: 
```
fix 1 all box/relax iso 0.0
```

t-brink:
OK, let me try to untangle the issues:

* Should a settings of `pressure=p` result in a stress tensor [p,p,p,0,0,0] or [p,p,p,None,None,None]? Both are fine with me. What I am not fine with is [p,p,p,p,p,p] unless I am missing something. Probably [p,p,p,0,0,0] is the better version.

* Can [p,p,p,None,None,None] also work for rotated boxes? I am not sure. It might work in the special case of hydrostatic pressure, since it is rotation invariant. But it might not for additional reasons. We should probably forbid this if we are not sure.

* Bug #1016 would definitely be fixed by  `pressure=p` resulting in a stress tensor [p,p,p,0,0,0] in any case.

Do I have this correct?

samwaseda:
> OK, let me try to untangle the issues:
> 
> * Should a settings of `pressure=p` result in a stress tensor [p,p,p,0,0,0] or [p,p,p,None,None,None]? Both are fine with me. What I am not fine with is [p,p,p,p,p,p] unless I am missing something. Probably [p,p,p,0,0,0] is the better version.

I guess what Jan is saying is [p, p, p, None, None, None] if p!=0. and otherwise [0,0,0,0,0,0] (which is what the code did before this PR)

> * Can [p,p,p,None,None,None] also work for rotated boxes? I am not sure. It might work in the special case of hydrostatic pressure, since it is rotation invariant. But it might not for additional reasons. We should probably forbid this if we are not sure.

No. That's what the `ValueError` forbids (which I wasn't aware of).

samwaseda:
And finally the hydrostatic pressure is not rotation invariant...

jan-janssen:
> > To my understanding with changes the following code does not work:
> > ```
> > job = pr.create_job(pr.job_type.Lammps, "lmp")
> > job.structure = pr.create_ase_bulk("Al")
> > job.calc_minimize(pressure=0.0)
> > job.run()
> > ```
> 
> Please try, it works for me. To my understanding the old code ran without raising an exception, but the stress tensor is not hydorstatic ([[p,0,0],[0,p,0],[0,0,p]]) but instead [[p,p,p],[p,p,p],[p,p,p]] which makes no sense to me.

For me it results in: 
```
fix ensemble all box/relax x 0.0 y 0.0 z 0.0 couple none
```
I would prefer: 
```
fix ensemble all box/relax iso 0.0
```
But both is fine, unless we have complex systems.

pyiron/lammps/control.py
```diff
@@ -208,32 +208,28 @@ def pressure_to_lammps(self, pressure, rotation_matrix):
...
             else:
-                pressure = np.append(pressure * np.ones(3), 3 * [None])
+                pressure = [float(pressure), float(pressure), float(pressure), None, None, None]
         else:
             if len(pressure) > 6:
                 raise ValueError("Pressure can have a maximum of 6 values, (x, y, z, xy, xz, and yz), but got " +
                                  "{}".format(len(pressure)))
-            all_pressures = 6 * [None]
-            all_pressures[:len(pressure)] = pressure
-            pressure = np.array(all_pressures, dtype=float)
```
samwaseda, line 226: Sorry actually quite a lot of lines have been changed but wouldn't it have been enough to erase just `dtype=float`?

t-brink:
I updated. I restored the behavior where pressure=0.0 is different from pressure=123.0, but I am not sure I like it.

Regarding the coupling of the different stress components, maybe that needs to be user-controllable. Special-casing to "iso" needs some more involved code changes, I guess it depends on the expected API. The documentation doesn't say what should happen. I could do it if this is desired.

pyiron/lammps/control.py
```diff
@@ -300,18 +296,12 @@ def calc_minimize(
...
-            if None in np.array([pressure]).flatten():
-                if rotation_matrix is not None and not np.isclose(np.linalg.det(rotation_matrix), 1):
-                    raise AssertionError('Pressure cannot contain None if upper triangle in cell is defined')
-            elif rotation_matrix is not None:
-                pressure = self.pressure_to_lammps(pressure, rotation_matrix)
+            pressure = self.pressure_to_lammps(pressure, rotation_matrix)
             str_press = ""
             for press, str_axis in zip(pressure, [" x ", " y ", " z ", " xy ", " xz ", " yz "]):
                 if press is not None:
                     str_press += str_axis + str(press*pressure_units )
```
liamhuber, line 305: Wait a second, are we converting pressure units twice? `pressure` should already have the right units coming out of `self.pressure_to_lammps`...

I just double-checked the master branch and it sure looks to me like we're double converting, but I also resonate with Tobias's earlier comment about maybe being sleepy 😅 Is anyone else seeing what I'm seeing?

pyiron/lammps/control.py
samwaseda: ok I just tested it on my local branch and saw that the only one change you need is to remove `dtype=float`. You don't need anything else.

pyiron/lammps/control.py
liamhuber: Indeed, I guess this would also get missed in the tests, since there we only check the output of `pressure_to_lammps`, and not the actual pressure that's written to the input field when we set a calc minimize!

tests/lammps/test_control.py
```diff
@@ -45,6 +46,14 @@ def test_mean(self):
...
             lc.measure_mean_value('something')
 
+    def test_pressure_to_lammps(self):
+        lc = LammpsControl()
+        # Correct normalization without rotation. Note that we convert from GPa to bar for LAMMPS.
+        no_rot = np.identity(3)
+        self.assertEqual(lc.pressure_to_lammps(1.0, no_rot), [10000.0, 10000.0, 10000.0, None, None, None])
+        self.assertEqual(lc.pressure_to_lammps([1.0, 2.0, 3.0], no_rot), [10000.0, 20000.0, 30000.0, None, None, None])
+        self.assertEqual(lc.pressure_to_lammps([1.0, 2.0, 3.0, None, None, None], no_rot), [10000.0, 20000.0, 30000.0, None, None, None])
+        self.assertEqual(lc.pressure_to_lammps([None, None, None, None, None, 2.0], no_rot), [None, None, None, None, None, 20000.0])
```
liamhuber, line 57: I apologize in advance if I'm missing something -- I looked over the diff carefully but I'm getting tired and might just be being dumb.

It looks to me like we allow using rotations in conjunction with fully defined pressure lists, but here there is no test for that, i.e. `pressure_to_lammps([a,b,c,d,e,f], some_rot)`. Don't we also need tests for this?

t-brink: There were no tests at all before, we should add a complete set before merging. But my brain capacity is diminishing today.

t-brink:
> And finally the hydrostatic pressure is not rotation invariant...

For a rotation matrix, it is R^-1 = R^T. It is also R^T R = R R^T = I. It is [[p,0,0],[0,p,0],[0,0,p]] = p * I. Therefore applying a rotation R: R (p*I) R^T = p * (R I R^T). It is R I = R, thus p * (R I R^T) = p * (R R^T) = p * I. Isn't it?

I'm going to sleep, I'll consider all comments again tomorrow. Thanks for in-depth review.

samwaseda:
> > And finally the hydrostatic pressure is not rotation invariant...
> 
> For a rotation matrix, it is R^-1 = R^T. It is also R^T R = R R^T = I. It is [[p,0,0],[0,p,0],[0,0,p]] = p * I. Therefore applying a rotation R: R (p*I) R^T = p * (R I R^T). It is R I = R, thus p * (R I R^T) = p * (R R^T) = p * I. Isn't it?
> 
> I'm going to sleep, I'll consider all comments again tomorrow. Thanks for in-depth review.

Yes, if the pressure is isotropic, but that's generally not the case...

jan-janssen:
As we revisit this part more often recently it might be reasonable to developer a clear concept and documentation, what is the behaviour we expect. @liamhuber already started this in https://github.com/pyiron/pyiron/pull/1017#issuecomment-702331163 

What I would like to add is the ability to define an `iso` constraint so all directions are coupled. From my perspective: 
```
job.calc_pressure(pressure=[0.0, 0.0, 0.0])
```
should equal: 
```
fix ensemble all box/relax x 0.0 y 0.0 z 0.0 couple none
```
while: 
```
job.calc_pressure(pressure=0.0)
```
should be: 
```
fix ensemble all box/relax iso 0.0
```
When I want to enforce a bcc structure at a concentration which favours fcc I can typically do so by using `relax iso` while `relax couple none` results in a phase transition. Therefore I would like to have a generic command to define `relax iso`.

samwaseda:
> As we revisit this part more often recently it might be reasonable to developer a clear concept and documentation, what is the behaviour we expect. @liamhuber already started this in [#1017 (comment)](https://github.com/pyiron/pyiron/pull/1017#issuecomment-702331163)
> 
> What I would like to add is the ability to define an `iso` constraint so all directions are coupled. From my perspective:
> 
> ```
> job.calc_pressure(pressure=[0.0, 0.0, 0.0])
> ```
> 
> should equal:
> 
> ```
> fix ensemble all box/relax x 0.0 y 0.0 z 0.0 couple none
> ```
> 
> while:
> 
> ```
> job.calc_pressure(pressure=0.0)
> ```
> 
> should be:
> 
> ```
> fix ensemble all box/relax iso 0.0
> ```
> 
> When I want to enforce a bcc structure at a concentration which favours fcc I can typically do so by using `relax iso` while `relax couple none` results in a phase transition. Therefore I would like to have a generic command to define `relax iso`.

Could you give me an example where `iso` makes sense? I have the feeling that in atomistic simulations it should always be anisotropic.

jan-janssen:
> Could you give me an example where `iso` makes sense? I have the feeling that in atomistic simulations it should always be anisotropic.

As I said I use `iso` to fix a particular crystal structure for example fix the bcc structure at high Ni concentration in: 
https://github.com/jan-janssen/NiCr-pyiron-example/blob/master/phasediagram_csv.ipynb
By fixing those structures we can calculate the free energy of these structures even when the physical structure would already be fcc.

samwaseda:
> > Could you give me an example where `iso` makes sense? I have the feeling that in atomistic simulations it should always be anisotropic.
> 
> As I said I use `iso` to fix a particular crystal structure for example fix the bcc structure at high Ni concentration in:
> https://github.com/jan-janssen/NiCr-pyiron-example/blob/master/phasediagram_csv.ipynb
> By fixing those structures we can calculate the free energy of these structures even when the physical structure would already be fcc.

aaaah ok now I understand it. But I still have the feeling that we mostly want anisotropic pressure relaxation, especially because if the system wants to go to fcc or bcc, it's usually up to the system to decide what to do... But fine I'd follow what the majority says.

jan-janssen:
> aaaah ok now I understand it. But I still have the feeling that we mostly want anisotropic pressure relaxation, especially because if the system wants to go to fcc or bcc, it's usually up to the system to decide what to do... But fine I'd follow what the majority says.

I just want the option to set this in the `calc_minimize` command it does not have to be `pressure=0` I am also fine with adding an `iso` flag, but as we now have this discussion I just wanted to bring this up.

liamhuber:
> I just want the option to set this in the `calc_minimize` command it does not have to be `pressure=0` I am also fine with adding an `iso` flag, but as we now have this discussion I just wanted to bring this up.

I am also flexible on the final syntax, but personally I find `pressure=p`=isotropic and `pressure=[p, p, p]`=non-isotropic.

samwaseda:
I also thought about it over night and I guess it's not a good idea to add `iso` because it makes a combination with `pressure` quite complicated. Maybe it's indeed a good idea to do it in the way as you both (@jan-janssen @liamhuber ) put it. Frankly when I use LAMMPS I use it in the hyper developer mode so I'd like to respect the opinions of normal users more.

pyiron/lammps/control.py
liamhuber: This is serious, but simple. It's fixed in [this PR](https://github.com/pyiron/pyiron/pull/1090), which should probably be merged before the one under discussion.

liamhuber: Ok, nevermind, it is not as simple as I hoped. Merge this first and we'll just live with the fact that non-zero pressure minimizations are using the wrong pressure. I don't think anyone is doing those anyway.

Hopefully once this PR is merged, the problems in that one will be easier to resolve.

t-brink:
Good morning!

I'll try to summarize different options and possible, reasonable behaviors, and we can discuss what is the best API, especially since I only know how I would use it. R is the rotation matrix, p or p1, p2, p3, ... are arbitrary pressures.

* `pressure=[p1,p2,p3,p4,p5,p6]`
     * in this case it is clear that the pressure tensor can be rotated if needed and everything is fine.
* `pressure=[p1,None,p2,p3,p4,None]` and R = I (or any other combination with `None`)
     * in this case there are also no problems, just apply the barostat in non-`None` directions
* `pressure=[p1,None,p2,p3,p4,None]` and R ≠ I (or any other combination with `None`)
     * this is ill defined and we should not allow it; maybe we think of something later, but let's just not allow it now
* `pressure=[p,p,p,None,None,None]` and R = I
     * this is fine
* `pressure=[p,p,p,None,None,None]` and R ≠ I
     * since the _isotropic_ (sorry for missing this word earlier), hydrostatic pressure tensor is translation invariant, I believe we can allow it; please think about it, too
* `pressure=p` and R = I
     * physically speaking, this is [p,p,p,0,0,0], i.e., it would be reasonable to expect `fix 1 all npt x <p> <p> 0.1 y <p> <p> 0.1 z <p> <p> 0.1 xy 0 0 0.1 xz 0 0 0.1 yz 0 0 0.1 couple none`, but this would usually not be what a user wants
     * as an MD user, I would likely expect this to be [p,p,p,None,None,None], i.e.,  `fix 1 all npt x <p> <p> 0.1 y <p> <p> 0.1 z <p> <p> 0.1 couple none`; other users might expect other things, though
     * currently there is no way to obtain `fix 1 all npt iso <p> <p> 0.1` (which is [p,p,p,None,None,None] but with coupling); I would be OK with special-casing this for a scalar vs. a list input, but probably a parameter `pressure_couple` (default: `False`) that _always_ couples x, y, and z would be my preferred option; this could be extended (if desired in the future) to support arbitrary coupling of directions, but this seems to be a very niche use case
* `pressure=p` and R ≠ I
     * as discussed above, I believe all the options for R = I should work, so the same possiblities apply
* as a user, I hate the idea that the behavior is different for p = 0 and p ≠ 0, and I hate the idea that `calc_minimize` is different from `calc_md` even more, and I hate the idea that this differs between R = I and R≠ I the most; it is extremely surprising and not intuitive at all, let's please try to avoid this?

That being said, if there is no clear outcome now, we could do the minimal fix proposed by @samwaseda (changing the type of a numpy array from `float64` to `object`) and revisit the definition of the `pressure` argument later (but soon).

liamhuber:
Concern:

Is it possible that we set a cell **B**, and a symmetric strain tensor, p=[pxx, pyy, pzz, pxy, pxz, pyz], but then when we rotate **B** so that is in the Lammps "upper triangular" coordinate frame with **R** that the rotated pressure becomes no longer symmetric?

I haven't thought of a specific set of vectors where this happens, I just have a general angst that it is mathematically possible. This would be an extremely troublesome edge case.

samwaseda:
> Good morning!
> 
> I'll try to summarize different options and possible, reasonable behaviors, and we can discuss what is the best API, especially since I only know how I would use it. R is the rotation matrix, p or p1, p2, p3, ... are arbitrary pressures.
> * `pressure=p` and R = I
>
> .....
> 
> That being said, if there is no clear outcome now, we could do the minimal fix proposed by @samwaseda (changing the type of a numpy array from `float64` to `object`) and revisit the definition of the `pressure` argument later (but soon).


Very nice!!! I mostly agree. Here are my comments:


> * `pressure=[p1,p2,p3,None,None,None]` and R ≠ I
>   

This case was not mentioned, but this shouldn't be allowed.

>   * pressure=p and R = I
>   * physically speaking, this is [p,p,p,0,0,0], i.e., it would be reasonable to expect `fix 1 all npt x <p> <p> 0.1 y <p> <p> 0.1 z <p> <p> 0.1 xy 0 0 0.1 xz 0 0 0.1 yz 0 0 0.1 couple none`, but this would usually not be what a user wants

There, I don't necessarily agree. I think the experimentalists will say the definition of pressure is p=(pxx+pyy+pzz)/3, so @jan-janssen 's claim on `iso` is in my opinion also valid. But this is something I guess we should agree on and which has to be clearly stated in the DocString, rather than discussing the metaphysics of the notation.

liamhuber:
> * `pressure=[p,p,p,None,None,None]` and R ≠ I
>   
>   * since the _isotropic_ (sorry for missing this word earlier), hydrostatic pressure tensor is translation invariant, I believe we can allow it; please think about it, too

With the special note that `[p,p,p,None,None,None] != [p1,p2,p3,None,None,None]`, I totally agree. I think this was established in earlier discussion, I just want to be 110% explicit.

>   * currently there is no way to obtain `fix 1 all npt iso <p> <p> 0.1` (which is [p,p,p,None,None,None] but with coupling); I would be OK with special-casing this for a scalar vs. a list input, but probably a parameter `pressure_couple` (default: `False`) that _always_ couples x, y, and z would be my preferred option; this could be extended (if desired in the future) to support arbitrary coupling of directions, but this seems to be a very niche use case

I don't have a physics objection to this, but from a software-developer perspective let's first think very hard if we can avoid introducing a new flag. As soon as we release a public-facing interface like that that we're on the hook to maintain it indefinitely -- it's functionality, it's documentation, everything. I see the need to allow the behaviour for niche cases, but I want to be very careful about assigning ourselves more work. If it's really an niche case a compromise might be something like `_pressure_couple` where we make the flag "private" to indicate that it might change/we might not support it forever.

> * as a user, I hate the idea that the behavior is different for p = 0 and p ≠ 0, and I hate the idea that `calc_minimize` is different from `calc_md` even more, and I hate the idea that this differs between R = I and R≠ I the most; it is extremely surprising and not intuitive at all, let's please try to avoid this?

I 100% share these frustrations. The only one I think we might not be able to reasonably avoid is the R = I and R≠ I difference, the others should be squashed ASAP. For the rotations, see my comment from ~10 minutes ago for concerns about why this might just be tricky.

samwaseda:
> Concern:
> 
> Is it possible that we set a cell **B**, and a symmetric strain tensor, p=[pxx, pyy, pzz, pxy, pxz, pyz], but then when we rotate **B** so that is in the Lammps "upper triangular" coordinate frame with **R** that the rotated pressure becomes no longer symmetric?
> 
> I haven't thought of a specific set of vectors where this happens, I just have a general angst that it is mathematically possible. This would be an extremely troublesome edge case.

No it's always symmetric, but I'm more concerned that the values are wrong. And I guess I just got a specific example where the values were wrong, but there I'm not even sure if I'm perfectly right or not. Anyway, this problem is not related to this PR so we should let @t-brink merge this one first and keep in mind that there could be a problem.

t-brink:
> Concern:
> 
> Is it possible that we set a cell **B**, and a symmetric strain tensor, p=[pxx, pyy, pzz, pxy, pxz, pyz], but then when we rotate **B** so that is in the Lammps "upper triangular" coordinate frame with **R** that the rotated pressure becomes no longer symmetric?
> 
> I haven't thought of a specific set of vectors where this happens, I just have a general angst that it is mathematically possible. This would be an extremely troublesome edge case.

If I am not severely mistaken, a symmetric tensor always remains symmetric under rotation. I also wrote a short python script to verify this by running a million random rotations on random, symmetric matrices, just in case.

The _strain_ tensor can become asymmetric in the LAMMPS case (decomposition into a strain part and a rotation part) but _not_ by pure rotation, but by change of shape. Meaning: when you apply a shear stress, the deformed sample should start deviating from the upper triangular convention. But it does not, instead there is also a rotation. This is a bit icky when the strains are not small, but this is inherent to LAMMPS. This has nothing to do with the input pressure tensor.

The _stress_ tensor cannot become asymmetric, since a change of reference frame (rotation) cannot change the physics.

samwaseda:
> The _stress_ tensor cannot become asymmetric, since a change of reference frame (rotation) cannot change the physics.

I guess a super important distinction is between "it cannot be" and "it is not allowed to be", because for certain frames, we know that certain things are **not allowed** to occur, but they **can** occur in pyiron. At least I'm too stupid to think of all possible cases, so I think I'll have to wait for them to sneak up on me to really understand what could go wrong.

> The strain tensor can become asymmetric in the LAMMPS case (decomposition into a strain part and a rotation part) but not by pure rotation, but by change of shape. Meaning: when you apply a shear stress, the deformed sample should start deviating from the upper triangular convention. But it does not, instead there is also a rotation. This is a bit icky when the strains are not small, but this is inherent to LAMMPS. This has nothing to do with the input pressure tensor.

An important difference between pyiron and LAMMPS is that LAMMPS only uses the Voigt notation and never the strain tensor, so that it can always be given in a symmetric strain tensor straightforwardly, and I'm not 100% sure if it has nothing to do with the input pressure, but I agree that it has nothing to do with this PR.

pyiron/lammps/control.py
```diff
@@ -201,42 +201,57 @@ def pressure_to_lammps(self, pressure, rotation_matrix):
...
             pressure (float/list/tuple/numpy.ndarray): The pressure(s) to convert.
             rotation_matrix (numpy.ndarray): The 3x3 matrix rotating from the pyiron to Lammps coordinate frame.
 
         Returns:
-            (numpy.ndarray): pxx, pyy, pzz, pxy, pxz, and pyz to be passed to Lammps.
+            (list): pxx, pyy, pzz, pxy, pxz, and pyz to be passed to Lammps.
+
+            or
+
+            (float): a single, isotropic pressure to be used with the "iso" option
```
t-brink, line 218: Is this the right way to document different return values?

pyiron/lammps/control.py
```diff
@@ -201,42 +201,57 @@ def pressure_to_lammps(self, pressure, rotation_matrix):
...
+        if np.isscalar(pressure):
+            return float(pressure) * LAMMPS_UNIT_CONVERSIONS[self["units"]]["pressure"]
+
+        # Normalize pressure to a list of 6 entries.
+        if len(pressure) > 6:
+            raise ValueError("Pressure can have a maximum of 6 values, (x, y, z, xy, xz, and yz), but got " +
+                             "{}".format(len(pressure)))
+        pressure = list(pressure) + (6 - len(pressure)) * [None]
+
+        if all(p is None for p in pressure):
```
t-brink, line 231: I changed the return type of the method to be a list in all cases, except for a single scalar. No more inconsistent returning of numpy arrays or lists. There is no need to use an array of type "object" instead, so don't.

pyiron/lammps/control.py
```diff
@@ -201,42 +201,57 @@ def pressure_to_lammps(self, pressure, rotation_matrix):
...
+            and np.isclose(pressure[0], pressure[1]) and np.isclose(pressure[1], pressure[2])
+            # Either all shear components must be None, or all shear components must be zero.
+            and (
+                (pressure[3] is None and pressure[4] is None and pressure[5] is None)
+                or
+                (pressure[3] is not None and np.isclose(pressure[3], 0.0) and
+                 pressure[4] is not None and np.isclose(pressure[4], 0.0) and
+                 pressure[5] is not None and np.isclose(pressure[5], 0.0))
+            )
+        )
```
t-brink, line 248: I do allow `pressure=[p,p,p,None,None,None]` and `pressure=[p,p,p,0,0,0]` for now with rotation ≠ I and skip the rotation. I'm not too attached to this, could be removed if nobody likes it.

t-brink:
Pushed an update using the @jan-janssen style `pressure=123.0` -> `iso` version as a proposal.

t-brink:
PS, this causes some other tests to fail, which expect `x 0 y 0 z 0` instead of `iso 0`. They would have to be updated if we agree on this behavior.

pyiron/lammps/control.py
liamhuber: Looks reasonable to me, but I would direct you to the google docstring style guide. I don't think I've seen the case of 'or' for different return types there before, but I'm ok with how you have it here.

pyiron/lammps/control.py
liamhuber: It's a pain-in-the-butt workaround on the backend, but I think it's nice to supply that behaviour on the frontend so I'm in favour.

liamhuber:
> > Concern:
> > Is it possible that we set a cell **B**, and a symmetric strain tensor, p=[pxx, pyy, pzz, pxy, pxz, pyz], but then when we rotate **B** so that is in the Lammps "upper triangular" coordinate frame with **R** that the rotated pressure becomes no longer symmetric?
> > I haven't thought of a specific set of vectors where this happens, I just have a general angst that it is mathematically possible. This would be an extremely troublesome edge case.
> 
> If I am not severely mistaken, a symmetric tensor always remains symmetric under rotation. I also wrote a short python script to verify this by running a million random rotations on random, symmetric matrices, just in case.

I am not prepared to say you're wrong, but I'm prepared to ask for more evidence. I took two minutes with both pen and paper and with google and didn't get a proof that this is the case, and my own numeric experiment shows that we might indeed be in trouble:

```python
import numpy as np
from scipy.spatial.transform import Rotation
r = np.random.rand(3,3)
P = 0.5 * (r + r.T)
assert(np.allclose(P, P.T))
R = Rotation.from_euler('xyz', np.random.rand(3)*np.pi).as_matrix()
assert(np.allclose(R.T@R, np.eye(3)))
assert(np.allclose(R@P, (R@P).T))

> ---------------------------------------------------------------------------
> AssertionError                            Traceback (most recent call last)
> <ipython-input-41-022318234aa5> in <module>
>       6 R = Rotation.from_euler('xyz', np.random.rand(3)*np.pi).as_matrix()
>       7 assert(np.allclose(R.T@R, np.eye(3)))
> ----> 8 assert(np.allclose(R@P, (R@P).T))
>
> AssertionError: 
```

pyiron/lammps/control.py
samwaseda: At least this part: `np.isclose(pressure[0], pressure[1]) and np.isclose(pressure[1], pressure[2])` could be `np.isclose(np.var(pressure[:3]), 0)`

t-brink:
> > > Concern:
> > > Is it possible that we set a cell **B**, and a symmetric strain tensor, p=[pxx, pyy, pzz, pxy, pxz, pyz], but then when we rotate **B** so that is in the Lammps "upper triangular" coordinate frame with **R** that the rotated pressure becomes no longer symmetric?
> > > I haven't thought of a specific set of vectors where this happens, I just have a general angst that it is mathematically possible. This would be an extremely troublesome edge case.
> > 
> > 
> > If I am not severely mistaken, a symmetric tensor always remains symmetric under rotation. I also wrote a short python script to verify this by running a million random rotations on random, symmetric matrices, just in case.
> 
> I am not prepared to say you're wrong, but I'm prepared to ask for more evidence. I took two minutes with both pen and paper and with google and didn't get a proof that this is the case, and my own numeric experiment shows that we might indeed be in trouble:
> 
> ```python
> import numpy as np
> from scipy.spatial.transform import Rotation
> r = np.random.rand(3,3)
> P = 0.5 * (r + r.T)
> assert(np.allclose(P, P.T))
> R = Rotation.from_euler('xyz', np.random.rand(3)*np.pi).as_matrix()
> assert(np.allclose(R.T@R, np.eye(3)))
> assert(np.allclose(R@P, (R@P).T))
> 
> > ---------------------------------------------------------------------------
> > AssertionError                            Traceback (most recent call last)
> > <ipython-input-41-022318234aa5> in <module>
> >       6 R = Rotation.from_euler('xyz', np.random.rand(3)*np.pi).as_matrix()
> >       7 assert(np.allclose(R.T@R, np.eye(3)))
> > ----> 8 assert(np.allclose(R@P, (R@P).T))
> >
> > AssertionError: 
> ```

A 2-tensor rotation of tensor T is R T R^-1!, not R T

pyiron/lammps/control.py
samwaseda: And this `pressure[3] is None and pressure[4] is None and pressure[5] is None` can be `np.all(pressure[3:]==None)` (I hope `pressure` is a numpy array)

liamhuber:
> A 2-tensor rotation of tensor T is R T R^-1!, not R T

Ah, perfect! Then I don't need my numeric test and can prove it on paper. Thanks!

pyiron/lammps/control.py
samwaseda: `pressure[0] is not None and pressure[1] is not None and pressure[2] is not None` -> `np.all(pressure[:3] != None)`

pyiron/lammps/control.py
```diff
@@ -201,42 +201,57 @@ def pressure_to_lammps(self, pressure, rotation_matrix):
...
-        # Cell degrees of freedom can be kept fixed using None for the pressure, check where that's done.
-        not_none_mask = (pressure!=None)
-        if not np.any(not_none_mask):
+        # If pressure is a scalar, only unit conversion is needed.
+        if np.isscalar(pressure):
+            return float(pressure) * LAMMPS_UNIT_CONVERSIONS[self["units"]]["pressure"]
+
+        # Normalize pressure to a list of 6 entries.
+        if len(pressure) > 6:
+            raise ValueError("Pressure can have a maximum of 6 values, (x, y, z, xy, xz, and yz), but got " +
```
jan-janssen, line 227: Is not this part a bit tricky? Let's say I want to transfer the external pressure from one simulation to the next, at the end of the simulation I get the pressure as 3x3 and now I manually have to transform it to 6x1? Can we included the 3x3 to 6x1 conversion here?

t-brink: That behavior is not new and it is the documented API (I just fiddled a bit with the implementation). As documented in the docstring, the API never allowed 6x6. Since this PR fixes a bug, could such discussion be moved to a separate issue, since this one is accumulating feature creep?

I actually agree in theory, but don't really care enough since the conversion to Voigt is trivial. I would prefer that someone who cares does the work, which includes tests, documentation, and likely a bunch of discussion. Meanwhile, my simulations don't work.

jan-janssen (approved):
As soon as the unit tests are fixed I am fine with merging this pull request.

samwaseda:
I see that most of the numpy arrays were now changed to list, but I might prefer numpy arrays.

t-brink:
> I see that most of the numpy arrays were now changed to list, but I might prefer numpy arrays.

Since the numpy arrays would have to be of type `object` (they need to contain `None`), I personally slightly prefer lists, which are simpler and have less overhead. Due to popular demand, though, I will change back to arrays in the next iteration together with fixing and expanding the tests.

jan-janssen:
Or you can use `np.nan` instead of `None`

t-brink:
> Or you can use `np.nan` instead of `None`

I do not like this, the meaning is different. NaN means some maths went wrong somewhere and `None` means there is no value here. Assigning a different meaning to NaN could hide and silence errors, i.e., the simulation would run but disable certain barostat directions silently, while in reality there was a division by zero or something. I would much rather program defensively, otherwise I would have already done this.

liamhuber (approved):

liamhuber:
Once the unittests are all passing, I'm happy.

samwaseda:
> Since the numpy arrays would have to be of type `object` (they need to contain `None`), I personally slightly prefer lists, which are simpler and have less overhead. Due to popular demand, though, I will change back to arrays in the next iteration together with fixing and expanding the tests.

Can you maybe give me the specific line where it causes a problem? I'm not sure if I understand the problem.

t-brink:
> > Since the numpy arrays would have to be of type `object` (they need to contain `None`), I personally slightly prefer lists, which are simpler and have less overhead. Due to popular demand, though, I will change back to arrays in the next iteration together with fixing and expanding the tests.
> 
> Can you maybe give me the specific line where it causes a problem? I'm not sure if I understand the problem.

`None` → NaN is the root cause of bug #1074 . There is this part in `calc_md()`:

https://github.com/pyiron/pyiron/blob/75b6b175cf714c4d3bc1c4089a30ba9dad234172/pyiron/lammps/control.py#L527-L532

and an equivalent one in `calc_minimize()`. They test if a component is `None`, but in a numpy array of dtype `float64` it would be NaN, since numpy autoconverts `None` to NaN.

NaN is a bad value to indicate "keep this cell direction fixed", because (1) it could be the result not of user input but of a math error _somewhere_ that is then promptly hidden and (2) because `np.nan == np.nan` is `False` (it has to be) which means you have to check using `np.isnan()`, which makes the code unclear and harder to understand (I personally directly understand the `value is not None` check but would start to wonder about the reason for `np.isnan()`, especially since in the user input, the special value is `None`).

Due to this whole thing, I personally prefer lists, where these hidden conversions do not take place. In the code with the lists all conversions are explicit and check for `None`. The same can be achieved with `dtype=object`, of course, that's a matter of taste.

samwaseda:
> `None` → NaN is the root cause of bug #1074 . There is this part in `calc_md()`:
> 
> https://github.com/pyiron/pyiron/blob/75b6b175cf714c4d3bc1c4089a30ba9dad234172/pyiron/lammps/control.py#L527-L532
> 
> and an equivalent one in `calc_minimize()`. They test if a component is `None`, but in a numpy array of dtype `float64` it would be NaN, since numpy autoconverts `None` to NaN.

Using `NaN` actively is not my option either, but the fact that it appeared in the first place was a bug and not what we wanted, as we probably all agree. But having `None` in a numpy array is a completely valid thing, isn't it?

There are essentially two things which I don't understand:

- Why were we converting the array to `float64` in the first place (which is something probably @jan-janssen or @liamhuber can explain - I presume one of them wrote it, in the hope that it wasn't me)?
- If we don't do the conversion, why shouldn't numpy arrays be used?

If there's no strong argument, I'd say whether to use numpy arrays or not goes beyond just a matter of taste, because:

- numpy is faster
- numpy requires less lines (which is quite crucial in pyiron)

samwaseda:
So, for example we currently have:

```
is_isotropic_hydrostatic = (
     pressure[0] is not None and pressure[1] is not None and pressure[2] is not None
     and np.isclose(pressure[0], pressure[1]) and np.isclose(pressure[1], pressure[2])
     and (
	 (pressure[3] is None and pressure[4] is None and pressure[5] is None)
	 or
	 (pressure[3] is not None and np.isclose(pressure[3], 0.0) and
	  pressure[4] is not None and np.isclose(pressure[4], 0.0) and
	  pressure[5] is not None and np.isclose(pressure[5], 0.0))
     )
 )
```

But this could be reduced to:

```
is_isotropic_hydrostatic = (
    np.all(pressure[:3]!=None) and np.isclose(np.var(pressure[:3]), 0)
    and (
        np.all(pressure[3:]==None)
        or np.all(np.isclose(pressure[3:][pressure[3:]!=None], 0))
    )
 )
```

And I'd say it's quite a difference.

pmrv:
> > `None` → NaN is the root cause of bug #1074 . There is this part in `calc_md()`:
> > https://github.com/pyiron/pyiron/blob/75b6b175cf714c4d3bc1c4089a30ba9dad234172/pyiron/lammps/control.py#L527-L532
> > 
> > and an equivalent one in `calc_minimize()`. They test if a component is `None`, but in a numpy array of dtype `float64` it would be NaN, since numpy autoconverts `None` to NaN.
> 
> Using `NaN` actively is not my option either, but the fact that it appeared in the first place was a bug and not what we wanted, as we probably all agree. But having `None` in a numpy array is a completely valid thing, isn't it?
As @t-brink pointed out this is only possible when `dtype=object`, negating any (potential) speed benefits of arrays vs. lists. But as we are talking about a list with _six_ elements, I fell speed concerns are a bit beside the point anyways.

If it is purely about lines of code I'll offer
```python
is_isotropic_hydrostatic = (
    all(p!=None for p in pressure[:3]) and all(np.isclose(p, pressure[0]) for p in pressure[1:3])
    and (
        all(t is not None for t in pressure[3:])
        or all(np.isclose(t, 0) for t in pressure[3:] if t is not None)
    )
 )

t-brink:
> So, for example we currently have:
> 
> ```
> is_isotropic_hydrostatic = (
>      pressure[0] is not None and pressure[1] is not None and pressure[2] is not None
>      and np.isclose(pressure[0], pressure[1]) and np.isclose(pressure[1], pressure[2])
>      and (
> 	 (pressure[3] is None and pressure[4] is None and pressure[5] is None)
> 	 or
> 	 (pressure[3] is not None and np.isclose(pressure[3], 0.0) and
> 	  pressure[4] is not None and np.isclose(pressure[4], 0.0) and
> 	  pressure[5] is not None and np.isclose(pressure[5], 0.0))
>      )
>  )
> ```
> 
> But this could be reduced to:
> 
> ```
> is_isotropic_hydrostatic = (
>     np.all(pressure[:3]!=None) and np.isclose(np.var(pressure[:3]), 0)
>     and (
>         np.all(pressure[3:]==None)
>         or np.all(np.isclose(pressure[3:][pressure[3:]!=None], 0))
>     )
>  )
> ```
> 
> And I'd say it's quite a difference.

Your version is shorter, but it takes me much longer to grasp. I had to think about the usage of `np.var()` and why it is equivalent.

**Your code is also wrong!** In `np.all(np.isclose(pressure[3:][pressure[3:]!=None], 0))` you only check that the components which are not `None` are zero. While in fact they **may not** be zero. Same goes for @pmrv 's version.

Your clever versions are too clever and introduce bugs and are harder to read. I believe this to be an antipattern. Less lines does not mean clearer (or faster). More lines that are easy to grasp are better.

Regarding speed: Numpy has a significant overhead that is only recovered if your arrays are large and your operations are vectorized. If you want to convince me otherwise, please do an actual benchmark on short lists/arrays and compare using numpy functions and arrays to native Python stuff. I expect that native Python types outperform numpy up to dozens or hundreds of entries.

I'll let this sit for a while, I don't want to rewrite the PR ten times. I'll probably propose the final version tomorrow or Monday, since changing the types also means changing the tests etc.

samwaseda:
> **Your code is also wrong!** In `np.all(np.isclose(pressure[3:][pressure[3:]!=None], 0))` you only check that the components which are not `None` are zero. While in fact they **may not** be `None`. Same goes for @pmrv 's version.

????? That's right, because the components which are `None` are `None`. And [your initial issue](https://github.com/pyiron/pyiron/issues/1074) was a perfect example for all shear components being `None`...

> 
> Your clever versions are too clever and introduce bugs and are harder to read. I believe this to be an antipattern. Less lines does not mean clearer (or faster). More lines that are easy to grasp are better.
> 

It's true that numpy sometimes makes it more difficult to understand, but in this specific case I don't agree, because in general it's extremely difficult to understand an `if-` statements which has a very large number of `and` and `or`.

> Regarding speed: Numpy has a significant overhead that is only recovered if your arrays are large and your operations are vectorized. If you want to convince me otherwise, please do an actual benchmark on short lists/arrays and compare using numpy functions and arrays to native Python stuff. I expect that native Python types outperform numpy up to dozens or hundreds of entries.

Ok that's possible. I'm not an expert on this so you're probably right.

> I'll let this sit for a while, I don't want to rewrite the PR ten times. I'll probably propose the final version tomorrow or Monday, since changing the types also means changing the tests etc.

I can also take it over if you don't want to continue. I'm used to changing codes many times and I'm famously happy to do that.

By the way one question: `is_isotropic_hydrostatic` is apparently currently `True` when some of the shear components are `None` and some of them are 0, but shouldn't they be either all `None` or all 0?

t-brink:
> > **Your code is also wrong!** In `np.all(np.isclose(pressure[3:][pressure[3:]!=None], 0))` you only check that the components which are not `None` are zero. While in fact they **may not** be `None`. Same goes for @pmrv 's version.
> 
> ????? That's right, because the components which are `None` are `None`. And [your initial issue](https://github.com/pyiron/pyiron/issues/1074) was a perfect example for all shear components being `None`...

My comment only refers to this one line in the `if` statement, not anything else. What I am saying is that that line should be `np.all(pressure[3:] != None) and np.all(np.isclose(pressure[3:], 0))` instead of `np.all(np.isclose(pressure[3:][pressure[3:]!=None], 0))`. See also comment at the end.

> > I'll let this sit for a while, I don't want to rewrite the PR ten times. I'll probably propose the final version tomorrow or Monday, since changing the types also means changing the tests etc.
> 
> I can also take it over if you don't want to continue. I'm used to changing codes many times and I'm famously happy to do that.

It's just that I am not sure what today's discussion is even about, really, since I already said that I will change back to numpy arrays in the next iteration and I thought that questions like the type of the arrays were already discussed. _To make this whole thing more actionable:_ I will push a new version with numpy arrays and full tests and stop discussing here. If you like it, merge it, otherwise please feel free to take over and change whatever you like. I feel this is much more productive for everybody.

> 
> By the way one question: `is_isotropic_hydrostatic` is apparently currently `True` when some of the shear components are `None` and some of them are 0, but shouldn't they be either all `None` or all 0?

For the tensor to be isotropic and hydrostatic and to be usable with any rotation, either all shear components must be `None` or all must be zero, yes. That is exactly what my code does and yours doesn't. That is the reason I said yours is wrong. By the way, that is also what the comment in the code says, which I put there to make it clearer.

samwaseda:
> > By the way one question: `is_isotropic_hydrostatic` is apparently currently `True` when some of the shear components are `None` and some of them are 0, but shouldn't they be either all `None` or all 0?
> 
> For the tensor to be isotropic and hydrostatic and to be usable with any rotation, either all shear components must be `None` or all must be zero, yes. That is exactly what my code does and yours doesn't. That is the reason I said yours is wrong. By the way, that is also what the comment in the code says, which I put there to make it clearer.

XD then it was a perfect example for the case where I didn't understand the list representation

samwaseda:
And sorry my final comment: since @pmrv gave his support for lists, and you gave me a good argument for not using numpy, I'm also ready to give in. Don't feel obliged to follow all my comments.

t-brink:
OK, my final proposal (under the assumption that all tests pass), someone please merge or take over. Notes:

* I stayed with lists for the sole reason that `np.isclose()` does not work with `dtype="object"`. All my attempts to work around this were uglier than just using lists. Feel free to improve if you really want, the test coverage should be quite good now.
* I also fixed the unit conversion in `calc_minimize()` since the tests would otherwise fail and it felt stupid not to do it. Thanks to @liamhuber for catching it.
* I didn't really grasp the tests in `tests/lammps/test_interactive.py`, so I did not change anything (they pass). They could really use some more test cases, especially for `calc_md()`, though.

Thanks for the reviews.

jan-janssen:
Thank you @t-brink

PR #1086, opened 2020-10-21
Implement unit test for selective dynamics

jan-janssen:
This test breaks currently

### DIFF
--- a/tests/thermodynamics/test_hessianJob.py
+++ b/tests/thermodynamics/test_hessianJob.py
@@ -49,5 +49,34 @@ class TestHessianJob(unittest.TestCase):
         self.assertAlmostEqual(self.job.output.pressures[0, 0, 0], -0.03)
 
 
+class TestSelectiveDynamics(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.file_location = os.path.dirname(os.path.abspath(__file__))
+        cls.project = Project(
+            os.path.join(cls.file_location, "hessian_class")
+        )
+        cls.project.remove_jobs_silently(recursive=True)
+        
+    def test_selective_dynamics(self):
+        structure = self.project.create_ase_bulk("Al", cubic=True)
+        structure.add_tag(selective_dynamics=None)
+        structure.selective_dynamics[0] = [True, True, True]
+        structure.selective_dynamics[1] = [True, True, True]
+        structure.selective_dynamics[2] = [False, False, False]
+        structure.selective_dynamics[3] = [False, False, False]
+        job = self.project.create_job(self.project.job_type.Lammps, "lmp")
+        job.structure = structure
+        job.run()
+        job_reload = self.project.load(job.job_name)
+        structure_reload = job_reload.get_structure()
+        structure_reload.add_tag(selective_dynamics=None)
+        structure_reload.selective_dynamics[0] = [True, True, True]
+        structure_reload.selective_dynamics[1] = [True, True, True]
+        structure_reload.selective_dynamics[2] = [False, False, False]
+        structure_reload.selective_dynamics[3] = [False, False, False]
+        self.assertTrue(True)
+
+
 if __name__ == "__main__":
     unittest.main()
--- a/tests/thermodynamics/test_hessianJob.py
+++ b/tests/thermodynamics/test_hessianJob.py
@@ -49,27 +49,5 @@ class TestHessianJob(unittest.TestCase):
         self.assertAlmostEqual(self.job.output.pressures[0, 0, 0], -0.03)
 
 
-class TestSelectiveDynamics(unittest.TestCase):
-    @classmethod
-    def setUpClass(cls):
-        cls.file_location = os.path.dirname(os.path.abspath(__file__))
-        cls.project = Project(
-            os.path.join(cls.file_location, "selective_dynamics")
-        )
-        cls.project.remove_jobs_silently(recursive=True)
-
-    def test_selective_dynamics(self):
-        structure = Atoms(
-            positions=[[0, 0, 0], [1, 1, 1]], elements=["Fe", "Fe"], cell=2 * np.eye(3)
-        )
-        structure.add_tag(selective_dynamics=None)
-        structure.selective_dynamics[0] = [True, True, True]
-        structure.selective_dynamics[1] = [False, False, False]
-        job = self.project.create_job("HessianJob", "hess")
-        job.structure = structure
-        job.to_hdf()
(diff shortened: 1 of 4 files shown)


jan-janssen:
```
ERROR: test_selective_dynamics (thermodynamics.test_hessianJob.TestSelectiveDynamics)
  File "/home/vsts/work/1/s/tests/thermodynamics/test_hessianJob.py", line 76, in test_selective_dynamics
    structure_reload.selective_dynamics[1] = [False, False, False]
  File "/home/vsts/work/1/s/pyiron/atomistics/structure/sparse_list.py", line 203, in __setitem__
    raise IndexError
IndexError

```

jan-janssen:
The bug is fixed with https://github.com/pyiron/pyiron/pull/1086/commits/bbb135351dc52aad8dcaed8b31e025f5f2e5f985

tests/thermodynamics/test_hessianJob.py
```diff
@@ -49,5 +49,33 @@ def test_forces(self):
...
+        job = self.project.create_job("HessianJob", "hess")
+        job.structure = structure
+        job.to_hdf()
+        job_reload = self.project.create_job("HessianJob", "hess")
+        job_reload.from_hdf()
+        structure_reload = job_reload.structure
+        structure_reload.add_tag(selective_dynamics=None)
+        structure_reload.selective_dynamics[0] = [True, True, True]
+        structure_reload.selective_dynamics[1] = [False, False, False]
+        self.assertTrue(True)
```
sudarsan-surendralal, line 77: This doesn't do anything right? Maybe we can check the change in positions to see if the atoms have moved or not.

tests/thermodynamics/test_hessianJob.py
```diff
@@ -49,5 +49,33 @@ def test_forces(self):
...
+        structure.add_tag(selective_dynamics=None)
+        structure.selective_dynamics[0] = [True, True, True]
+        structure.selective_dynamics[1] = [False, False, False]
+        job = self.project.create_job("HessianJob", "hess")
+        job.structure = structure
+        job.to_hdf()
+        job_reload = self.project.create_job("HessianJob", "hess")
+        job_reload.from_hdf()
+        structure_reload = job_reload.structure
+        structure_reload.add_tag(selective_dynamics=None)
```
sudarsan-surendralal, line 74: I thought the default value should be something like `[True, True, True]` or `[False, False, False]`?

tests/thermodynamics/test_hessianJob.py
jan-janssen: The issue is that without the fix reloading the selective dynamics tag from the HDF5 file gives the following error: 
```
ERROR: test_selective_dynamics (thermodynamics.test_hessianJob.TestSelectiveDynamics)
  File "/home/vsts/work/1/s/tests/thermodynamics/test_hessianJob.py", line 76, in test_selective_dynamics
    structure_reload.selective_dynamics[1] = [False, False, False]
  File "/home/vsts/work/1/s/pyiron/atomistics/structure/sparse_list.py", line 203, in __setitem__
    raise IndexError
IndexError
```
Therefore this test basically just confirms that the reloading works and the selective dynamics can be overwritten afterwards.

tests/thermodynamics/test_hessianJob.py
jan-janssen: Typically I use `None` to apply multiple filters and afterwards check which atoms have not been specified yet.

tests/thermodynamics/test_hessianJob.py
pmrv: Maybe it's clearer to do the `from_hdf` in an `try/except IndexError` and call `self.fail` in the except clause to indicate test failure.

jan-janssen: The bug is now fixed - I do not want it to fail - I just say it was failing before, as the length of the tag array was not restored correctly.

sudarsan-surendralal: > The issue is that without the fix reloading the selective dynamics tag from the HDF5 file gives the following error:
> 
> ```
> ERROR: test_selective_dynamics (thermodynamics.test_hessianJob.TestSelectiveDynamics)
>   File "/home/vsts/work/1/s/tests/thermodynamics/test_hessianJob.py", line 76, in test_selective_dynamics
>     structure_reload.selective_dynamics[1] = [False, False, False]
>   File "/home/vsts/work/1/s/pyiron/atomistics/structure/sparse_list.py", line 203, in __setitem__
>     raise IndexError
> IndexError
> ```
> 
> Therefore this test basically just confirms that the reloading works and the selective dynamics can be overwritten afterwards.

Then you should do something like `job["output/structure"].to_object()` and print it's selective dynamics values to check.

tests/thermodynamics/test_hessianJob.py
```diff
@@ -49,5 +49,27 @@ def test_forces(self):
...
+        structure = Atoms(
+            positions=[[0, 0, 0], [1, 1, 1]], elements=["Fe", "Fe"], cell=2 * np.eye(3)
+        )
+        structure.add_tag(selective_dynamics=None)
+        structure.selective_dynamics[0] = [True, True, True]
+        structure.selective_dynamics[1] = [False, False, False]
+        job = self.project.create_job("HessianJob", "hess")
+        job.structure = structure
+        job.to_hdf()
+        self.assertEqual(job["output/structure"].to_object().selective_dynamics.tolist(), [[True, True, True], [False, False, False]])
```
sudarsan-surendralal, line 71: Shouldn't it be `input/structure`? Also there are trailing whitespaces

jan-janssen:
I updated the tests, without the fix they fail with: 
```
ERROR: test_from_hdf (atomistics.structure.test_atoms.TestAtoms)
  File "/home/vsts/work/1/s/tests/atomistics/structure/test_atoms.py", line 237, in test_from_hdf
    basis.selective_dynamics[6] = [True, True, True]
  File "/home/vsts/work/1/s/pyiron/atomistics/structure/sparse_list.py", line 203, in __setitem__
    raise IndexError
IndexError

```

tests/thermodynamics/test_hessianJob.py
pmrv: I understand that the test don't fail after the other PR, but my reasoning is that failing explicitly in case of `IndexError` makes it easier to identify a regression in the  future.  I'm also happy with the more explicit asserts from @sudarsan-surendralal, though.

pmrv (approved):

sudarsan-surendralal (approved):

PR #1090, opened 2020-10-22
No double units

liamhuber:
We were converting to Lammps pressure units twice in `calc_minimize`! No more.

Not in the scope of this PR: testing anything except that the pressure units are being converted to the control field correctly.

### DIFF
--- a/tests/lammps/test_control.py
+++ b/tests/lammps/test_control.py
@@ -3,10 +3,33 @@
 # Distributed under the terms of "New BSD License", see the LICENSE file.
 
 import unittest
-from pyiron.lammps.control import LammpsControl
+from pyiron.lammps.control import LammpsControl, LAMMPS_UNIT_CONVERSIONS
+import numpy as np
 
 
 class TestLammps(unittest.TestCase):
+    def test_calc_minimize(self):
+        lc = LammpsControl()
+        p = 1
+        lc.calc_minimize(pressure=p, rotation_matrix=np.eye(3))
+        self.assertTrue(
+            np.isclose(
+                float(lc['fix___ensemble'].split(' x ')[1].split(' ')[0]),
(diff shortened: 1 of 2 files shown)


liamhuber:
To hell with this. I am going to wait until [PR 1083](https://github.com/pyiron/pyiron/pull/1083) is merged before attempting to sort this out. I don't think anyone is doing constant (non-zero) pressure minimizations right now anyway.

tests/lammps/test_control.py
```diff
@@ -3,10 +3,33 @@
...
+        lc = LammpsControl()
+        p = 1
+        lc.calc_md(temperature=300, pressure=p, rotation_matrix=np.eye(3))
+        self.assertTrue(
+            np.isclose(
+                float(lc['fix___ensemble'].split(' x ')[1].split(' ')[0]),
+                p * LAMMPS_UNIT_CONVERSIONS[lc["units"]]["pressure"]
+            )
+        )
+
```
t-brink, line 32: The tests assume a specific way of implementing `pressure=p` and parsing the LAMMPS input file. Probably that's the best one can do without running the simulation, but could you keep an eye on this in case we change this?

I would also like to request tests for different pressure inputs, the code paths are different. Otherwise thanks for catching it.

t-brink (changes requested):

tests/lammps/test_control.py
liamhuber: Sure thing. I actually found in the Lammps base tests that there's a bit more extensive testing of the input. I think there should be a way to use that to make these tests more in-depth and also avoid duplicating code. I'll wait until your PR is in though, and see what the syntax winds up being for the pressure input :)

liamhuber:
Resolved as part of [1074](https://github.com/pyiron/pyiron/issues/1074)

PR #1092, opened 2020-10-23
ScipyMinimize: Set status to finished at the end of run_static()

jan-janssen:


samwaseda:
Since it also parses convergence check, we can also think about distinguishing between `finished` and `not_converged`.

samwaseda (approved):
Yes, this looks great to me. Actually we should generally make use of `not_converged` more, because it's a crucial piece of information in many DFT calculations (which are often underrated in pyiron since many developers use LAMMPS more...)

PR #1095, opened 2020-10-23
SciPyMinimizer: Keep default values from self.input

jan-janssen:


samwaseda (commented):
Here, I have to involve @liamhuber as well, because we had a similar discussion before.

So the usage of `calc_minimize` is anyway optional in the case of `ScipyMinimizer` (it anyway doesn't do anything else). In this sense, we can consider `calc_minimize` as a setter function. The question is, do we provide default values in setter functions? If we do, we might have the danger that someone encounters the same problem as you did. If not, then we can ask why do we need a setter function in the first place? I'm fine with both, but currently in many places the two cases coexist, and I just want you to think about it before we merge this one and leave this question unanswered.

PR #1098, opened 2020-10-26
Fix plot3d

jan-janssen:
The `copy()` part was missing.


samwaseda (approved):
I didn't anticipate this one. Good job

liamhuber (approved):
I find the fix fast and elegant.

However, I'm suuuper sketched out that this is how we need to do it. This has some smell further upstream. Later we really need to come back to how we're handling the structure of `Atoms`... (I'm still in favour, atm, of a multiple inheritence setup, with which this issue would not have come up.) But this PR is not a good place for an extended discussion on that.

PR #1099, opened 2020-10-26
Simplify logic

liamhuber:
There was some debate over how to handle these lines over in [PR 1083](https://github.com/pyiron/pyiron/pull/1083), and I didn't want to wade in with a third opinion ;) But now that that's safely merged, here's my take on the logic for checking if a pressure is isotropic and hydrostatic.

### DIFF
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -235,20 +235,7 @@ class LammpsControl(GenericParameters):
 
         # If necessary, rotate the pressure tensor to the Lammps coordinate frame.
         # Isotropic, hydrostatic pressures are rotation invariant.
-        is_isotropic_hydrostatic = (
-            # All diagonal components must be equal and not None.
-            pressure[0] is not None and pressure[1] is not None and pressure[2] is not None
-            and np.isclose(pressure[0], pressure[1]) and np.isclose(pressure[1], pressure[2])
-            # Either all shear components must be None, or all shear components must be zero.
-            and (
-                (pressure[3] is None and pressure[4] is None and pressure[5] is None)
-                or
-                (pressure[3] is not None and np.isclose(pressure[3], 0.0) and
-                 pressure[4] is not None and np.isclose(pressure[4], 0.0) and
-                 pressure[5] is not None and np.isclose(pressure[5], 0.0))
-            )
-        )
-        if not np.isclose(np.matrix.trace(rotation_matrix), 3) and not is_isotropic_hydrostatic:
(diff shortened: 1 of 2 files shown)


pyiron/lammps/control.py
```diff
@@ -270,6 +257,16 @@ def pressure_to_lammps(self, pressure, rotation_matrix):
                  else p)
                 for p in pressure]
 
+    @staticmethod
+    def _is_isotropic_hydrostatic(pressure):
+        return (
+            None not in pressure[:3] and np.allclose(pressure[:3], pressure[0])
+            and (
+                all(p is None for p in pressure[3:])
+                or (None not in pressure[3:] and np.allclose(pressure[3:], 0.))
```
samwaseda, line 266: I find it still difficult to understand. I'd like to propose two possibilities:

1. We simply say it's an input and whoever puts 1.0e-16 instead of 0 must know what they're doing -> Instead of checking whether the values are close to 0, we check whether it's a clear 0 or not
2. We include one more step before `return` to round up `pressure`.

What do you think?

samwaseda: In both cases we'd have `all(p==0 for p in pressure[3:])` instead of `None not in pressure[3:] and np.allclose(pressure[3:], 0.)`

liamhuber: @samwaseda ok, I have no objection to breaking it down further to improve readability, which is what I interpret (2) to be.

I'm hard rejecting (1). The only way comparing to a clear 0 is fair is if we cast what we're comparing as an int, and that sounds super sketchy. Plus it doesn't help us for the isotropic evaluation where we need to make sure the first three components compare equally to each other. Anyhow, when I get a few minutes I'll do (2) and hopefully remove the desire for (1) ;)

samwaseda: And/or we can maybe separate the two parts:

‘’’
if (None in pressure[:3] or not np.allclose(pressure[:3], pressure[0])):
    return False
if ...
‘’’

Again this is just a suggestion. I don’t strongly stick to it.

samwaseda: (Sorry I’m writing it on my phone I know it looks horrible)

liamhuber:
@samwaseda I'm still using `is` for `None` comparisons (because PEP8, which I'm married to) and `np.allclose` (which I feel is the right tool for the job, but am not married to). Let me know if you find it more readable.

There's a small amount of duplicate code, but at this point it still doesn't feel like it's worth introducing a whole new function.

samwaseda (approved):
I like this one. It’s very clear now.

t-brink (approved):
Seems good. Approved under the condition that the pending tests pass.

PR #1100, opened 2020-10-26
change pressure tolerance

samwaseda:
I included a pressure tolerance attribute. Default set to 1 MPa. I guess it's low enough.


jan-janssen:
I like the changes, we just have to clarify whether we want to overwrite the values in the input list or not: 
https://github.com/pyiron/pyiron/pull/1095

samwaseda:
> I like the changes, we just have to clarify whether we want to overwrite the values in the input list or not:
> #1095

For the sake of consistency I followed the same principle but if we end up deciding in #1095 not to overwrite I will follow the decision as well. It's up to the state of agreement which one should be merged first, I'd say.

PR #1104, opened 2020-10-27
Fix #1093

### LINKED ISSUE #1093: Plotting repeats stops working
If I look at the repeated structure manually, repeat seems to be working. But plot bombs and just shows the unit cell.

Joerg had the idea that maybe the cell was not getting done right and the atoms are being wrapped back into the unit cell, so I displaced the 0th atom (then you should see two atoms near each other), but this has no effect.

I'm on the current heads of pyiron and pyiron_base.

MWE:
```pyiron
struct = pr.create_ase_bulk('Al', cubic=True).repeat(2)
struct.positions[0, 0] += 0.6
struct.plot3d()
```

pmrv:
The underlying problem is that `Atoms.__copy__` copies the `Visualize`
class, but does not set a the new `Atoms` object as its reference.
However since `Visualize` does not keep any state expect that reference,
we can side-step this problem by making `Atoms.visualize` a property
that returns a new `Visualize` instance on the fly.  To avoid
unnecessary instantiations `Visualize` is now a singleton.  I also went
ahead and renamed the member to `_visualize` since it is not documented
and we labeled the whole module where it comes from as internal.

This effectively reverts PR #1098.  Since this avoid special casing `Atoms.visualize` in the copy method, I thought this was a bit neater.

### DIFF
--- a/pyiron/atomistics/structure/_visualize.py
+++ b/pyiron/atomistics/structure/_visualize.py
@@ -6,6 +6,7 @@ import numpy as np
 import warnings
 from matplotlib.colors import rgb2hex
 from pyiron_base import Settings
+from pyiron_base.generic.util import Singleton
 from scipy.interpolate import interp1d
 
 __author__ = "Joerg Neugebauer, Sudarsan Surendralal"
@@ -21,7 +22,7 @@ __date__ = "Sep 1, 2017"
 
 s = Settings()
 
-class Visualize:
+class Visualize(Singleton):
     def __init__(self, atoms):
         self._ref_atoms = atoms
 
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -192,7 +192,6 @@ class Atoms(ASEAtoms):
             self.dimension = len(self.positions[0])
         else:
             self.dimension = 0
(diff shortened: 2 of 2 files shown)


jan-janssen:
While there are parts where the singleton is necessary I am not sure if it is the best programming pattern, in this case I prefer https://github.com/pyiron/pyiron/pull/1098

pmrv:
I'm not married to using `Singleton`.  In this case initializing `Visualize` is just the reference to the atoms class and should be fairly quick to do every time we do a 3d plot (as in: definitely quicker than doing the actual 3d plot).

jan-janssen:
But that would be the same as having just functions rather than a class, but if I understand Liam correctly he prefers to have a class https://github.com/pyiron/pyiron/pull/1073#pullrequestreview-512398209 that is why I fixed it in the copy method in https://github.com/pyiron/pyiron/pull/1098 so the bug was already solved. So I am not exactly sure what is the purpose of this pull request.

samwaseda:
I want a more descriptive title...

pmrv:
I just wanted to suggest a different fix, but have no hard opinions against #1098.  Since it seems the Singleton breaks the tests, I'll just close it here.

PR #1107, opened 2020-10-27
Condition if no unoccupied states are there

sudarsan-surendralal:
Fixing parsing on OUTCAR files (when no unoccupied states are present)


jan-janssen:
Does it make sense to add Lifangs OUTCAR file or a reduced version of that OUTCAR file to the tests?

sudarsan-surendralal:
> 
> 
> Does it make sense to add Lifangs OUTCAR file or a reduced version of that OUTCAR file to the tests?

It's a little large (1.2 MB). Plus the content of the OUTCAR isn't that unique/different. It's just that the function should have considered this special case.

jan-janssen:
Alternatively can we add this special case to one of the existing OUTCAR files?

lfzhu-phys:
@jan-janssen @sudarsan-surendralal Thanks a lot for the analysis:)

sudarsan-surendralal:
> 
> 
> Alternatively can we add this special case to one of the existing OUTCAR files?

I've done this now.

jan-janssen (approved):
Looks good to me

sudarsan-surendralal:
@lfzhu-phys, you should be able to work now with the master branch.

lfzhu-phys:
@sudarsan-surendralal Great, thanks a lot, I will try it out:)

Issue #1108, opened 2020-10-28
ase version 3.20.1 incompatible with loading a job 


cmmmalik:
While doing ` pr.load_from_jobpath(jobid)` runs into an error because ase recent update (version 3.20.1) does not support 'pbc'  as an argument to ase.cell.Cell
```
>>> pr.load_from_jobpath(job_id=1)
{'array': array([[0.   , 2.025, 2.025],
       [2.025, 0.   , 2.025],
       [2.025, 2.025, 0.   ]]), 'pbc': array([ True,  True,  True])}
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/am77cuxa/git_repository/pyiron/0.0.0/lib/python3.7/site-packages/pyiron-0.0.0-py3.7.egg/pyiron/project.py", line 233, in load_from_jobpath
    job_id=job_id, db_entry=db_entry, convert_to_object=convert_to_object
  File "/home/am77cuxa/git_repository/pyiron_base/0.0.0/lib/python3.7/site-packages/pyiron_base-0.1.23.post0.dev2-py3.7.egg/pyiron_base/project/generic.py", line 762, in load_from_jobpath
    convert_to_object=convert_to_object, project=job.project_hdf5.copy()
  File "/home/am77cuxa/git_repository/pyiron_base/0.0.0/lib/python3.7/site-packages/pyiron_base-0.1.23.post0.dev2-py3.7.egg/pyiron_base/job/core.py", line 541, in load_object
    return self.to_object(project=job_dir, job_name=self._name)
  File "/home/am77cuxa/git_repository/pyiron_base/0.0.0/lib/python3.7/site-packages/pyiron_base-0.1.23.post0.dev2-py3.7.egg/pyiron_base/job/core.py", line 480, in to_object
    return self.project_hdf5.to_object(object_type, **qwargs)
  File "/home/am77cuxa/git_repository/pyiron_base/0.0.0/lib/python3.7/site-packages/pyiron_base-0.1.23.post0.dev2-py3.7.egg/pyiron_base/generic/hdfio.py", line 1319, in to_object
    group_name = self.h5_path.split('/')[-1]
  File "/home/am77cuxa/git_repository/pyiron/0.0.0/lib/python3.7/site-packages/pyiron-0.0.0-py3.7.egg/pyiron/gpaw/gpaw.py", line 193, in from_hdf
    super(Gpaw, self).from_hdf(hdf=hdf, group_name=group_name)
[...]

PR #1110, opened 2020-10-28
Increase ase requirement

jan-janssen:


jan-janssen:
I assume this breaks as discussed in https://github.com/pyiron/pyiron/issues/1108

PR #1120, opened 2020-10-29
Get neighbors by distance

niklassiemer:
In PR https://github.com/pyiron/pyiron/pull/1015 the function `get_neighbors_by_distance` is introduced. However, this function after all behaves different from what I would expect:
It requires the number of neighbors (default is 100) which are considered and, moreover, code like
`get_neighbors_by_distance(10)` would result in 10 neighbors considered, not in a radius of 10.
With this PR I propose a new function `get_number_of_neighbors_in_sphere` which is used in `get_neighbors_by_distance`, such that all neighbors in a given distance are actually found.


pyiron/atomistics/structure/atoms.py
```diff
@@ -1118,24 +1118,75 @@ def get_extended_positions(self, width):
...
+                neigh = self._get_neighbors(
+                    num_neighbors=num_neighbors,
+                    t_vec=False,
+                    tolerance=tolerance,
+                    id_list=id_list,
+                    cutoff_radius=cutoff_radius,
+                    boundary_width_factor=boundary_width_factor,
+                )
+                neigh.indices = [indices[dist < np.inf] for indices, dist in zip(neigh.indices, neigh.distances)]
+                neigh.distances = [dist[dist < np.inf] for dist in neigh.distances]
```
samwaseda, line 1164: Is it really necessary to update `neigh.distances`?

pyiron/atomistics/structure/atoms.py
```diff
@@ -1118,24 +1118,75 @@ def get_extended_positions(self, width):
...
+            while num_neighbors_old < num_neighbors:
+                neigh = self._get_neighbors(
+                    num_neighbors=num_neighbors,
+                    t_vec=False,
+                    tolerance=tolerance,
+                    id_list=id_list,
+                    cutoff_radius=cutoff_radius,
+                    boundary_width_factor=boundary_width_factor,
+                )
+                neigh.indices = [indices[dist < np.inf] for indices, dist in zip(neigh.indices, neigh.distances)]
```
samwaseda, line 1163: Do I see it correctly that what you want is equivalent to:

```
num_neighbors = np.sum(neigh.distances<np.inf, axis=-1).max()
```

?

pyiron/atomistics/structure/atoms.py
```diff
@@ -1118,24 +1118,75 @@ def get_extended_positions(self, width):
...
+                num_neighbors=num_neighbors,
+                t_vec=False,
+                tolerance=tolerance,
+                id_list=id_list,
+                cutoff_radius=cutoff_radius,
+                boundary_width_factor=boundary_width_factor,
+            )
+            num_neighbors = max(len(elem) for elem in neigh.indices)
+        else:
+            num_neighbors = int(0.5*cutoff_radius**3)
```
samwaseda, line 1152: I guess it makes more sense to make it dependent on the density, rather than just the length. How about something like `num_neighbors = int(1.1*cutoff_radius**3/self.get_volume(per_atom=True))`?

samwaseda (commented):
Since I personally never use `get_neighbors_by_distance`, I'm fine with any implementation. I just want to say that if the goal is to get a good number of neighbors, this algorithm is probably a bit too expensive, because it internally runs `_get_neighbors` possibly multiple times (which itself is a relatively expensive function). Besides, there's still no guarantee that you get the correct number of neighbors in the end, since `_get_neighbors` itself relies on `_get_boundary_layer_width`, which estimates the width required from the density. In total, what in my opinion makes the most sense is to estimate the number of atoms from the global density (or Voronoi, if you really want to be sure), i.e. it should be:

```
num_neighbors = (1+buffer)*length**3/volume_per_atom
```

And maybe it's up to the user to determine `buffer`.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1118,24 +1118,75 @@ def get_extended_positions(self, width):
...
         id_list=None,
-        cutoff_radius=np.inf,
         boundary_width_factor=1.2,
     ):
         """
 
         Args:
-            num_neighbors (int): number of neighbors
+            cutoff_radius (float): Upper bound of the distance to which the search must be done
+            num_neighbors (int): maximum number of neighbors found
```
samwaseda, line 1185: Maybe `int/None`?

pyiron/atomistics/structure/atoms.py
samwaseda: Sorry since you provide `id_list` (which is a good idea), the volume per atom should rather be:

```
volume_per_atom = self.get_volume(per_atom=True)
if id_list is not None:
    volume_per_atom = self.get_volume()/len(id_list)
```

pyiron/atomistics/structure/atoms.py
niklassiemer: Might be correct. Summing over Booleans is converted to summing 1 or 0, axis=-1 tells to sum all axes except the first one. Therefore, the result is a vector with the dimension of the inner vectors. The maximum value of the vector, indeed, is the number of neighbors I am searching. Please correct me, if I understand any of the points wrong!

pyiron/atomistics/structure/atoms.py
niklassiemer: To make it dependent on the density is something nice, however, the volume within the sphere would be 
`4/3 * np.pi * cutuff_radius**3` and thus 
```
volume_per_atom = self.get_volume(per_atom=True)
if id_list is not None:
    volume_per_atom = self.get_volume()/len(id_list)
num_neighbors = int(1.1*4./3.*np.pi*cutoff_radius**3/volume_per_atom)
```

pyiron/atomistics/structure/atoms.py
niklassiemer: No, I do not think so. I will remove this.

pyiron/atomistics/structure/atoms.py
niklassiemer: Yes, fine with me.

niklassiemer:
> Since I personally never use `get_neighbors_by_distance`, I'm fine with any implementation. I just want to say that if the goal is to get a good number of neighbors, this algorithm is probably a bit too expensive, because it internally runs `_get_neighbors` possibly multiple times (which itself is a relatively expensive function). Besides, there's still no guarantee that you get the correct number of neighbors in the end, since `_get_neighbors` itself relies on `_get_boundary_layer_width`, which estimates the width required from the density. In total, what in my opinion makes the most sense is to estimate the number of atoms from the global density (or Voronoi, if you really want to be sure), i.e. it should be:
> 
> ```
> num_neighbors = (1+buffer)*length**3/volume_per_atom
> ```
> 
> And maybe it's up to the user to determine `buffer`, or you can set a ridiculously large number for `buffer`, since if the number of neighbors is too large, the tree search simply ends so you will probably not lose much computation time.

Thank you very much for your comments! 
I am not married to the new implementation. I just had the feeling, that having an non obvious upper bound for the number of neighbors is counter intuitive for a function like `get_neighbors_by_distance`. Like in the notebook, where we compute an rdf with 8 Angstrom sphere, there are more than 100 neighbors to consider. In the end, I also would rather use just `get_neighbors` with an appropriate num_neighbors, since this is probably faster and returns a proper matrix instead of nested lists. 
I like the idea of giving just large initial guess for `num_neighbors` :
```
volume_per_atom = self.get_volume(per_atom=True)
if id_list is not None:
    volume_per_atom = self.get_volume()/len(id_list)
num_neighbors = int((1+buffer)*4./3.*np.pi*cutoff_radius**3/volume_per_atom)
```
with an default buffer of 20% and the possibility to extend it in the call.

niklassiemer:
Sorry, I did not want to close and reopen...

niklassiemer:
In the end, I would add a warning if the maximum number of neighbors found is identical to the `num_neighbors` specified/approximated, since this could indicate, that not all neighbors in the given distance have been found.

samwaseda:
> I just had the feeling, that having an non obvious upper bound for the number of neighbors is counter intuitive for a function like `get_neighbors_by_distance`.

Yes! Here I fully support your idea. I also like the fact that you gave a default cutoff radius.

> I like the idea of giving just large initial guess for `num_neighbors` :
> 
> ```
> volume_per_atom = self.get_volume(per_atom=True)
> if id_list is not None:
>     volume_per_atom = self.get_volume()/len(id_list)
> num_neighbors = int((1+buffer)*4./3.*np.pi*cutoff_radius**3/volume_per_atom)
> ```
> 
> with an default buffer of 20% and the possibility to extend it in the call.

Sounds great, but I guess since `get_neighbors_by_distance` makes more sense for non-uniform density and probably the number of neighbors doesn't affect the computation time, you can probably also give a very large number (this being said, I'm also fine with 20%)

samwaseda:
> In the end, I would add a warning if the maximum number of neighbors found is identical to the `num_neighbors` specified/approximated, since this could indicate, that not all neighbors in the given distance have been found.

Or you can also check whether all of the atoms get at least one 'np.inf'. Another possibility would be to transpose `neigh.distances` and check if the last component contains only `np.inf`, i.e. `np.all(neigh.distances.T[-1]==np.inf)`

niklassiemer:
Thanks a lot, conceptually it is the same, however, way more efficient!

pyiron/atomistics/structure/atoms.py
```diff
@@ -1118,32 +1118,95 @@ def get_extended_positions(self, width):
...
+                    tolerance=tolerance,
+                    id_list=id_list,
+                    cutoff_radius=cutoff_radius,
+                    boundary_width_factor=boundary_width_factor,
+                )
+                num_neighbors_old = num_neighbors
+                num_neighbors = np.sum(neigh.distances < np.inf, axis=-1).max()
+                if num_neighbors == num_neighbors_old:
+                    num_neighbors = 2 * num_neighbors
+        return num_neighbors
```
samwaseda, line 1173: It's maybe my personal preference, but I might find it very useful if this function returns the number of neighbors for each atom instead of returning the maximum number, because getting the maximum from a numpy array should be easy, since it only requires `.max()`

niklassiemer: True, and since we do not need it internally any more, this is added value.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1118,32 +1118,105 @@ def get_extended_positions(self, width):
...
+            id_list=None,
+            boundary_width_factor=1.2,
+            num_neighbors_estimate_buffer=1.0,
+    ):
+        """
+        Function to compute the maximum number of neighbors in a sphere around each atom.
+        Args:
+            cutoff_radius (float): Upper bound of the distance to which the search must be done
+            num_neighbors (int/None): maximum number of neighbors found
+            tolerance (int): tolerance (round decimal points) used for computing neighbor shells
```
samwaseda, line 1135: Since this function doesn't consider the shells, you can get rid of this argument.

niklassiemer: Thanks. I will take it out

samwaseda:
Very nice work! I merge this one now because I'd like to add (unrelated) modifications to `get_distance_by_distance` as well.

niklassiemer:
Thanks a lot for your comment and for the helpful review! I probably should have said that this is ready to merge in my opinion...

PR #1122, opened 2020-10-30
Fix Lammps Logfile location

jan-janssen:


jan-janssen:
This additionally requires https://github.com/pyiron/pylammpsmpi/pull/48 to fix the logging of parallel Lammps library jobs.

PR #1123, opened 2020-10-30
Add camera option to animate_structure as in plot3d

pmrv:
This just passes the parameter through to the NGL object.


jan-janssen (approved):
Looks good to me

pmrv:
I should also add that this makes the "orthographic" view the default instead of the "perspective" before, but I feel it's also much more useful.

jan-janssen:
As it is consistent with `plot3d()` I am fine with it

PR #1125, opened 2020-10-30
Use Docker file for mybinder

jan-janssen:


muh-hassani:
@jan-janssen why did you close this?

jan-janssen:
Because mybinder still triggers a new build every time we commit to the master branch, the only way to change this is a separate repository. Still it might make sense to have a docker container with all the dependencies, that would also accelerate our testing infrastructure.

muh-hassani:
Oh, I see. So it can't be used in this case :(

jan-janssen:
I created a pull request to build only the pyiron dependencies.

PR #1129, opened 2020-10-30
Extend elastic

samwaseda:
Add certain functionalities to the elastic class (e.g. rotations and global elastic constants)


samwaseda:
This class should be reviewed only after [this PR](https://github.com/pyiron/pyiron/pull/1079) is merged.

PR #1133, opened 2020-11-02
Lammps Structure raise ValueError 

jan-janssen:


pyiron/lammps/structure.py
```diff
@@ -613,7 +613,10 @@ def structure_atomic(self):
         for id_eam, el_eam in enumerate(self._el_eam_lst):
             if el_eam in el_struct_lst:
                 id_el = list(el_struct_lst).index(el_eam)
-                el = el_obj_lst[id_el]
+                if id_el in el_obj_lst.keys():
```
samwaseda, line 616: Isn't this a list and not a dictionary?

jan-janssen: You are right - I was going over it too quickly

samwaseda (approved):
Very good! Thanks for the quick fix as well.

PR #1141, opened 2020-11-03
Fix hdf

sudarsan-surendralal:
Changes in parameter name to allow `to_object()`

### DIFF
--- a/pyiron/atomistics/volumetric/generic.py
+++ b/pyiron/atomistics/volumetric/generic.py
@@ -270,32 +270,32 @@ class VolumetricData(object):
         else:
             return np.average(np.average(self._total_data, axis=0), 0)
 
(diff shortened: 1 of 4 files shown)


samwaseda:
I guess I need a bit more explanation to understand what was the problem

sudarsan-surendralal:
The hdf class in `pyiron_base` has a nice `to_object` method which allows us to create objects from their hdf5 path alone (without loading any jobs) which makes things much faster: https://github.com/pyiron/pyiron_base/blob/d14fa9bb4f4d0fcc0e45b7d31cc0ba9ed163ee94/pyiron_base/generic/hdfio.py#L1287

This didn't work for some objects since the `from_hdf()` has the wrong parameter name `hdf5` instead of `hdf`. So just fixed this

samwaseda (approved):

Issue #1067, opened 2020-10-17
Scipyoptimizer doesn't work in interactive mode (with GPAW)


cmmmalik:
Scipyoptimizer gives out an error when it is used with GPAW for relaxation. The problem occurs when interaction mode(which must be set True for scipyoptimizer) is turned on. See the below code and output

```from pyiron import Project
pr = Project("gpaw")
job = pr.create_job(pr.job_type.Gpaw, "relaxation")
job.structure = pr.create_ase_bulk("Al")
job.server.run_mode.interactive =True
minm = job.create_job(pr.job_type.ScipyMinimizer,"test_rel")
minm.run()
```
**output**
```The job test_rel was saved and received the ID: 1
The job relaxation was saved and received the ID: 2
<ipython-input-8-094269622dd3> in <module>
----> 1 minm.run()

~/git_packages/pyiron_base/pyiron_base/job/generic.py in run(self, delete_existing_job, repair, debug, run_mode, run_again)
    681                 status = "created"
    682             if status == "initialized":
--> 683                 self._run_if_new(debug=debug)
    684             elif status == "created":
    685                 que_id = self._run_if_created()

~/git_packages/pyiron_base/pyiron_base/job/generic.py in _run_if_new(self, debug)
   1391         else:
   1392             self.save()
-> 1393             self.run()
   1394 
   1395     def _run_if_created(self):

~/git_packages/pyiron_base/pyiron_base/job/generic.py in run(self, delete_existing_job, repair, debug, run_mode, run_again)
    683                 self._run_if_new(debug=debug)
    684             elif status == "created":
--> 685                 que_id = self._run_if_created()
    686                 if que_id:
    687                     self._logger.info(

~/git_packages/pyiron_base/pyiron_base/job/generic.py in _run_if_created(self)
   1408             self.run_if_manually()
   1409         elif self.server.run_mode.modal:
-> 1410             self.run_static()
   1411         elif self.server.run_mode.non_modal or self.server.run_mode.thread:
   1412             self.run_if_non_modal()

~/git_packages/pyiron/pyiron/interactive/scipy_minimizer.py in run_static(self)
     76                 gtol=self.input["ionic_force_tolerance"],
     77                 disp=False,
---> 78                 full_output=True,
     79             )
     80             self.output._convergence = output[4]

[...]

PR #1142, opened 2020-11-03
Scipy unit

samwaseda:
Problem stated [here](https://github.com/pyiron/pyiron/issues/1067).

With this pull request, the target pressure can be put via:

- A number (e.g. `pressure=0`) if hydrostatic and isotropic
- A list (e.g. `pressure=[0, 0, 0]`) if anisotropic
- Voigt representation (e.g. `pressure=[0, 0, 0, 0, 0, 0]`) if also shear component
- Tensor (e.g. `pressure=np.zeros((3,3))`)


samwaseda:
And indeed this one is an enhancement and a bugfix at the same time...

pyiron/interactive/scipy_minimizer.py
```diff
@@ -110,14 +123,42 @@ def run_static(self):
         else:
             self.status.not_converged = True
 
+    @staticmethod
+    def _tensor_to_voigt(s, strain=False):
```
jan-janssen, line 127: Can We use the same for Lammps `calc_md` and `calc_minimize`? - Does it make sense to have a `voigt` module or something like this?

samwaseda: Yes probably it does, but I'd say that's a bit more than just a few extra lines that I would add in this PR because it sounds again like another Pandora's box...

PR #1150, opened 2020-11-04
add remove_inf in order to give the possibility of getting numpy arra…

samwaseda:
…y instead of list

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1186,22 +1186,30 @@ class Atoms(ASEAtoms):
         id_list=None,
         boundary_width_factor=1.2,
         num_neighbors_estimate_buffer=1.0,
+        remove_inf=True,
     ):
         """
 
         Args:
-            cutoff_radius (float): Upper bound of the distance to which the search must be done
-            num_neighbors (int/None): maximum number of neighbors found; if None this is estimated based on the density.
-            t_vec (bool): True: compute distance vectors
-                        (pbc are automatically taken into account)
-            tolerance (int): tolerance (round decimal points) used for computing neighbor shells
+            cutoff_radius (float): Upper bound of the distance to which the
+                search must be done
+            num_neighbors (int/None): maximum number of neighbors found; if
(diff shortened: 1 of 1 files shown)


pyiron/atomistics/structure/atoms.py
```diff
@@ -1186,22 +1186,30 @@ def get_neighbors_by_distance(
...
+            num_neighbors_estimate_buffer (float): Extra volume taken into
+                account for the num_neighbors estimation
+            remove_inf (bool): Whether or not remove np.inf which fill all
+                slots of the neighbors above distance limit
 
-            pyiron.atomistics.structure.atoms.Neighbors: Neighbors instances with the neighbor indices, distances
-            and vectors
+        Returns:
+            pyiron.atomistics.structure.atoms.Neighbors: Neighbors instances
+                with the neighbor indices, distances and vectors
```
niklassiemer, line 1212: Could we add the different output of the function here as well? 
as list of np.arrays, each (or as np.array if  remove_inf=False)

samwaseda: aaaah it's true that it's maybe not good that the output type depends on that... Let me think about it.

niklassiemer: Otherwise, the neighbors object usually has these quantities as np.arrays instead of as lists... 
How about another boll `as_np_arrays=False` since this would directly tell the user that the ouput is converted.
The docstring could be along the lines.
as_np_arrays (bool): Return indices, distances, and vectors as np_arrays instead of lists; automatically sets remove_inf=True

If this is not true and `remove_inf=False` we would return a list including np.inf... However, I do not know any usecase for this... Maybe even the bare renaming to `as_np_arrays` plus a docstring that this omits the removal of np.inf in these quantites would help?

samwaseda: Well, for the very limited use case that I had in mind I don't think it's worth creating a complicated argument, so I guess I'm gonna close this PR. Thank you for taking your time @niklassiemer (and especially noticing that I was about to do something that I wouldn't have wanted myself)

PR #1151, opened 2020-11-05
Convergence check NELM=1

### LINKED ISSUE #1148: VASP RPA calculation
When preparing for an RPA calculation, using NELM=1, the following line will always result in a not_converged status:

https://github.com/pyiron/pyiron/blob/c17d4f3628f2ad5c3d5f8485d367e835c8361ebb/pyiron/vasp/base.py#L425

I noticed this when trying to make a FlexibleMaster job which executes a series of VASP calculations, which halts when encountering this not_converged job. Is there some way to circumvent this?

SanderBorgmans:
This should fix the not_converged status of NELM=1 jobs for ALGO=EXACT or EIGENVAL
Should fix #1148


sudarsan-surendralal (approved):
Looks good to me. Thanks!

PR #1153, opened 2020-11-05
Get extended positions

samwaseda:
Since we started using `get_extended_positions` also independently of `get_neighbors` and I find it quite confusing that it not only returns the distances but also indices, I made indices an option.


samwaseda:
Sorry I don't understand git here. I'm gonna close the pull request to start over.

PR #1154, opened 2020-11-05
add return_indices in get_extended_positions

samwaseda:
Since we started using get_extended_positions also independently of get_neighbors and I find it quite confusing that it not only returns the distances but also indices, I made indices an option.

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1087,7 +1087,7 @@ class Atoms(ASEAtoms):
         xyz = self.get_scaled_positions(wrap=False)
         return xyz[:, 0], xyz[:, 1], xyz[:, 2]
 
-    def get_extended_positions(self, width):
+    def get_extended_positions(self, width, return_indices=False):
         """
         Get all atoms in the boundary around the supercell which have a distance
         to the supercell boundary of less than dist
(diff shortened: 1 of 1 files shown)


pyiron/atomistics/structure/atoms.py
```diff
@@ -1096,7 +1096,9 @@ def get_extended_positions(self, width):
             dist (float): Distance in Angstrom
 
         Returns:
-            pyiron.atomistics.structure.atoms.Atoms: The required boundary region
+            pyiron.atomistics.structure.atoms.Atoms, numpy.ndarray:
+                Positions of all atoms in the extended box, indices of atoms in
+                their original option (if return_indices=True)
```
niklassiemer, line 1101: Is this function currently used by users or is this intended for internal use? 
Since this changes the default behavior, we could include ehm, a note? (Do we have that concept?) 
The default behavior of this function changed as of ... - To restore the old behavior specify `return_indices=True`

samwaseda: Oh in general you're right, but since this function is still extremely new (maybe a month old?) and I never introduced it anywhere, I don't think there's anyone using it.

niklassiemer: Ok, I just thought I would run a notebook and it just does not work anymore xD
Then I am fine with it!

niklassiemer (approved):

PR #1157, opened 2020-11-05
SQS: Fix input list

jan-janssen:


jan-janssen:
@liamhuber It is an emergency fix for the workshop next week but does not solve the general issue of the input list class.

PR #1160, opened 2020-11-06
replace ase get_distances_array

samwaseda:
I replaced the lines of `get_distances_array` since the ASE version was extremely slow. Now it's more than 10x faster.


pyiron/atomistics/structure/atoms.py
```diff
@@ -1997,20 +1995,22 @@ def get_distances_array(self, a0=None, a1=None, mic=True, vector=False):
...
             a0 = a1
-            a1 = None
         if a0 is None:
-            a0 = self.positions
+            a0 = a1 = self.positions
         a0 = np.array(a0).reshape(-1, 3)
-        if a1 is not None:
-            a1 = np.array(a1).reshape(-1, 3)
+        a1 = np.array(a1).reshape(-1, 3)
+        x = np.einsum(
```
sudarsan-surendralal, line 2002: A nitpick on coding style: It'll be nice to have a better descriptive variable than `x`. I know we can't change `a0` and `a1` since we have to be compatible with ASE but maybe something like `diff_relative` would make this much more readable

sudarsan-surendralal:
```
ERROR: test_get_distances_array (atomistics.structure.test_atoms.TestAtoms)
  File "/home/runner/work/pyiron/pyiron/tests/atomistics/structure/test_atoms.py", line 875, in test_get_distances_array
    self.assertTrue(np.allclose(basis.get_distances_array(a0=0.5*np.ones(3)),
  File "/home/runner/work/pyiron/pyiron/pyiron/atomistics/structure/atoms.py", line 2001, in get_distances_array
    a1 = np.array(a1).reshape(-1, 3)
ValueError: cannot reshape array of size 1 into shape (3)

```

The function fails of `a0` is given as a position rather than a list of positions.

pyiron/atomistics/structure/atoms.py
samwaseda: Actually `get_distances` in [ase](https://wiki.fysik.dtu.dk/ase/_modules/ase/geometry/geometry.html#get_distances) is `p0` and `p1`, so we are anyway somewhat changing the names...

sudarsan-surendralal: In that case we should also refractor to use `p0` and `p1`. This will help make our the transition from ase workflows to pyiron seamless

samwaseda: Done!

samwaseda: Oh no it's `p1` and `p2`...

samwaseda: Done!

sudarsan-surendralal:
```
  File "/home/runner/work/pyiron/pyiron/tests/atomistics/structure/test_atoms.py", line 875, in test_get_distances_array
    self.assertTrue(np.allclose(basis.get_distances_array(a0=0.5*np.ones(3)),
TypeError: get_distances_array() got an unexpected keyword argument 'a0'
```

I think you have to change the arugments in this test too

samwaseda:
Maybe I'm gonna wait for [this PR](https://github.com/pyiron/pyiron/pull/1163) to be merged first, because the two functions do almost the same thing -> for `get_distances_array` I can use `find_mic` internally (other way around is not possible)

sudarsan-surendralal (approved):
Looks good!

PR #1163, opened 2020-11-08
Find mic

samwaseda:
I found `find_mic` in ASE hopelessly slow and I find it too tiring to look for the path every time, so I implemented it in pyiron. Below you can find a speed comparison - the new implementation is more than an order of magnitude faster.

Unknown

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1944,8 +1944,7 @@ class Atoms(ASEAtoms):
 
         Args:
             v (list/numpy.ndarray): 3d vector or a list/array of 3d vectors
-            vectors (bool): Whether to return vectors (distances are returned
-                if False)
+            vectors (bool): Whether to return vectors (distances are returned if False)
 
         Returns: numpy.ndarray of the same shape as input with mic
         """
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1952,7 +1952,7 @@ class Atoms(ASEAtoms):
         vecs = np.asarray(v).reshape(-1, 3)
         if any(self.pbc):
             vecs = np.einsum('ji,nj->ni', np.linalg.inv(self.cell), vecs)
-            vecs[self.pbc] -= np.rint(vecs)[self.pbc]
+            vecs[:,self.pbc] -= np.rint(vecs)[:,self.pbc]
             vecs = np.einsum('ji,nj->ni', self.cell, vecs)
         if vectors:
             return vecs.reshape(np.asarray(v).shape)
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1937,6 +1937,27 @@ class Atoms(ASEAtoms):
         )
         return self.analyse_ovito_voronoi_volume()
 
+    def find_mic(self, v, vectors=True):
+        """
+        Find vectors following minimum image convention (mic). In principle this
+        function does the same as ase.geometry.find_mic
+
+        Args:
+            v (list/numpy.ndarray): 3d vector or a list/array of 3d vectors
+            vectors (bool): Whether to return vectors (distances are returned
+                if False)
+
+        Returns: numpy.ndarray of the same shape as input with mic
+        """
+        vecs = np.asarray(v).reshape(-1, 3)
+        if any(self.pbc):
+            vecs = np.einsum('ji,nj->ni', np.linalg.inv(self.cell), vecs)
+            vecs[self.pbc] -= np.rint(vecs)[self.pbc]
+            vecs = np.einsum('ji,nj->ni', self.cell, vecs)
+        if vectors:
+            return vecs.reshape(np.asarray(v).shape)
+        return np.linalg.norm(vecs, axis=-1).reshape(np.asarray(v).shape[:-1])
+
     def get_distance(self, a0, a1, mic=True, vector=False):
         """
         Return distance between two atoms.
--- a/tests/atomistics/structure/test_atoms.py
+++ b/tests/atomistics/structure/test_atoms.py
@@ -14,6 +14,7 @@ from pyiron.atomistics.structure.periodic_table import PeriodicTable, ChemicalEl
 from pyiron_base import FileHDFio, ProjectHDFio, Project
 from ase.cell import Cell as ASECell
 from ase.atoms import Atoms as ASEAtoms
+from ase.geometry.geometry import find_mic
 
 
 class TestAtoms(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


samwaseda:
By the way, the speed test was done for an implementation where both distances and vectors are returned (which has now been disabled because I don't really see why someone should need both of them returned). So in reality if only the vectors are returned, which I imagine is the most common way that this function is used, then it's likely that it's even faster than the graph above.

sudarsan-surendralal:
```
FAIL: test_find_mic (atomistics.structure.test_atoms.TestAtoms)
  File "/home/runner/work/pyiron/pyiron/tests/atomistics/structure/test_atoms.py", line 879, in test_find_mic
    self.assertAlmostEqual(
AssertionError: 1.0139157955647577 != 0 within 7 places (1.0139157955647577 difference)

```

Is this simply the cell shifted by a lattice vector?

samwaseda:
> ```
> FAIL: test_find_mic (atomistics.structure.test_atoms.TestAtoms)
>   File "/home/runner/work/pyiron/pyiron/tests/atomistics/structure/test_atoms.py", line 879, in test_find_mic
>     self.assertAlmostEqual(
> AssertionError: 1.0139157955647577 != 0 within 7 places (1.0139157955647577 difference)
> 
> ```
> 
> 
> Is this simply the cell shifted by a lattice vector?

Oh interesting. Maybe yes. It never failed on my local computer, but since I use random variables there, maybe something rather unexpected happened.

samwaseda:
ok I just let it run 100 times and it happened a few times. I guess there's a slight shift of box in ASE.

samwaseda:
In my opinion, ASE is wrong:

```
from ase.geometry.geometry import find_mic
import numpy as np

# look for a vector which gives conflict between ASE and my new implementation:
for ii in range(100):
    cell = 0.1*(np.random.random((3,3))-0.5)+np.eye(3)
    basis = Atoms("Fe", positions=[3*[0.5]], cell=cell, pbc=True)
    absolute_vector = 2*np.random.random((1, 3))-1
    vecs, dists = find_mic(absolute_vector, cell, pbc=True)
    vec_diff = vecs-basis.find_mic(absolute_vector, vectors=True)
    if not np.isclose(np.linalg.norm(vec_diff, axis=-1).max(), 0):
        print('Error: vector', ii, vec_diff)
        break
    dist_diff = dists-basis.find_mic(absolute_vector, vectors=False)
    if not np.isclose(np.linalg.norm(dist_diff, axis=-1).max(), 0):
        print('Error: distance', ii, dist_diff)
        break

# get relative vector (scaled coordinates)
relative_vector = np.linalg.inv(cell.T).dot(absolute_vector[-1])

# check relative vector times cell vectors gives the original vector
assert np.isclose(
    relative_vector[0]*cell[0]+relative_vector[1]*cell[1]+relative_vector[2]*cell[2], absolute_vector[0]
).all()

# Components with a length more than half the box is wrapped
relative_vector -= np.rint(relative_vector)

# Check relative vector times cell vectors gives my find_mic -> this one is fine
assert np.isclose(
    relative_vector[0]*cell[0]+relative_vector[1]*cell[1]+relative_vector[2]*cell[2],
    basis.find_mic(absolute_vector, vectors=True)[0]
).all()

# Check relative vector times cell vectors gives ASE find_mic -> Error
assert np.isclose(
    relative_vector[0]*cell[0]+relative_vector[1]*cell[1]+relative_vector[2]*cell[2],
    find_mic(v, cell=basis.cell, pbc=True)[0]
).all()

```

pyiron/atomistics/structure/atoms.py
```diff
@@ -1937,6 +1937,27 @@ def get_voronoi_volume(self):
         )
         return self.analyse_ovito_voronoi_volume()
 
+    def find_mic(self, v, vectors=True):
+        """
+        Find vectors following minimum image convention (mic). In principle this
```
liamhuber, line 1942: Nit: grammar in docstrings; please end sentences with a period.

liamhuber, line 1942: I am ok accepting this as is, but putting myself in the shoes of a new MSc student I see that it might be useful to include a very brief description of what "minimum image convention" is.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1937,6 +1937,27 @@ def get_voronoi_volume(self):
...
+        Args:
+            v (list/numpy.ndarray): 3d vector or a list/array of 3d vectors
+            vectors (bool): Whether to return vectors (distances are returned
+                if False)
+
+        Returns: numpy.ndarray of the same shape as input with mic
+        """
+        vecs = np.asarray(v).reshape(-1, 3)
+        if any(self.pbc):
+            vecs = np.einsum('ji,nj->ni', np.linalg.inv(self.cell), vecs)
```
liamhuber, line 1954: what I see here

pyiron/atomistics/structure/atoms.py
```diff
@@ -1937,6 +1937,27 @@ def get_voronoi_volume(self):
...
 
+    def find_mic(self, v, vectors=True):
+        """
+        Find vectors following minimum image convention (mic). In principle this
+        function does the same as ase.geometry.find_mic
+
+        Args:
+            v (list/numpy.ndarray): 3d vector or a list/array of 3d vectors
+            vectors (bool): Whether to return vectors (distances are returned
+                if False)
```
liamhuber, line 1948: In Soviet Russia you don't review code, code reviews YOU!

...what length are we using for line breaks? I've been using 120, where this line break is unnecessary...but I just checked PEP8 and Guido recommends 79! Most of the code base seems to use 120 (where the code base cares about this at all...). This is probably because 120 is the pycharm default. 

This is a place I can accept violating PEP8, but if that's what we decide we should do it consciously, not by accident like me.

tests/atomistics/structure/test_atoms.py
```diff
@@ -14,6 +14,7 @@
 from pyiron_base import FileHDFio, ProjectHDFio, Project
 from ase.cell import Cell as ASECell
 from ase.atoms import Atoms as ASEAtoms
+from ase.geometry.geometry import find_mic
```
liamhuber, line 17: I get a bit of an uneasy feeling having our test simply run our code against theirs. It's not something I'd say needs to change before the code is merged and the only easy brute force test I can think of is making sure that none of the distances exceed more than half the longest vector that can sit inside the box (e.g. corner-to-corner). 

However, I looked back in more detail at your conversation just now and I guess my unease is not unjustified. I have no suggestion atm, but I would be happier with an alternative if you can think of one.

liamhuber (approved):
I have some questions and minor stuff I'd *like* to see changed, but nothing that would prohibit merging if you push ahead without first appeasing me ;)

pyiron/atomistics/structure/atoms.py
samwaseda: yes, I strictly followed the 80 character limit (79? what's the matter with that guy?). But it's true that this looks horrible, so I'm gonna rather follow your suggestion :D

tests/atomistics/structure/test_atoms.py
samwaseda: I don't know if the changes I just made answer what you're saying, but I anyway removed the ASE `find_mic`

PR #1170, opened 2020-11-11
Fix neighbors

sudarsan-surendralal:
The current implementation of `get_neighbors_by_distance` would fail in cases where the volume per atom of the supercell is large. For example

```
import numpy as np
from pyiron import Project

pr = Project(".")
dx = 0.7
r_O = [0, 0, 0]
r_H1 = [dx, dx, 0]
r_H2 = [-dx, dx, 0]
unit_cell = 10 * np.eye(3)
water = pr.create_atoms(elements=['H', 'H', 'O'],
                        positions=[r_H1, r_H2, r_O],
                        cell=unit_cell, pbc=True)
water.get_neighbors_by_distance(1.3)
```

I've just made a simple fix so that at least 1 neighbor is counted.

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1233,7 +1233,7 @@ class Atoms(ASEAtoms):
             volume_per_atom = self.get_volume(per_atom=True)
             if id_list is not None:
                 volume_per_atom = self.get_volume() / len(id_list)
-            num_neighbors = max(1, int((4 + num_neighbors_estimate_buffer) *
+            num_neighbors = max(4, int((1 + num_neighbors_estimate_buffer) *
                                        4. / 3. * np.pi * cutoff_radius ** 3 / volume_per_atom))
 
         neigh = self._get_neighbors(
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -1233,7 +1233,7 @@ class Atoms(ASEAtoms):
             volume_per_atom = self.get_volume(per_atom=True)
             if id_list is not None:
                 volume_per_atom = self.get_volume() / len(id_list)
-            num_neighbors = max(1, int((1 + num_neighbors_estimate_buffer) *
+            num_neighbors = max(1, int((4 + num_neighbors_estimate_buffer) *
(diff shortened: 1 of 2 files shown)


pyiron/atomistics/structure/atoms.py
```diff
@@ -1234,8 +1233,8 @@ def get_neighbors_by_distance(
             volume_per_atom = self.get_volume(per_atom=True)
             if id_list is not None:
                 volume_per_atom = self.get_volume() / len(id_list)
-            num_neighbors = int((1 + num_neighbors_estimate_buffer) *
-                                4. / 3. * np.pi * cutoff_radius ** 3 / volume_per_atom)
+            num_neighbors = max(1, int((1 + num_neighbors_estimate_buffer) *
+                                       4. / 3. * np.pi * cutoff_radius ** 3 / volume_per_atom))
```
niklassiemer, line 1237: We could also just do `int(...)+1`

niklassiemer (approved):
Yes, we should make sure the estimate is at least 1.

pyiron/atomistics/structure/atoms.py
sudarsan-surendralal: Yes but find this a bit more intuitive and saves (even though only a little) computation time

samwaseda: Is 1 a really solid estimation? I'm asking this because while for a large radius the estimation for the number of atoms from the density is a robust algorithm, there might be strong fluctuations for a very small radius. In other words, we know from a usual radial distribution function that the value becomes stable at high radii, the behaviour for small radii might become quite unstable. If taking a few atoms is still acceptable from the computational point of view, I might put 3 or 4.

sudarsan-surendralal: Good point!

samwaseda (approved):

sudarsan-surendralal:
@jan-janssen is it possible to update the cmti version to the master branch? This bugfix is kind of important since it breaks the workflows for many users (electrochemical simulations)

jan-janssen:
The cluster is using the latest version on conda. I released a new version but I would like to briefly discuss the changes before updating the cluster.

PR #1190, opened 2020-11-13
Revert "SciPyMinimizer: Keep default values from self.input"

samwaseda:
Reverts pyiron/pyiron#1095

Sorry I have to revert this one since the question remained unanswered.

### DIFF
--- a/pyiron/interactive/scipy_minimizer.py
+++ b/pyiron/interactive/scipy_minimizer.py
@@ -233,9 +233,9 @@ class ScipyMinimizer(InteractiveWrapper):
         max_iter=100,
         pressure=None,
         algorithm='CG',
-        ionic_energy_tolerance=None,
-        ionic_force_tolerance=None,
-        pressure_tolerance=None,
+        ionic_energy_tolerance=0,
+        ionic_force_tolerance=1.0e-2,
+        pressure_tolerance=1.0e-3,
         volume_only=False,
     ):
         """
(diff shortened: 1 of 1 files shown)


jan-janssen:
The question was open for more than a week.

samwaseda:
> The question was open for more than a week.

It doesn't mean that you can merge it by default. I'm not saying you can't, but it's not an argument that justifies merging.

jan-janssen:
I try to keep the project structured and close issues and pull requests or preferably merge them if I do not see any progress or no further contributions to the discussion.

samwaseda:
> I try to keep the project structured and close issues and pull requests or preferably merge them if I do not see any progress or no further contributions to the discussion.

Then you can also scrap the changes. I don't see why you should merge them forcibly.

The usage that you mentioned first is still not quite a striking example for me:

```
minm = job.create_job(pr.job_type.ScipyMinimizer, "test_rel")
minm.input.ionic_force_tolerance = 0.001
minm.calc_minimize(pressure=0, volume_only=True)
```

For me setting input values is either via `job.input` or through setter functions (in this case `calc_minimize`, which I consider a setter function since SciPy minimizer doesn't do anything else). The example above mixes them up. Again, I'm also fine with the idea that untouched arguments don't do anything, but since I know that in different modules the behaviour is different and it's also a conceptual question, I can't just let it go without having talked about it.

liamhuber:
@samwaseda sorry, I missed the original review request by a million years.

I'm not familiar with the class, but it sounds like it's a high level question of whether or not `job.advanced_setter` defaults should overwrite `job.input` fields, especially if those fields have been explicitly written to. 

This sounds like something that we should move to the discussion board. If I've summarized the problem nicely above you could open with that, otherwise please correct my misunderstanding.

To give you a sneak-peek at my answer: this smells really bad! The "setter" is way to powerful if this is even coming up. My preferred solution is to change our hierarchy, e.g. by breaking up the `job` into something that interprets positions and returns forces and energies, and something that takes positions and forces and evolves the positions.

Regarding closing/merging, I think my tolerance for open issues and merge requests is higher than @jan-janssen's, but I have no objection to cleaning them up. I agree with Sam though that they should be closed rather than merged if there was not a clearly positive conclusion before they went stale.

PR #1193, opened 2020-11-14
Pyscal consistency

liamhuber:
Very minor touchups to the wonderful [switch from ovito to pyscal](https://github.com/pyiron/pyiron/pull/1192)

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -2018,10 +2018,10 @@ class Atoms(ASEAtoms):
         """
         warnings.warn(
             "This function doesn't account for periodic boundary conditions. Call "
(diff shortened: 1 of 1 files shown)


liamhuber:
I really appreciate [this switch](https://github.com/pyiron/pyiron/pull/1192). Just this weekend I ran into needing voronoi analysis on a system that doesn't have ovito, so the timing couldn't be better 👍

liamhuber:
I don't think this requires review, nor even that it stay up -- it's effectively just documentation improvement.

As soon as the tests complete I'll merge it. I just with through the PR process so there's a clean record of the change.

jan-janssen (approved):
Looks good to me - sorry seems like I was a bit tired after this week.

liamhuber:
@jan-janssen no worries! It happens and the beauty of this workflow is that patches like this are so easy 🙂 lots of great changes lately!

PR #1194, opened 2020-11-15
[minor] add new line in spins.in

samwaseda:
Add new line in `spins.in` which bash currently thinks doesn't end, which (in my personal case) is causing parsing problems.


samwaseda:
I didn't wait 24h but since this is a super minor change I merge it now.

Issue #1195, opened 2020-11-16
Safer phonopy operation


liamhuber:
**Summary**

Phonopy functions by taking a reference, calculating symmetry, making a series of small displacements, and assessing the forces and energies at each of these displacements. If the reference job is **anything except `calc_static`**, phonopy *runs* but gives crazy results. 

This is obvious in retrospect -- imagine a `calc_minimize` ref, every displacement gets undone by relaxing! -- but has been a problem more than once so let's just catch it. It's especially a pain because the only way you can tell something's wrong is that the physics are bad, there's no code error atm.

**Detailed ~~Description~~ architecture rant**

This really [brings me back](https://github.com/orgs/pyiron/teams/pyiron/discussions/25) to the fact that we should be separating out the *representations* of our structures (Lammps, Vasp, etc), from the *evolution* of our structure (static (i.e. none), MD, minimization, or more complex evolutions like phonopy which branch the evolution into many dimensions, each of which has a small displacement).

If how we evaluated forces and energies from a structure, and how we changed that structure and interpreted those energies were better separated, we would never run into problems like this.

I'm not saying it should already be this way and that we were wrong to take the path we took to get where we are now, just that looking forward it's clear there are better places to go.

`/end_rant`

**Further Information, Files, and Links**

I made a [small patch for this](https://github.com/pyiron/pyiron/pull/1196), but haven't tested it or fleshed it out. I think this is an excellent topic for one of our more junior developers to take over and finish that PR!

liamhuber:
Fixed in #1196

PR #1196, opened 2020-11-16
Add check that reference job is static

liamhuber:
If you pass a `calc_md` or `calc_minimize` job as a phonopy reference, phonopy will run but will silently return crazy results (at least for minimize, not 100% sure for md). Let's stop that by validating the input.

I draft the solution here, but don't have time to test it and bring it all the way home. I think this is an excellent topic for one of our junior developers to take over and finish this PR!

[Linked issue](https://github.com/pyiron/pyiron/issues/1195).

### DIFF
--- a/pyiron/atomistics/master/phonopy.py
+++ b/pyiron/atomistics/master/phonopy.py
@@ -393,3 +393,10 @@ class PhonopyJob(AtomisticParallelMaster):
         ax.set_ylabel("DOS")
         ax.set_title("Phonon DOS vs Energy")
         return ax
+
+    def validate_ready_to_run(self):
+        if self.ref_job._generic_input["calc_mode"] != "static":
+            raise ValueError("Phonopy reference jobs should be static calculations, but got {}".format(
+                self.ref_job._generic_input["calc_mode"]
+            ))
+        super().validate_ready_to_run()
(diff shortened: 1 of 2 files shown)


caglayanoaras:
I did some quick local tests and this works for me. Am I supposed to something else?

liamhuber:
> I did some quick local tests and this works for me. Am I supposed to something else?

Awesome, thanks @caglayanoaras! 

Yes, please also go into the tests/atomistics/master/phonopy and add a test where we pass in a non-static job and force a failure. Probably it's important to do this by manually executing `phonopy_job.validate_ready_to_run` rather than actually executing `run` on the job, as this is slower and thus typically frowned upon in our unit tests.

caglayanoaras:
I have added a test function. One check is failing because of whitespace but it shouldn't fail in my opinion. It takes like an extra 0.3 seconds to run the test.
Any other suggestions or tasks?

jan-janssen (approved):
Looks good to me

caglayanoaras:
May I merge pull request and then delete phonopy_check branch? @liamhuber

liamhuber:
> May I merge pull request and then delete phonopy_check branch? @liamhuber

@caglayanoaras of course 😄

PR #1199, opened 2020-11-16
Spinoff structures

liamhuber:
The atomistic `Project` class already has a lot of methods for creating structures, and I can see that we will only have more coming. For instance, I want easy access to `ase.io.read`.

So here I package these methods up in a new class, and add that class as an attribute of `Project`. I.e. `project.create_ase_bulk('Al')` --> `project.structure.create_ase_bulk('Al')`. This is a bit more typing, but in cleans up and organizes our tab completion, and the class structure will help as we extend the ability for generating structures (e.g. making grain boundaries, dislocations, etc.)

What is *not* done here:
- New, nicely structured tests for the `StructureGenerator` class.
- Very many new extensions (only binding ASE's read).

What I most want feedback on is really serious thought as to what we should name this new attribute of `Project`, which I currently just call `structure`.

### DIFF
--- a/pyiron/atomistics/structure/factory.py
+++ b/pyiron/atomistics/structure/factory.py
@@ -123,7 +123,7 @@ class StructureFactory:
             return None
 
     @staticmethod
-    def hkl_surface(lattice, hkl, layers, vacuum=1.0, center=False, pbc=True):
+    def surface_hkl(lattice, hkl, layers, vacuum=1.0, center=False, pbc=True):
         """
         Use ase.build.surface to build a surface with surface normal (hkl).
 
--- a/pyiron/atomistics/structure/factory.py
+++ b/pyiron/atomistics/structure/factory.py
@@ -151,7 +151,7 @@ class StructureFactory:
         return ase_to_pyiron(surface)
 
     @staticmethod
-    def structure(element, bravais_basis, lattice_constant):
+    def crystal(element, bravais_basis, lattice_constant):
         """
         Create a crystal structure using pyiron's native crystal structure generator
 
--- a/pyiron/atomistics/structure/factory.py
+++ b/pyiron/atomistics/structure/factory.py
@@ -52,6 +52,7 @@ class StructureFactory:
         """
         Returns a ASE's read result, wrapped as a `pyiron.atomstic.structure.atoms.Atoms` object.
 
+        ase.io.read docstring:
         """
         return ase_to_pyiron(read(*args, **kwargs))
     ase_read.__doc__ += read.__doc__
--- a/pyiron/atomistics/structure/factory.py
+++ b/pyiron/atomistics/structure/factory.py
@@ -47,7 +47,7 @@ __date__ = "May 1, 2020"
 s = Settings()
 
 
-class StructureGenerator:
+class StructureFactory:
     def ase_read(self, *args, **kwargs):
         """
         Returns a ASE's read result, wrapped as a `pyiron.atomstic.structure.atoms.Atoms` object.
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -415,6 +415,7 @@ class LammpsControl(GenericParameters):
         tloop=None,
         initial_temperature=None,
         langevin=False,
+        drag=None,
         delta_temp=None,
         delta_press=None,
         job_name="",
@@ -454,6 +455,9 @@ class LammpsControl(GenericParameters):
                                                be used). If any other number is given, this value is going to be used
                                                for the initial temperature.
             langevin (bool): (True or False) Activate Langevin dynamics
+            drag (float): Drag to apply. Note: Using this flag will destroy your thermodynamics and you will no longer
+                acheive the NVT/NPT/NPH/NVE ensemble averages you would otherwise expect. (Default is None, no
+                additional drag (Langevin drag still applied in case of a Langevin thermostat).)
             delta_temp (float): Thermostat timescale, but in your Lammps time units, whatever those are. (DEPRECATED.)
             delta_press (float): Barostat timescale, but in your Lammps time units, whatever those are. (DEPRECATED.)
             job_name (str): Job name of the job to generate a unique random seed.
@@ -574,6 +578,9 @@ class LammpsControl(GenericParameters):
         if tloop is not None:
             fix_ensemble_str += " tloop " + str(tloop)
 
+        if drag is not None:
+            fix_ensemble_str += " drag " + str(drag)
+
         self.remove_keys(["minimize"])
         self.modify(
             fix___ensemble=fix_ensemble_str,
@@ -619,6 +626,7 @@ class LammpsControl(GenericParameters):
         seed=None,
         initial_temperature=None,
         langevin=False,
+        drag=None,
         job_name="",
         rotation_matrix=None
     ):
@@ -672,6 +680,7 @@ class LammpsControl(GenericParameters):
             tloop=None,
             initial_temperature=initial_temperature,
             langevin=langevin,
+            drag=drag,
             job_name=job_name,
             rotation_matrix=rotation_matrix
         )
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -415,7 +415,6 @@ class LammpsControl(GenericParameters):
         tloop=None,
         initial_temperature=None,
         langevin=False,
-        drag=None,
         delta_temp=None,
         delta_press=None,
         job_name="",
@@ -455,9 +454,6 @@ class LammpsControl(GenericParameters):
                                                be used). If any other number is given, this value is going to be used
                                                for the initial temperature.
             langevin (bool): (True or False) Activate Langevin dynamics
-            drag (float): Drag to apply. Note: Using this flag will destroy your thermodynamics and you will no longer
-                acheive the NVT/NPT/NPH/NVE ensemble averages you would otherwise expect. (Default is None, no
-                additional drag (Langevin drag still applied in case of a Langevin thermostat).)
             delta_temp (float): Thermostat timescale, but in your Lammps time units, whatever those are. (DEPRECATED.)
             delta_press (float): Barostat timescale, but in your Lammps time units, whatever those are. (DEPRECATED.)
             job_name (str): Job name of the job to generate a unique random seed.
@@ -578,9 +574,6 @@ class LammpsControl(GenericParameters):
         if tloop is not None:
             fix_ensemble_str += " tloop " + str(tloop)
 
-        if drag is not None:
-            fix_ensemble_str += " drag " + str(drag)
-
         self.remove_keys(["minimize"])
         self.modify(
             fix___ensemble=fix_ensemble_str,
@@ -626,7 +619,6 @@ class LammpsControl(GenericParameters):
         seed=None,
         initial_temperature=None,
         langevin=False,
-        drag=None,
         job_name="",
         rotation_matrix=None
     ):
@@ -680,7 +672,6 @@ class LammpsControl(GenericParameters):
             tloop=None,
             initial_temperature=initial_temperature,
             langevin=langevin,
-            drag=drag,
             job_name=job_name,
             rotation_matrix=rotation_matrix
         )
--- a/pyiron/project.py
+++ b/pyiron/project.py
@@ -449,7 +449,7 @@ class Project(ProjectCore):
             cubic=False,
     ):
         """
-        Deprecated as of v.0.3.12, please use `Project.structure.create_ase_bulk`.
+        Deprecated as of v.0.3, please use `Project.create.structure.ase_bulk`.
 
         Creating bulk systems using ASE bulk module. Crystal structure and lattice constant(s) will be guessed if not
         provided.
@@ -469,7 +469,7 @@ class Project(ProjectCore):
             pyiron.atomistics.structure.atoms.Atoms: Required bulk structure
         """
         warnings.warn(
-            "Project.create_ase_bulk is deprecated as of v0.3.12. Please use Project.structure.create_ase_bulk.",
+            "Project.create_ase_bulk is deprecated as of v0.3. Please use Project.create.structure.ase_bulk.",
             DeprecationWarning
         )
         return self.create.structure.ase_bulk(name=name, crystalstructure=crystalstructure, a=a, c=c,
@@ -477,7 +477,7 @@ class Project(ProjectCore):
 
     def create_structure(self, element, bravais_basis, lattice_constant):
         """
-        Deprecated as of v.0.3.12, please use `Project.structure.create_structure`.
+        Deprecated as of v.0.3, please use `Project.create.structure.structure`.
 
         Create a crystal structure using pyiron's native crystal structure generator
 
@@ -491,7 +491,7 @@ class Project(ProjectCore):
 
         """
         warnings.warn(
-            "Project.create_structure is deprecated as of v0.3.12. Please use Project.structure.create_structure.",
+            "Project.create_structure is deprecated as of v0.3. Please use Project.create.structure.structure.",
             DeprecationWarning
         )
         return self.create.structure.structure(element=element, bravais_basis=bravais_basis,
@@ -501,7 +501,7 @@ class Project(ProjectCore):
             self, element, surface_type, size=(1, 1, 1), vacuum=1.0, center=False, pbc=None, **kwargs
     ):
         """
-        Deprecated as of v.0.3.12, please use `Project.structure.create_surface`.
+        Deprecated as of v.0.3, please use `Project.create.structure.surface`.
 
         Generate a surface based on the ase.build.surface module.
 
@@ -522,7 +522,7 @@ class Project(ProjectCore):
 
         """
         warnings.warn(
-            "Project.create_surface is deprecated as of v0.3.12. Please use Project.structure.create_surface.",
+            "Project.create_surface is deprecated as of v0.3. Please use Project.create.structure.surface.",
             DeprecationWarning
         )
         return self.create.structure.surface(element=element, surface_type=surface_type, size=size,
@@ -552,7 +552,7 @@ class Project(ProjectCore):
             **qwargs
     ):
         """
-        Deprecated as of v.0.3.12, please use `Project.structure.create_atoms`.
+        Deprecated as of v.0.3, please use `Project.create.structure.atoms`.
 
         Creates a atomistics.structure.atoms.Atoms instance.
 
@@ -582,7 +582,7 @@ class Project(ProjectCore):
             pyiron.atomistics.structure.atoms.Atoms: The required structure instance
         """
         warnings.warn(
-            "Project.create_atoms is deprecated as of v0.3.12. Please use Project.structure.create_atoms.",
+            "Project.create_atoms is deprecated as of v0.3. Please use Project.create.structure.atoms.",
             DeprecationWarning
         )
         return self.create.structure.atoms(
@@ -610,7 +610,7 @@ class Project(ProjectCore):
 
     def create_element(self, parent_element, new_element_name=None, spin=None, potential_file=None):
         """
-        Deprecated as of v.0.3.12, please use `Project.structure.create_element`.
+        Deprecated as of v.0.3, please use `Project.create.structure.element`.
 
         Args:
             parent_element (str, int): The parent element eq. "N", "O", "Mg" etc.
@@ -622,7 +622,7 @@ class Project(ProjectCore):
             atomistics.structure.periodic_table.ChemicalElement instance
         """
         warnings.warn(
-            "Project.create_element is deprecated as of v0.3.12. Please use Project.structure.create_element.",
+            "Project.create_element is deprecated as of v0.3. Please use Project.create.structure.element.",
             DeprecationWarning
         )
         return self.create.structure.element(
--- a/pyiron/project.py
+++ b/pyiron/project.py
@@ -449,8 +449,6 @@ class Project(ProjectCore):
             cubic=False,
     ):
         """
-        Deprecated as of v.0.3, please use `Project.create.structure.ase_bulk`.
-
         Creating bulk systems using ASE bulk module. Crystal structure and lattice constant(s) will be guessed if not
         provided.
 
@@ -468,17 +466,15 @@ class Project(ProjectCore):
 
             pyiron.atomistics.structure.atoms.Atoms: Required bulk structure
         """
-        warnings.warn(
-            "Project.create_ase_bulk is deprecated as of v0.3. Please use Project.create.structure.ase_bulk.",
-            DeprecationWarning
-        )
+        # warnings.warn(
+        #     "Project.create_ase_bulk is deprecated as of v0.3. Please use Project.create.structure.ase_bulk.",
+        #     DeprecationWarning
+        # )
         return self.create.structure.ase_bulk(name=name, crystalstructure=crystalstructure, a=a, c=c,
                                               covera=covera, u=u, orthorhombic=orthorhombic, cubic=cubic)
 
     def create_structure(self, element, bravais_basis, lattice_constant):
         """
-        Deprecated as of v.0.3, please use `Project.create.structure.structure`.
-
         Create a crystal structure using pyiron's native crystal structure generator
 
         Args:
@@ -490,10 +486,10 @@ class Project(ProjectCore):
             pyiron.atomistics.structure.atoms.Atoms: The required crystal structure
 
         """
-        warnings.warn(
-            "Project.create_structure is deprecated as of v0.3. Please use Project.create.structure.structure.",
-            DeprecationWarning
-        )
+        # warnings.warn(
+        #     "Project.create_structure is deprecated as of v0.3. Please use Project.create.structure.structure.",
+        #     DeprecationWarning
+        # )
         return self.create.structure.structure(element=element, bravais_basis=bravais_basis,
                                                lattice_constant=lattice_constant)
 
@@ -501,8 +497,6 @@ class Project(ProjectCore):
             self, element, surface_type, size=(1, 1, 1), vacuum=1.0, center=False, pbc=None, **kwargs
     ):
         """
-        Deprecated as of v.0.3, please use `Project.create.structure.surface`.
-
         Generate a surface based on the ase.build.surface module.
 
         Args:
@@ -521,10 +515,10 @@ class Project(ProjectCore):
             pyiron.atomistics.structure.atoms.Atoms instance: Required surface
 
         """
-        warnings.warn(
-            "Project.create_surface is deprecated as of v0.3. Please use Project.create.structure.surface.",
-            DeprecationWarning
-        )
+        # warnings.warn(
+        #     "Project.create_surface is deprecated as of v0.3. Please use Project.create.structure.surface.",
+        #     DeprecationWarning
+        # )
         return self.create.structure.surface(element=element, surface_type=surface_type, size=size,
                                              vacuum=vacuum, center=center, pbc=pbc, **kwargs)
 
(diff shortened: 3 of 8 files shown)


jan-janssen:
I am a bit critical on this one as `pr.structure` is too similar to `job.structure`. An alternative syntax would be `pr.create.structure.bulk("Al")` Then we would have one creator which could also be used to create jobs like `pr.create.job.Lammps(job_name="lmp")`.

liamhuber:
> I am a bit critical on this one as `pr.structure` is too similar to `job.structure`. 

Absolutely! This is an aspect I really want to take a minute to think about.

> An alternative syntax would be `pr.create.structure.bulk("Al")` Then we would have one creator which could also be used to create jobs like `pr.create.job.Lammps(job_name="lmp")`.

I like this in general, but let's iterate it a bit more. Issues that come to mind:

- In this PR I won't implement such a setup for everyone. However, I'd be happy to make a (mostly empty) intermediate object(s) that we can populate later. I don't see a problem with having an intermediate transition period where we use `pr.create.structure` but `pr.create_job` because the `job` stuff is just not finished yet.
- Even here we need to be careful with our language choice. E.g. `pr.create.structure.load` feels a bit funny, with "create" and "load" both appearing in the same pattern.
- Finally, *this* is a good moment to think whether we want to move the language `structure`->`sample`. I know there was nothing but negative feedback over on the discussion for this, but further discussions with Erik Bitzek's group, and thinking about pyiron moving towards also treating FEM, etc... I'm still not convinced we want to stick with "structure" forever.

pmrv:
I have a modest issue with the fact that `StructureGenerator` is a class, but all its members are `staticmethod` and all further extensions I can think of (e.g. dislocation or grain boundary generators) wouldn't change that.  I think we had a similar discussion already with `Visualization`, but I feel that this couples `Project` and structures too closely together.  Is there a reason that the structure generation functions live on the project and not as function in a separate module (say `pyiron.atomistic.structure....`)?  I can only see that it saves typing for users (which is a very valid reason), but from a programming view point I am less convinced.

liamhuber:
> I have a modest issue with the fact that `StructureGenerator` is a class, but all its members are `staticmethod` and all further extensions I can think of (e.g. dislocation or grain boundary generators) wouldn't change that. 

Yeah, I had actually thought first to simply do this:

```python
from pyiron.atomistics.structure import generator

class Project(...):
    def __init__(...):
        ...
        self.structure = generator
```

But I couldn't find anything on google about the best practices of simply assigning an entire module to a class attribute, and in my own tests it had weird side effects like you could do `pr.structure.np.arange(5)`. 

I also am not keen on a class that's all methods and no data, but I couldn't think of a better architecture.

What's nice so far is that this is all under-the-hood, so even if we stick with the class structure now and later find a way to make it only a module, it shouldn't make any difference on the user end so it's an easy change 😄 

> I think we had a similar discussion already with `Visualization`, but I feel that this couples `Project` and structures too closely together. Is there a reason that the structure generation functions live on the project and not as function in a separate module (say `pyiron.atomistic.structure....`)? I can only see that it saves typing for users (which is a very valid reason), but from a programming view point I am less convinced.

I don't know where our "zen of pyiron" went, but one of the ideas was that *everything* comes through `Project` s.t. you need no other imports. We bend this already with `import pyiron_contrib` etc, but even that activates stuff in `Project` under the hood.

Other than that I share your perspective, but I think the Zen of Pyiron wins.

liamhuber:
> Is there a reason that the structure generation functions live on the project and not as function in a separate module (say `pyiron.atomistic.structure....`)? 

Actually, I just realized they are living in such a module from a maintenance and development perspective, we just use the class interface so that `Project` can bind to this module easier. This is satisfactory, no? (I mean, other than the 'it's a class' issue discussed earlier.)

liamhuber:
Ok, I abstracted out a `create` attribute for `Project`. Ideally we'll move it up to pyiron_base, but since all it has in it right now is `self.structure = StructureGenerator` we can do the move in another PR.

Current behaviour looks like this:
```python
struct = pr.create.structure.ase_bulk('Al', cubic=True) 
native_struct = pr.create.structure.structure('Fe')  # A little redundant?
slab = pr.create.structure.surface(*some_args, **and_kwargs_i_dont_know)
from_file = pr.create.structure.ase_read('my_lammps_dump,  format='lammps-data', style='atomic')
```

Obviously there's some QoL stuff that can be added, e.g. wrapping up `ase_read` for Lammps dumps, POSCARS, etc. I don't mind the foundation here though.

I suggest that everyone already tagged (@jan-janssen @sudarsan-surendralal @pmrv) iterates until we're all happy, then we shelve it and introduce it to the group at the Friday meeting. My thought is to get most of the implementation dirt out of the way, then just ask (a) whether we actually want to do this, and (b) is the nomenclature good?

jan-janssen (dismissed):
At the current stage I am against merging this as I do not see it being beneficial with the the amount of changes required in the documentation if we want to release this feature in the next version. Instead I would suggest moving it to `pyiron_contrib` and then have multiple iterations there before we merge it into `pyiron`.

liamhuber:
> At the current stage I am against merging this as I do not see it being beneficial with the the amount of changes required in the documentation if we want to release this feature in the next version. Instead I would suggest moving it to `pyiron_contrib` and then have multiple iterations there before we merge it into `pyiron`.

I am ok with having this development in pyiron_contrib, but I would need guidance on how to do the implementation. Currently all the contrib work is in the form of new jobs which get automatically parsed -- we never instantiate a contrib `Project` or `Atoms`. 

The way that is immediately obvious to me is to copy project.py, atomistics/structure/generator.py, and (coming soon from some changes suggested by Joerg at the tech dev meeting) atomistics/structure/atoms.py over to pyiron_contrib. I'm still not sure how project instantiation should work, but as long as pyiron_contrib is ahead of pyiron in python path I guess the contrib-style `Atoms` object would be the one getting created. But I have serious reservations about trying to maintain two copies of `Project`, `Atoms`, and `generator`. 

Also, there is still full backwards compatibility here, so none of the documentation notebooks should actually break, they just become outdated. Obviously not ideal (hence me being fine with moving this to contrib) but also not a deal breaker for me.

jan-janssen:
As this change mainly effects the project object my idea was creating a:
```
from pyiron_contrib import Project
```
which is derived from the current pyiron Project class and adds the additional functionality. 

Then I took another look and noticed that a universal creator is most likely relevant for all pyiron developments, so maybe instead of placing it in pyiron_contrib it should already be part of pyiron_base. Sorry for the confusing directions. 

I created a job creator for pyiron base in: 
https://github.com/pyiron/pyiron_base/pull/81
and then based on this I basically reimplemented your changes in: 
https://github.com/pyiron/pyiron/pull/1200

For now I only add new code rather than overwriting everything to have an alternative interface. Again the pyiron pull request could also be implemented inside pyiron_contrib .

jan-janssen:
One more thing, regarding the Deprecation Warnings we should only change the interface when moving from 0.X to 0.Y and not when moving from 0.X.Y to 0.X.Z - https://en.wikipedia.org/wiki/Software_versioning#Degree_of_compatibility

liamhuber:
> As this change mainly effects the project object my idea was creating a:
> 
> ```
> from pyiron_contrib import Project
> ```
> 
> which is derived from the current pyiron Project class and adds the additional functionality.

Yes, with the exception that the subclassed atomistic project would need to be in the atomistics directory, i.e. `from pyiron_contrib.atomistics import Project`, as this subclassed project changes only structures which are a unique property of the atomistics domain.

> Then I took another look and noticed that a universal creator is most likely relevant for all pyiron developments, so maybe instead of placing it in pyiron_contrib it should already be part of pyiron_base. Sorry for the confusing directions.

Indeed, this is exactly what I had in mind:

>> Ok, I abstracted out a create attribute for Project. Ideally we'll move it up to pyiron_base, but since all it has in it right now is self.structure = StructureGenerator we can do the move in another PR.

> I created a job creator for pyiron base in:
> [pyiron/pyiron_base#81](https://github.com/pyiron/pyiron_base/pull/81)

Very cool. I will go review there properly, but I took a look already and it's nice.

> and then based on this I basically reimplemented your changes in:
> #1200

This, unfortunately, I find not so nice.

Subclassing `Creator` is fine, where it (atm) just adds the structure factory. I'm not a fan of jamming all the factories into one file as is implied, and once there's just the `Creator` object it should fit right in `project.py` and not need  a special `factory.py` file.

I think factory files are more logically stored next to the objects they're creating, i.e. where the `generator.py` file is now. I'd be happy to rename `generator.py` to `factory.py`.

I am 100% against adding an interface layer that exclusively re-implements code existing in another module. This is one more thing to keep updated, one more place to repeat docstrings and long lists of args, and generally bloats the code and creates maintenance headaches. Using `pyiron.atomstics.generic.factory.StructureCreator` in #1200 seems strictly inferior to `pyiron.atomistics.structure.generator.StructureCreator` in this PR.

What I *do* like is the use of the name `factory` instead of `generator`. I was naively following the generator module name, but in fact generators are already a separate beast in python. I fully support switching our language over to "factories" for what is currently being built, including renaming `JobCreator` and `StructureCreator` to `*Factory`. Although for the `Creator` object I think sticking with creator is nice, since this is something that *houses* factories, and thus can reasonably have a different name.

> For now I only add new code rather than overwriting everything to have an alternative interface. Again the pyiron pull request could also be implemented inside pyiron_contrib .

...Why? All the changes here are under the hood and not accessible from the user-side `Project` import. All the existing methods all still exist, albeit with a deprecation warning. Only developer-side stuff was "overwritten" (i.e. moved from module level to class level), and this was done consistently including in the test suite. 

I don't see what the objection is to this PR. Other than adding the deprecation warnings it does what #1200 does but in a way that's easier to maintain.

liamhuber:
> One more thing, regarding the Deprecation Warnings we should only change the interface when moving from 0.X to 0.Y and not when moving from 0.X.Y to 0.X.Z - https://en.wikipedia.org/wiki/Software_versioning#Degree_of_compatibility

Sure thing, will do.

If you want, we can comment out the deprecation warnings in this PR so that the current documentation still looks "clean" when people implement it. Then after the version switch all we need to do is uncomment the warnings and update the version number in them.

liamhuber:
If it is really important to hide these changes from users we can always hide the new structure attribute (or easier, `create` over in base) behind an underscore. Then we can have the new architecture and release it to users as easily as a refactor:rename.

sudarsan-surendralal:
I like the original idea of what is proposed here. I also like the new `factory` module very much.

Just a suggestion: In principle, if we are relying on auto complete, we could use this to genreate any object. IMO, our syntax should be something like `pr.generate_structure.ase_bulk("Al")`, `pr.generate_structure.surface("Al", "fcc111", (4, 4, 4))`, `pr.generate_job.Lammps("lmp")` or even `pr.generate_object.ElectronicStructure()`. By separating the `generators`,  we reduce the size of the auto complete lookup which makes things much easier for users. `pr.create.structure.structure('Fe')` for example seems a bit long and not that nice to look at. Also, didn't jupyter at some time have a limit on the levels you can look up? We can overload the generators based on the packages the users use (jobs in `pyiron_base`, atomistic objects in `pyiron`, and experimental objects in `pyiron_experimental`). We could also have generators independent of the `Project` class. For example, in `__init__.py`, we could have something like

```
from pyiron.atomistics.structure.factory import StructureFactory
structure_generator = StructureFactory()
```

and then the users can call them in the notebook like:

```
from pyiron import structure_generator
struct = structure_genreator.ase_bulk("Al")
```

liamhuber:
> I like the original idea of what is proposed here. I also like the new `factory` module very much.

Glad to hear it! Thanks for all the feedback 😄 

> Just a suggestion: In principle, if we are relying on auto complete, we could use this to genreate any object. IMO, our syntax should be something like `pr.generate_structure.ase_bulk("Al")`, `pr.generate_structure.surface("Al", "fcc111", (4, 4, 4))`, `pr.generate_job.Lammps("lmp")` or even `pr.generate_object.ElectronicStructure()`. By separating the `generators`, we reduce the size of the auto complete lookup which makes things much easier for users. `pr.create.structure.structure('Fe')` for example seems a bit long and not that nice to look at. 

So actually, I'm specifically looking to avoid the `generate_structure`, `generate_job`, etc. setup. This is basically what we have now and I find the tab-completion lists extremely crowded.

It sounds like we all agree that separating the underlying architecture off into creation factories is a nice idea. Given that, it's my preference to have the interface reflect that archiceture. In this way, if we have separate structure and job factories, it would actually require re-inserting another interface layer to get *back* to `generate_structure`, `generate_job`...

Personally I am most in favour of how it is now, and Jan has added the `create` step already over in [pyiron_base](https://github.com/pyiron/pyiron_base/pull/81). However, I would be still be open to eliminating one layer, e.g. something like `pr.create.structure.ase_bulk` --> `pr.structures.ase_bulk`, `pr.create.job.Lammps` --> `pr.jobs.Lammps`. 

> Also, didn't jupyter at some time have a limit on the levels you can look up? 

Ayayay, I hope not! We definitely haven't hit that limit yet, as I've been using this branch actively and don't have problems, but if there really is a hard wall then we should be careful how many layers we add...

> We can overload the generators based on the packages the users use (jobs in `pyiron_base`, atomistic objects in `pyiron`, and experimental objects in `pyiron_experimental`). 

Yes! I think this is a really excellent idea. @jan-janssen has already moved in this direction (same pyiron_base PR linked above) in the sense that the job factory automatically updates its createable classes based on the class list controlled by what's been imported.

Right now the base `Project` has a `Creator` class that has a `JobCreator` attribute, then here the atomistic `Project` duplicates this by subclassing these base classes. Right now this sub-classed `Creator` only adds the `StructureFactory`. In principle, we should be able to take the exact trick applied to automatically update `JobCreator` and apply it to the `Creator` object, so that creation fields also depend on what pyiron packages have been imported.

That's just an implementation difference though, and to the user it will look the exact same. Also while I'm sure I could go figure out what `JobCreator` is doing and mimic it, it will probably be faster and more efficient if someone who already knows how that works implements this. Thus, I really like it but let's leave it for another PR.

> We could also have generators independent of the `Project` class. For example, in `__init__.py`, we could have something like
> 
> ```
> from pyiron.atomistics.structure.factory import StructureFactory
> structure_generator = StructureFactory()
> ```
> 
> and then the users can call them in the notebook like:
> 
> ```
> from pyiron import structure_generator
> struct = structure_genreator.ase_bulk("Al")
> ```

I guess this is already possible, and actually is done in some of the tests.

However, I think this behaviour should be discouraged as "un-pyironic". The major disadvantage is that it makes backwards compatibility very hard to maintain. As of this PR, `pr.create_ase_bulk` and `pr.create.structure.ase_bulk` both work without complaint and everyone's notebooks will carry on working. If we were to later rename `StructureFactory` to `StructureGenerator` it is trivially easy to make sure both those interfaces keep working, but if people are explicitly importing `StructureFactory` in their notebooks then all of a sudden they'll start crashing.

It comes a little bit back to a discussion last Friday of how/what we want to make private. I currently think we should follow regular python standards for what gets hidden by an `_`, but simply encourage regular users to work within the Zen of Pyiron and access everything through `Project`.

jan-janssen:
> Personally I am most in favour of how it is now, and Jan has added the `create` step already over in [pyiron_base](https://github.com/pyiron/pyiron_base/pull/81). However, I would be still be open to eliminating one layer, e.g. something like `pr.create.structure.ase_bulk` --> `pr.structures.ase_bulk`, `pr.create.job.Lammps` --> `pr.jobs.Lammps`.

I already mentioned this above, but I find `pr.structure.ase_bulk` confusing as `job.structure` has a different usage and currently we also have `job.create_job()` so I would start by introducing just `pr.create.*` and `job.create.*`. 
 
> > Also, didn't jupyter at some time have a limit on the levels you can look up?
> 
> Ayayay, I hope not! We definitely haven't hit that limit yet, as I've been using this branch actively and don't have problems, but if there really is a hard wall then we should be careful how many layers we add...

We currently fix jedi - the package used for the autocompletion - when importing pyiron_base: 
https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/__init__.py#L42

> That's just an implementation difference though, and to the user it will look the exact same. Also while I'm sure I could go figure out what `JobCreator` is doing and mimic it, it will probably be faster and more efficient if someone who already knows how that works implements this. Thus, I really like it but let's leave it for another PR.

The magic is basically the combination of a dictionary and a singleton, we can create similar implementation for the StructureCreator but I agree with Liam that should most likely be addressed in a separate pull request.

pyiron/lammps/control.py
```diff
@@ -415,6 +415,7 @@ def calc_md(
         tloop=None,
         initial_temperature=None,
         langevin=False,
+        drag=None,
```
jan-janssen, line 418: Can we move these changes to a separate pull request? they have nothing todo with the structure creator.

pyiron/project.py
```diff
@@ -113,6 +114,11 @@ def __init__(self, path="", user=None, sql_query=None, default_working_directory
         )
         self.job_type = JobTypeChoice()
         self.object_type = ObjectTypeChoice()
+        self._create = _Creator()
+
+    @property
+    def create(self):
+        return self._create
```
jan-janssen, line 121:
```suggestion
```
These lines are no longer necessary when merging with the current master branch, as the property is already implemented in `pyiron_base`

pyiron/project.py
```diff
@@ -18,7 +19,7 @@
 from pyiron.vasp.potential import VaspPotential
 import pyiron.atomistics.structure.pyironase as ase
 from pyiron.atomistics.structure.atoms import Atoms
-from pyiron.atomistics.structure.generator import create_surface, create_ase_bulk, create_structure
+from pyiron.atomistics.structure.factory import StructureFactory
```
jan-janssen, line 22:
```suggestion
from pyiron.atomistics.structure.factory import StructureFactory
from pyiron_base import Creator as CreatorBase
```

pyiron/project.py
```diff
@@ -573,58 +611,23 @@ def create_element(
...
+            new_element_name=new_element_name,
+            spin=spin,
+            potential_file=potential_file
+        )
 
-        Returns:
 
-        """
-        ProjectGUI(self)
+class _Creator:
```
jan-janssen, line 626:
```suggestion
class _Creator(CreatorBase):
```

tests/atomistics/structure/test_atoms.py
```diff
@@ -38,6 +38,7 @@ def setUpClass(cls):
         C = Atom("C").element
         cls.C3 = Atoms([C, C, C], positions=[[0, 0, 0], [0, 0, 2], [0, 2, 0]])
         cls.C2 = Atoms(2 * [Atom("C")])
+        cls.g = StructureFactory()
```
jan-janssen, line 41: Can we use more readable variables? In particular in tests that would be useful.

pyiron/lammps/control.py
liamhuber: Ack! Of course. This was just a git error from being on the wrong branch at some point.

pyiron/project.py
liamhuber: Yes, now that it's on pyiron base I will use that.

tests/atomistics/structure/test_atoms.py
liamhuber: Yes, sure. This used to be at least a little more sensible when this thing was a "Generator", but I'll make it more verbose.

pyiron/project.py
liamhuber: @jan-janssen my understanding is that this allows (a) `from pyiron import CreatorBase` and (b) the use of `CreatorBase` freely throughout the pyiron/pyiron module. Is this correct and complete?

If so, I don't understand the purpose here. (a) doesn't seem like something we want people to do, and (b) `CreatorBase` is used only in `pyiron.project.Project` so we don't reduce the number of import lines. Please expand if I'm missing some benefit.

jan-janssen: I guess you now did the same in line 11 so we do not need it here.

pyiron/atomistics/structure/factory.py
```diff
@@ -0,0 +1,333 @@
...
+__maintainer__ = "Sudarsan Surendralal"
+__email__ = "surendralal@mpie.de"
+__status__ = "production"
+__date__ = "May 1, 2020"
+
+s = Settings()
+
+
+class StructureFactory:
+    def ase_read(self, *args, **kwargs):
```
jan-janssen, line 51:
```suggestion
    @staticmethod
    def ase_read(*args, **kwargs):
```

pyiron/atomistics/structure/factory.py
```diff
@@ -0,0 +1,333 @@
...
+                surface = surface_type(symbol=element, size=size, **kwargs)
+                z_max = np.max(surface.positions[:, 2])
+                surface.cell[2, 2] = z_max + vacuum
+            surface.pbc = pbc
+            return ase_to_pyiron(surface)
+        else:
+            return None
+
+    @staticmethod
+    def hkl_surface(lattice, hkl, layers, vacuum=1.0, center=False, pbc=True):
```
jan-janssen, line 126:
```suggestion
    def surface_hkl(lattice, hkl, layers, vacuum=1.0, center=False, pbc=True):
```

pyiron/atomistics/structure/factory.py
```diff
@@ -0,0 +1,333 @@
...
+        surface = ase_surf(lattice, hkl, layers)
+        z_max = np.max(surface.positions[:, 2])
+        surface.cell[2, 2] = z_max + vacuum
+        if center:
+            surface.positions += 0.5 * surface.cell[2] - [0, 0, z_max/2]
+        surface.pbc = pbc
+        return ase_to_pyiron(surface)
+
+    @staticmethod
+    def structure(element, bravais_basis, lattice_constant):
```
jan-janssen, line 154:
```suggestion
    def crystal(element, bravais_basis, lattice_constant):
```
The CrystalStructure class is the original pyiron structure generator but I have the feeling most of us use the ASE structure generator these days so I would remove this function. But maybe we should move that discussion to a separate pull request.

pyiron/project.py
```diff
@@ -396,31 +397,51 @@ def _calculation_validation(self, path, files_available, rel_path=None, copy_raw
...
     def create_ase_bulk(
-        name,
-        crystalstructure=None,
-        a=None,
-        c=None,
-        covera=None,
-        u=None,
-        orthorhombic=False,
-        cubic=False,
+            self,
```
jan-janssen, line 436:
```suggestion
```
We do not need self in static functions

jan-janssen: My mistake I just realised it is no longer a static function.

pyiron/project.py
```diff
@@ -440,31 +461,84 @@ def create_ase_bulk(
...
+
+        Returns:
+            pyiron.atomistics.structure.atoms.Atoms: The required crystal structure
+
+        """
+        # warnings.warn(
+        #     "Project.create_structure is deprecated as of v0.3. Please use Project.create.structure.structure.",
+        #     DeprecationWarning
+        # )
+        return self.create.structure.structure(element=element, bravais_basis=bravais_basis,
```
jan-janssen, line 488:
```suggestion
        return self.create.structure.crystal(element=element, bravais_basis=bravais_basis,
```
see comment above

tests/atomistics/structure/test_atoms.py
```diff
@@ -1483,19 +1484,19 @@ def test__setitem__(self):
...
-        surface = create_surface("Al", "fcc111", size=(4, 4, 4), vacuum=10)
+        surface = self.struct_factory.surface("Al", "fcc111", size=(4, 4, 4), vacuum=10)
         self.assertTrue(all(surface.pbc))
-        surface = create_surface("Al", "fcc111", size=(4, 4, 4), vacuum=10, pbc=[True, True, False])
+        surface = self.struct_factory.surface("Al", "fcc111", size=(4, 4, 4), vacuum=10, pbc=[True, True, False])
         self.assertTrue(all(surface.pbc[0:2]))
         self.assertFalse(surface.pbc[2])
         self.assertIsInstance(surface, Atoms)
-        hkl_surface = create_hkl_surface(Al_bulk, [10, 8, 7], layers=20, vacuum=10)
+        hkl_surface = self.struct_factory.hkl_surface(Al_bulk, [10, 8, 7], layers=20, vacuum=10)
```
jan-janssen, line 1496:
```suggestion 
        hkl_surface = self.struct_factory.surface_hkl(Al_bulk, [10, 8, 7], layers=20, vacuum=10)
```

tests/atomistics/structure/test_atoms.py
```diff
@@ -1483,19 +1484,19 @@ def test__setitem__(self):
...
         self.assertIsInstance(hkl_surface, Atoms)
         self.assertTrue(all(hkl_surface.pbc))
-        hkl_surface_center = create_hkl_surface(
+        hkl_surface_center = self.struct_factory.hkl_surface(
```
jan-janssen, line 1499:
```suggestion
        hkl_surface_center = self.struct_factory.surface_hkl(
```

jan-janssen:
guess we are getting closer now, @liamhuber can you prepare the code to pass the tests then we can discuss it on Friday and hopefully merge it on the weekend to have a new release with the pyscal integration and the new creator.

pyiron/atomistics/structure/factory.py
liamhuber: @jan-janssen I'm ok with this suggestion for this PR, but actually I was toying around with the idea of modifying `StructureFactory.__call__` so that we can simply say `pr.create.structure('Al', 'fcc', 4.04)` while still being able to say `pr.create.structure.surface(...)` etc.

In fact, if I were being *really* bold, I would want `pr.create.structure(...)` to be equivalent to our current `pr.create_ase_bulk(...)`, and `pr.create.bulk(...)` to be equivalent to our current `pr.create_structure`.

Thoughts? I agree it can be a separate PR though, and in the meantime will commit your suggestions.

jan-janssen: I like the idea with the `__call__` I was not even aware of this and I am happy to match it with `pr.create.bulk()` but I would still prefer to remove the old `Crystalstructure` class completely unless somebody else is using it regularly. Again this should be a separate pull request.

liamhuber:
@jan-janssen I've got to go pick up my son, but I plan to incorporate your renaming suggestions en masse. Will work more on this later.

pyiron/project.py
```diff
@@ -165,16 +168,7 @@ def create_job(self, job_type, job_name, delete_existing_job=False):
...
-            job_type,
-            project=ProjectHDFio(project=self.copy(), file_name=job_name),
-            job_name=job_name,
-            job_class_dict=self.job_type.job_class_dict,
-            delete_existing_job=delete_existing_job,
-        )
-        if self.user is not None:
-            job.user = self.user
-        return job
+        return super().create_job(job_type=job_type, job_name=job_name, delete_existing_job=delete_existing_job)
```
jan-janssen, line 171: Does this work? I agree it should work once it is mapped to `pr.create` but currently I have the feeling this might break the functionality so I would prefer to move it to a separate pull request.

pyiron/atomistics/structure/factory.py
liamhuber: Agreed on both counts.

pyiron/project.py
liamhuber: You bet, why wouldn't it? The code is line for line the exact same, and `pyiron.project` is anyhow just importing `JobClass` from pyiron_base.

Sure, I have no problem moving to another PR. There's some related changes I want to make anyhow now that we've got all this creator stuff into base.

pyiron/atomistics/structure/factory.py
liamhuber: I know it looks like this one can be static, but actually the line below it (`ase_read.__doc__ = read.__doc__`) only works when it's got `self`. I can kind of guess why, but it surprised me.

jan-janssen:
The failed tests seem to be related to https://github.com/pyiron/pyiron/pull/1202

liamhuber:
> The failed tests seem to be related to #1202

Yeah, I was just digging into this now. I'd like to get that merged ASAP so we can actually see easily whether this one is breaking anything...

liamhuber:
The check failure is just some dumb style complaints IMO. I'm content to ignore them.

jan-janssen (approved):
Looks good to me

PR #1202, opened 2020-11-18
Update to pyscal to 2.10.3

jan-janssen:
To fix Voronoi for triclinic cells https://github.com/srmnitc/pyscal/pull/86


jan-janssen:
@srmnitc After the update the tests start to fail:

```
1586
FAIL: test_analyse_pyscal_centro_symmetry (atomistics.structure.test_pyscal.Testpyscalatoms)
1587
1588
1589
  File "/home/runner/work/pyiron/pyiron/tests/atomistics/structure/test_pyscal.py", line 95, in test_analyse_pyscal_centro_symmetry
1590
    self.assertTrue(all([np.isclose(v, 22.113675) for v in self.si_dia.analyse_pyscal_centro_symmetry()]))
1591
AssertionError: False is not true
1592

1593
```

tests/atomistics/structure/test_pyscal.py
```diff
@@ -92,7 +92,11 @@ def test_analyse_pyscal_centro_symmetry(self):
         self.assertTrue(all([np.isclose(v, 0.0) for v in self.al_fcc.analyse_pyscal_centro_symmetry()]))
         self.assertTrue(all([np.isclose(v, 6.177675) for v in self.fe_bcc.analyse_pyscal_centro_symmetry()]))
         self.assertTrue(all([np.isclose(v, 8.7025) for v in self.ti_hcp.analyse_pyscal_centro_symmetry()]))
-        self.assertTrue(all([np.isclose(v, 22.113675) for v in self.si_dia.analyse_pyscal_centro_symmetry()]))
+        print(self.si_dia.analyse_pyscal_centro_symmetry())
+        self.assertTrue(all([
```
liamhuber, line 96: I found this hard to read. So far I've got this as an improvement:
```python
self.asserTrue(all([
    any(np.isclose(centrosym, [22.113675, 16.58525625, 11.0568375])) 
    for centrosym in self.si_dia.analyse_pyscal_centro_symmetry()
])
```

But honestly that's not yet much improvement.

liamhuber:
```suggestion
        self.assertTrue(np.allclose
            np.unique(self.si_dia.analyse_pyscal_centro_symmetry()), 
            [11.0568375, 16.58525625, 22.113675]
        ))
```

Also works, but I'm actually more comfortable with forcing the sort.

I find these easier to understand the intent, but it might just be perspective

liamhuber: Also, I'm new to these suggestions and I don't see how to recommend getting rid of the lines below this as well...

liamhuber (approved):
I found the test a little hard to read and propose some changes. I would be content to have it merged as is if you don't like them.

tests/atomistics/structure/test_pyscal.py
jan-janssen: > Also, I'm new to these suggestions and I don't see how to recommend getting rid of the lines below this as well...

You click the plus and then move your cursor down

jan-janssen:
The more important issue is that the results do not agree with Ovito ... https://github.com/jan-janssen/ovito-vs-pyscal/blob/master/ovito-vs-pyscal-centro-diamond.ipynb

liamhuber:
> The more important issue is that the results do not agree with Ovito ... https://github.com/jan-janssen/ovito-vs-pyscal/blob/master/ovito-vs-pyscal-centro-diamond.ipynb

Aha, then my dreams of an extremely fast merge are unlikely ;) I hope anyhow my readability suggestion is helpful. And thanks for the dodge to this issue over in #1199

srmnitc:
https://github.com/srmnitc/pyscal/pull/89 will fix the diamond issue.

.ci_support/environment.yml
```diff
@@ -17,7 +17,7 @@ dependencies:
   - phonopy >=2.4.2
   - pymatgen >=2020.9.14
   - pyiron_base >=0.1.29
-  - pyscal >=2.10.1
+  - pyscal >=2.10.3
```
srmnitc, line 20: The new release version with the centrosymmetry fix will be 2.10.4.

srmnitc (approved):

srmnitc:
> The more important issue is that the results do not agree with Ovito ... https://github.com/jan-janssen/ovito-vs-pyscal/blob/master/ovito-vs-pyscal-centro-diamond.ipynb

Now resolved in the [new release](https://github.com/srmnitc/pyscal/releases/tag/2.10.4). Also the `calculate_centrosymmetry` now returns a list of cs values so you can just do-
```
cs = sys.calculate_centrosymmetry(nmax=num_neighbors)
```
instead of
```
atoms = sys.atoms
cs = [atom.centrosymmetry for atom in atoms]
```

srmnitc:
@jan-janssen Probably should have spotted the first time, but I think the test cases for centrosymmetry should use 12 neighbors for fcc and 8 for bcc. If 8 is chosen for bcc, then the cs values should be 0. The problem with 12 neighbors for bcc is that it is not clear which neighbors (other than the first shell) will be selected. For diamond one cannot get 0 value for cs, but to be consistent it should be tested for 4 neighbors.

jan-janssen:
@srmnitc thx for the explanation

jan-janssen:
This is ready to be merged once the tests passed.

PR #1204, opened 2020-11-19
Bugfix get neighbors

samwaseda:
Relatively minor issue.


samwaseda:
@edmen-dev @jan-janssen Since the new features are strongly non-canonical, I don't create a notebook but I filled all functions with very detailed explanations. So if you want to look for more info, just browse through items via auto-complete. But one way or other, the main attributes are the same: distances, indices and vectors.

PR #1207, opened 2020-11-20
Inherit from PyironFactory

liamhuber:
Dependent on [pyiron/pyiron_base#83](https://github.com/pyiron/pyiron_base/pull/83)

### DIFF
--- a/pyiron/atomistics/structure/factory.py
+++ b/pyiron/atomistics/structure/factory.py
@@ -30,7 +30,7 @@ import numpy as np
 from pyiron.atomistics.structure.pyironase import publication as publication_ase
 from pyiron.atomistics.structure.atoms import CrystalStructure, ase_to_pyiron, Atoms
 from pyiron.atomistics.structure.periodic_table import PeriodicTable
-from pyiron_base import Settings
+from pyiron_base import Settings, PyironFactory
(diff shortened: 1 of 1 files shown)


liamhuber:
Now that the pyiton_base dependency is on tag `0.1.30`, this should be ready for review.

liamhuber:
@jan-janssen I want to re-run the tests but don't want to add any silly empty commits to do it [(as in SO results I found)](https://stackoverflow.com/questions/52408592/how-to-relaunch-github-check-without-pushing-new-commits) -- do we have a sneaky way to do this? I guess most/all the tests should pass now that the dependencies are updated.

jan-janssen:
> @jan-janssen I want to re-run the tests but don't want to add any silly empty commits to do it [(as in SO results I found)](https://stackoverflow.com/questions/52408592/how-to-relaunch-github-check-without-pushing-new-commits) -- do we have a sneaky way to do this? I guess most/all the tests should pass now that the dependencies are updated.

I can rerun the tests by clicking on the failed test and then there is a run again button (not sure if it is available for everybody) but the empty commit is definitely the easiest way.

jan-janssen (approved):
Looks good to me

Issue #537, opened 2020-02-29
Does it make sense to multiply it by 2*pi?


samwaseda:
The name `kpoints_per_angstrom` in `get_k_mesh_by_cell` makes me think that this is the number of k-points per angstrom, but in reality it is 1/angstrom and multiplied by 2*pi. Is it really intuitive? Besides, from my point of view it makes more sense to have `kpoints_per_angstrom` as the first argument and `cell` after that, since I'm usually interested to know how many points I need for the **current** cell and not what it looks like for an unrelated structure.

https://github.com/pyiron/pyiron/blob/7fc724ba91964323050a1d0c5603677df0a54801/pyiron/dft/job/generic.py#L83

P.S.
I just realised that I had changed `kpoints_per_in_angstrom` to `kpoints_per_angstrom` around a year ago without knowing that `in` stood for "inverse". I'd be very much interested to know whether anyone was actually using this command.

samwaseda:
Maybe I should say clearly how I would use it: I'd like to get let's say something like 24 k-points per lattice in each direction. Then I'm fine with any of the following options:
```
job.get_k_mesh_by_cell(24)
job.get_k_mesh_by_cell(24/lattice_constant)
job.get_k_mesh_by_cell(24*lattice_constant)
```
But **NOT** with
```
get_k_mesh_by_cell(None, 24/lattice_constant/2/np.pi)
get_k_mesh_by_cell(None, 24*lattice_constant/2/np.pi)
```

jan-janssen:
If I remember correctly @yury-lysogorskiy implemented this functionality for his high throughput calculation, so maybe he can explain the usage.

liamhuber:
While we're talking about this, I'm pretty sure the variable name needs to be `kpoints_per_reciprocal_angstrom`

jan-janssen:
I just had a quick call with @yury-lysogorskiy and we decided to open a pull request with his new changes https://github.com/pyiron/pyiron/pull/565 . @samwaseda Can you take a look at this?

samwaseda:
So you decided to not reflect any of my suggestions above?

jan-janssen:
No. Yury had already made the changes before, all I want is to merge both discussions.

samwaseda:
Ok I submitted the comment

jan-janssen:
@samwaseda I close the issue for now and we can continue the discussion in https://github.com/pyiron/pyiron/pull/565

PR #1208, opened 2020-11-20
K mesh ang

sudarsan-surendralal:


pyiron/dft/job/generic.py
```diff
@@ -338,6 +386,37 @@ def get_electronic_structure(self):
...
             return es_obj
 
+    def modify_kpoints(self):
+        if self.k_mesh_per_reciprocal_angstrom is not None:
+            self.set_kpoints(center_shift=self.k_mesh_center_shift,
+                             kpoints_per_reciprocal_angstrom=self.k_mesh_per_reciprocal_angstrom,
+                             symmetry_reduction=self.reduce_kpoint_symmetry)
+
+
+def get_k_mesh_by_density(cell, kmesh_density_per_inverse_angstrom=1.0):
```
samwaseda, line 396: Is there a particular reason that `kmesh_density_per_inverse_angstrom` has a default value?

sudarsan-surendralal: No particular reason since this is a helper function @yury-lysogorskiy wrote. The function within the class does not have a default value.

sudarsan-surendralal:
The tests fail because of a factor of `2 * np.pi`. @samwaseda and @yury-lysogorskiy, any preference as to which units we should use?

```
FAIL: test_set_kpoints (sphinx.test_base.TestSphinx)
  File "/home/runner/work/pyiron/pyiron/tests/sphinx/test_base.py", line 273, in test_set_kpoints
    self.assertEqual(self.sphinx_band_structure.get_k_mesh_by_cell(2.81/2/np.pi), [1,1,1])
remove directory: /home/runner/work/pyiron/pyiron/tests/testing/testing_serial_non_modal/
AssertionError: Lists differ: [7, 7, 7] != [1, 1, 1]
```

samwaseda:
> The tests fail because of a factor of `2 * np.pi`. @samwaseda and @yury-lysogorskiy, any preference as to which units we should use?
> 
> ```
> FAIL: test_set_kpoints (sphinx.test_base.TestSphinx)
>   File "/home/runner/work/pyiron/pyiron/tests/sphinx/test_base.py", line 273, in test_set_kpoints
>     self.assertEqual(self.sphinx_band_structure.get_k_mesh_by_cell(2.81/2/np.pi), [1,1,1])
> remove directory: /home/runner/work/pyiron/pyiron/tests/testing/testing_serial_non_modal/
> AssertionError: Lists differ: [7, 7, 7] != [1, 1, 1]
> ```

That sounds weird to me because 2pi is still there. Anyway we had a similar discussion [here](https://github.com/pyiron/pyiron/issues/537).

pyiron/dft/job/generic.py
samwaseda: Then isn't it better to remove the default value and set the same order as in `get_k_mesh_by_cell`?

samwaseda:
> The tests fail because of a factor of `2 * np.pi`. @samwaseda and @yury-lysogorskiy, any preference as to which units we should use?
> 
> ```
> FAIL: test_set_kpoints (sphinx.test_base.TestSphinx)
>   File "/home/runner/work/pyiron/pyiron/tests/sphinx/test_base.py", line 273, in test_set_kpoints
>     self.assertEqual(self.sphinx_band_structure.get_k_mesh_by_cell(2.81/2/np.pi), [1,1,1])
> remove directory: /home/runner/work/pyiron/pyiron/tests/testing/testing_serial_non_modal/
> AssertionError: Lists differ: [7, 7, 7] != [1, 1, 1]
> ```

I just corrected the test. It was because it now apparently takes the inverse value.

sudarsan-surendralal:
> 
> 
> > The tests fail because of a factor of `2 * np.pi`. @samwaseda and @yury-lysogorskiy, any preference as to which units we should use?
> > ```
> > ======================================================================
> > FAIL: test_set_kpoints (sphinx.test_base.TestSphinx)
> > ----------------------------------------------------------------------
> > Traceback (most recent call last):
> >   File "/home/runner/work/pyiron/pyiron/tests/sphinx/test_base.py", line 273, in test_set_kpoints
> >     self.assertEqual(self.sphinx_band_structure.get_k_mesh_by_cell(2.81/2/np.pi), [1,1,1])
> > remove directory: /home/runner/work/pyiron/pyiron/tests/testing/testing_serial_non_modal/
> > AssertionError: Lists differ: [7, 7, 7] != [1, 1, 1]
> > ```
> 
> I just corrected the test. It was because it now apparently takes the inverse value.

Ah, I see! In this implementation, a lower `k_mesh_per_reciprocal_angstrom` value results in a denser mesh. Will this affect your workflows?

samwaseda:
> 
> Ah, I see! In this implementation, a lower `k_mesh_per_reciprocal_angstrom` value results in a denser mesh. Will this affect your workflows?

No it's perfectly fine :). But the DocString:

```
Number of unreduced k-points per Angstrom of the lattice vector
```

might be a bit misleading, because if it's really the number per angstrom, I would think `k_mesh_per_reciprocal_angstrom=1` means one k-point per angstrom.

pyiron/dft/job/generic.py
```diff
@@ -173,22 +218,21 @@ def get_kpoints(self):
...
+            list/numpy.ndarray: Mesh size
 
-            Args:
-                kpoints_per_reciprocal_angstrom: (float) number of k-points per reciprocal angstrom (i.e. per 2*pi / box_length)
-                cell: (list/ndarray) 3x3 cell. If not set, the current cell is used.
         """
         if cell is None:
             if self.structure is None:
-                raise AssertionError('structure not set')
+                raise ValueError("Can't genreate k-points without structure being set")
```
samwaseda, line 233: Maybe `if cell is None and self.structure is None` (saves one line :D)?

And `genreate` -> `generate`

pyiron/dft/job/generic.py
yury-lysogorskiy: Now, from my previous practical experience, I would recommend default value of `kmesh_density_per_inverse_angstrom=0.5`, even if that would be overwritten by the external calling method...

pyiron/dft/job/generic.py
sudarsan-surendralal: Typo fixed but we do need two if statements since we might have to overwrite the cell parameter.

samwaseda: Ah you're right

pyiron/dft/job/generic.py
yury-lysogorskiy: And if we still could make a name-changes, I found a better name for that parameter: `kmesh_spacing`

samwaseda: I'm fine with giving a default value to k point mesh or k point density, but I think we shouldn't do both at the same time. And if we don't give a default value in one place, it shouldn't appear in other places either for the same argument. This being said, I would have preferred @yury-lysogorskiy's suggestion to set a default value for the k point density (over setting k-point mesh), but in the current implementation in pyiron it's probably not a very practical solution.

And I agree with @yury-lysogorskiy on the name change. Rather than making the description of what it does accurately, it's important that the user understands clearly whether an increasing `k_xxx` leads to an increasing number of k points or other way around. In this regard, I also think `spacing` does a good job.

sudarsan-surendralal: I've made the name change. By default `k_mesh_spacing` is `None` which means whatever k-point setting is the default (4x4x4 I think) will be used. Only when you set `k_mesh_spacing`  explicitly or set it through `set_kpoints()` does this value change based on cell shape.

samwaseda (approved):
Brilliant!

yury-lysogorskiy:
I'm afraid that there is still incoherent behaviour:
```python
job_periodic=pr.create_job(pr.job_type.Vasp, "Li_fcc_static")
crystal=pr.create_ase_bulk("Li","fcc",a=3.5, cubic=False)
job_periodic.structure=crystal
job_periodic.set_kpoints(k_mesh_spacing=0.15)
# now job_periodic.input.kpoints has 21 x 21 x 21, which is CORRECT
```
Then, run `PhonopyJob`:
```python
ph_job = pr.create_job(pr.job_type.PhonopyJob, "phonoy_vasp")
ph_job.input["supercell"]=[2,2,2]
ph_job.ref_job=job_periodic
ph_job.run()
```
Supercell's VASP job has CORRECT `KPOINTS` file with 10x10x10 kmesh, but
`supercell_subjob.input.kpoints` has WRONG old 21x21x21.
That probably happend because after calling `modify_kpoints` in `VaspBase.write_input`, updated k-mesh wasn't stored into HDF.

PR #1211, opened 2020-11-21
Get layers and add `Analyse`

samwaseda:
Now, after so many times discussing the issue, I finally created a new class `Analyse`, which should include structure analysis tools. For now, there is only one new function `get_layers`, which enumerates layers in a structure. Feel free to leave comments.


pyiron/atomistics/structure/atoms.py
```diff
@@ -192,6 +193,7 @@ def __init__(
         else:
             self.dimension = 0
         self.visualize = Visualize(self)
```
jan-janssen, line 195: Should not self.visualize and self.analyse be hidden variables plus the corresponding properties to prevent the user from modifying them accidentally.

samwaseda: The ultimate goal is to reduce the number of items in `Atoms`, since the tab completion list is now a total chaos. So for this one, no, it shouldn't be a hidden variable.

For `self.visualize` I'm less sure though.

jan-janssen: I mean:
```
def __init__(self):
    self._visualize = ...

@property
def visualize(self):
    return self._visualize 
```

samwaseda: Ah sounds like a good idea

pyiron/atomistics/structure/analyse.py
```diff
@@ -0,0 +1,72 @@
...
+__maintainer__ = "Sam Waseda"
+__email__ = "waseda@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
+
+s = Settings()
+
+class Analyse:
+    """ Class to analyse atom structure.  """
+    def __init__(self, ref_structure):
```
liamhuber, line 24: Nit: I guess this can just be `structure` unless we wind up creating child structures based off it. 

Also, using `self._structure = structure.copy()` might be safer so no one accidentally modifies a structure in place during analysis.

pyiron/atomistics/structure/analyse.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Args:
+            ref_structure (pyiron.atomistics.structure.atoms.Atoms): reference Atom structure
+        """
+        self._ref_structure = ref_structure
+
+    def __repr__(self):
+        """
+        Returns: __repr__
```
liamhuber, line 33: Nit: Codacy will probably complain about the 1 line docstring taking three lines.

pyiron/atomistics/structure/analyse.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        return ''
```
liamhuber, line 35: Nit: Honestly, I would rather just see the ugly python string? 

```python
job.structure.analyse
> 
```

feels funny... `"Analysis object"` would be fine in place of the machine-readable default as well.

pyiron/atomistics/structure/analyse.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def get_layers(self, distance_threshold=0.01, id_list=None):
+        """
+        Get an array of layer numbers.
```
liamhuber, line 39: Do I understand correctly that this is somehow scanning linkages in the cartesian x,y,z directions?

I think the docstring still needs more fleshing out. 

Also, I don't understand it well enough so maybe it takes care of this already, but what about non-rectangular cells?

Finally, does there need to be some sort of option to use the unwrapped positions, or even to force-wrap the positions?

pyiron/atomistics/structure/analyse.py
```diff
@@ -0,0 +1,72 @@
...
+            id_list = np.arange(len(self._ref_structure))
+        layers = []
+        for x in self._ref_structure.positions[np.array(id_list)].T:
+            cluster = AgglomerativeClustering(
+                linkage='complete',
+                n_clusters=None,
+                distance_threshold=distance_threshold
+            ).fit(x.reshape(-1,1))
+            args = np.unique(cluster.labels_, return_index=True)[1]
+            layers.append(x[args].argsort().argsort()[cluster.labels_])
```
liamhuber, line 71: Nit: Once we're argsorting our argsorts, some sort of technical explanation or decomposition down to well-named methods is probably helpful 😉

pyiron/atomistics/structure/atoms.py
```diff
@@ -191,7 +192,16 @@ def __init__(
             self.dimension = len(self.positions[0])
         else:
             self.dimension = 0
-        self.visualize = Visualize(self)
+        self._visualize = Visualize(self)
+        self._analyse = Analyse(self)
```
liamhuber, line 196: So this is not an issue with this PR per-se, as we already had `analyse_pyscal...`, but I see now that we mix American `Visualize` and British `Analyse`. In a future PR we should probably rename to pick just one.

tests/atomistics/structure/test_analyse.py
```diff
@@ -0,0 +1,28 @@
...
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+
+import unittest
+import numpy as np
+from pyiron.atomistics.structure.atoms import CrystalStructure
+
+
+class TestAtoms(unittest.TestCase):
+    @classmethod
```
liamhuber, line 11: Nit: If these are just passed I'm pretty sure they can be omitted without the code complaining.

tests/atomistics/structure/test_analyse.py
```diff
@@ -0,0 +1,28 @@
...
+    def setUpClass(cls):
+        pass
+
+    def test_allow_ragged(self):
+        a_0 = 4
+        struct = CrystalStructure(elements='Al', lattice_constants=a_0, bravais_basis='fcc').repeat(10)
+        self.assertEqual(
+            struct.analyse.get_layers().tolist(), np.rint(2*struct.positions/a_0).astype(int).tolist()
+        )
+
```
liamhuber, line 25: Is there a test to add to check if the `id_list` capability is also working as expected?

pyiron/atomistics/structure/analyse.py
```diff
@@ -0,0 +1,72 @@
...
+)
+__version__ = "1.0"
+__maintainer__ = "Sam Waseda"
+__email__ = "waseda@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
+
+s = Settings()
+
+class Analyse:
```
liamhuber, line 22: Should inherit from pyiron_base.PyironFactory

liamhuber (changes requested):
Looking good. I covered it in nits and some larger-than-nit-but-still-small comments, but I don't see any structural problems. I'm excited how quickly all these changes are being implemented.

I would like to see the existing analysis methods moved over here (we can keep the current interface, but under the hood it should just be running the stuff here). I'm also content if we save that for an immediately-following PR.

pyiron/atomistics/structure/analyse.py
samwaseda: We can't copy the structure, because `structure.analyse.do_something()` should always reflect the current structure.

> Nit: I guess this can just be structure unless we wind up creating child structures based off it.

Is it about the variable name (i.e. are you suggesting to change `ref_structure` to `structure`)?

pyiron/atomistics/structure/analyse.py
samwaseda: > Do I understand correctly that this is somehow scanning linkages in the cartesian x,y,z directions?

Yes!

> 
> I think the docstring still needs more fleshing out.
> 
> Also, I don't understand it well enough so maybe it takes care of this already, but what about non-rectangular cells?

I also thought about it, but for now it simply looks at the absolute coordinates. I can also include something like `frame` which would take a 2d tensor to define the planes.

> 
> Finally, does there need to be some sort of option to use the unwrapped positions, or even to force-wrap the positions?

You have to have wrapped the positions beforehand. Do you want me to add an extra argument for that?

pyiron/atomistics/structure/atoms.py
samwaseda: Yeah first I had `analyze` but then I saw the conflict with pyscal. Since `visualize` is not supposed to be used explicitly, we can change it to `visualise`. Otherwise we simply explain Brits prefer analytic stuff and North Americans prefer visual effects 😎.

pyiron/atomistics/structure/analyse.py
samwaseda: What does it change?

pyiron/atomistics/structure/analyse.py
samwaseda: I wasn't really sure what to put there but by removing it I can reach 100% coverage so I remove it :)

pyiron/atomistics/structure/analyse.py
liamhuber: > We can't copy the structure, because `structure.analyse.do_something()` should always reflect the current structure.

You're right, I was not thinking clearly. I mean we always *can* copy a structure, but in this case I now see the foolishness of doing so.

> Is it about the variable name (i.e. are you suggesting to change `ref_structure` to `structure`)?

Yes. I think `ref` here is redundant so let's scratch it.

pyiron/atomistics/structure/analyse.py
liamhuber: > I also thought about it, but for now it simply looks at the absolute coordinates. I can also include something like frame which would take a 2d tensor to define the planes.

Ok, I guess this should be clarified in the docstring but I don't have any problem with it as current behaviour.

> You have to have wrapped the positions beforehand. Do you want me to add an extra argument for that?

Hmm, no, maybe no requirement to wrap beforehand since, as you say, you can do that already at the structure level. But does allowing a flag to use unwrapped coordinates make sense?

pyiron/atomistics/structure/atoms.py
liamhuber: 😝

pyiron/atomistics/structure/analyse.py
liamhuber: I spoke too soon anyhow. 

At the moment it adds an equivalence between `__getitem__` and `getattr`, which isn't much.

I had originally suggested it because eventually I guess we'll be iterating over factory classes and adding them programmatically depending on the modules which have been imported. But this was stupid of me, because that's just for factories which will belong to the `Creator` class of `Project` -- which this obviously doesn't. So please do not inherit from `PyironFactory` 😛

pyiron/atomistics/structure/analyse.py
samwaseda: I added wrapping

pyiron/atomistics/structure/analyse.py
samwaseda: I just tried out a few more possibilities, but I don't really think I can make it easier to understand... Let me think about it a bit more.

liamhuber: I'm not sure the code itself needs to be any easier, just explained in english or (better (in my opinion)) wrapped in a well-named function, either in part or whole.

pmrv: My code somewhere has a similar `argsort(argsort)` call and the following comment attached that makes it reasonably clear (to me)

```
    # argsort(argsort) looks weird, but what we're trying to do here is
    # permuting the plane indices such that label 0 corresponds to the plane
    # closest to the origin, 1 to the second closest, etc.  For that we first
    # calculate the mean x position of all labels and argsort it.  This gives a
    # mapping from the sorted indices to the labels returned by the clustering,
    # i.e. argsort(x)[0] maps the first sorted index 0 to the cluster index
    # closest to the origin, etc.  However we actually want to have the inverse
    # mapping, from cluster indices to sorted indices!  Curiously argsort also
    # inverts index permutations.  Inverting a index permutation means building
    # an array where the ith item is the index at which i appears in the
    # original permutation, which is just sorting indices into the original
    # permutation array according to the value in the permuation array. i.e.
    # argsorting.
```

feel free to adapt it.

pmrv: How fast is the agglomerative clustering btw?  I use this exact code for my own plane finding, but with DBSCAN and can't remember big problems.  Just curious, I see no need to change it if this is tested and fast, tough.  Everyone has their own pet clustering algorithm it seems. :')

samwaseda: > My code somewhere has a similar `argsort(argsort)` call and the following comment attached that makes it reasonably clear (to me)
> 
> ```
>     # argsort(argsort) looks weird, but what we're trying to do here is
>     # permuting the plane indices such that label 0 corresponds to the plane
>     # closest to the origin, 1 to the second closest, etc.  For that we first
>     # calculate the mean x position of all labels and argsort it.  This gives a
>     # mapping from the sorted indices to the labels returned by the clustering,
>     # i.e. argsort(x)[0] maps the first sorted index 0 to the cluster index
>     # closest to the origin, etc.  However we actually want to have the inverse
>     # mapping, from cluster indices to sorted indices!  Curiously argsort also
>     # inverts index permutations.  Inverting a index permutation means building
>     # an array where the ith item is the index at which i appears in the
>     # original permutation, which is just sorting indices into the original
>     # permutation array according to the value in the permuation array. i.e.
>     # argsorting.
> ```
> 
> feel free to adapt it.

I copy&pasted it (even though I find it very long).

samwaseda: > How fast is the agglomerative clustering btw? I use this exact code for my own plane finding, but with DBSCAN and can't remember big problems. Just curious, I see no need to change it if this is tested and fast, tough. Everyone has their own pet clustering algorithm it seems. :')

I don't really think there'd be a speed issue. There are simply two reasons why I use the agglomerative clustering: 1, I know what it does behind the scenes; 2, it doesn't have unidentified points.

samwaseda: Btw @pmrv, whenever you have a code like this, I would strongly encourage you to implement it in pyiron as well. I guess we all have to get used to implementing our own tools in pyiron, but also rejecting some ideas if we think it shouldn't be in pyiron.

pyiron/atomistics/structure/analyse.py
samwaseda: > Also, I don't understand it well enough so maybe it takes care of this already, but what about non-rectangular cells?
> 

In the end I added `planes` where you can say things like `planes=[1,1,1]` to get planes along the `[1,1,1]`-direction.

jan-janssen (approved):
Looks good to me, once this is merged we can migrate the analysis functions from pyscal and phonopy as well.

pyiron/atomistics/structure/analyse.py
pmrv: > > How fast is the agglomerative clustering btw? I use this exact code for my own plane finding, but with DBSCAN and can't remember big problems. Just curious, I see no need to change it if this is tested and fast, tough. Everyone has their own pet clustering algorithm it seems. :')
> 
> I don't really think there'd be a speed issue. There are simply two reasons why I use the agglomerative clustering: 1, I know what it does behind the scenes; 2, it doesn't have unidentified points.

I agree, well-understood code is most important.

> Btw @pmrv, whenever you have a code like this, I would strongly encourage you to implement it in pyiron as well. 

Guilty as charged, it's been in my backlog.

samwaseda: Sorry for leaving comments incessantly, but I changed my mind and separated the commands in more lines+the variable names are now probably easier to understand. Sorry @pmrv for creating this fuss, but in the end I found your comments too long...

samwaseda:
I just added one more feature: now the algorithm can see layers across periodic boundaries. You can test it with the following lines:

```
structure = Project('.').create_structure('Al', 'fcc', 4).repeat(2)
structure.cell[1,0] += 0.01
structure.center_coordinates_in_unit_cell()
layers = structure.analyse.get_layers()
structure.plot3d(scalar_field=layers[:,0])
```

liamhuber:
> Liam-proof writing? Super unlikely though

😝 Guilty! Also, the leading sentence of the docstring is missing a terminating "." 😉

liamhuber (approved):
Super duper :)

pyiron/atomistics/structure/analyse.py
```diff
@@ -0,0 +1,133 @@
...
+            raise ValueError('distance_threshold must be a positive float')
+        if id_list is not None and len(id_list)==0:
+            raise ValueError('id_list must contain at least one id')
+        if wrap_atoms and planes is None:
+            positions, indices = self._structure.get_extended_positions(
+                width=distance_threshold, return_indices=True
+            )
+            if id_list is not None:
+                id_list = np.arange(len(self._structure))[np.array(id_list)]
+                id_list = np.any(id_list[:,None]==indices[None,:], axis=0)
```
pmrv, line 93: Nit: use np.newaxis

tests/atomistics/structure/test_analyse.py
```diff
@@ -0,0 +1,47 @@
...
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+
+import unittest
+import numpy as np
+from pyiron.atomistics.structure.atoms import Atoms, CrystalStructure
+
+
+class TestAtoms(unittest.TestCase):
+    def test_get_layers(self):
```
pmrv, line 11: Please break the tests into smaller functions and add either docstrings or short messages to the asserts.  This helps a lot if anyone else has to debug the tests.

pmrv (changes requested):

pyiron/atomistics/structure/analyse.py
samwaseda: It's not the same thing (that was what I thought I discovered recently)

samwaseda:
@jan-janssen somehow Codacy is not showing anything. It only says master is not analysed. Is it related to the changes you made over the weekend?

pyiron/atomistics/structure/analyse.py
pmrv: But `type(np.newaxis) == NoneType`, so it must be the same, no?

samwaseda: Yes so I thought it doesn’t matter which one I use.

pmrv: Yes, it doesn't matter for the code, but `np.newaxis` is documenting your intent.

samwaseda: ok :D

samwaseda:
If you (@pmrv) approve, I'll merge it, because I guess @jan-janssen is eager to move the structure analysis tools to the new class.

pyiron/atomistics/structure/analyse.py
```diff
@@ -0,0 +1,133 @@
...
+)
+__version__ = "1.0"
+__maintainer__ = "Sam Waseda"
+__email__ = "waseda@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
+
+s = Settings()
+
+def get_average_of_unique_labels(labels, values):
```
pmrv, line 23: Is this part of this PR or #1220?

samwaseda: Both. I think it's more appropriate to have it in `analyse.py` and I would import it in `neighbors.py` (so I'll delete the one in `neighbors.py` as soon as this one gets merged)

pmrv (approved):

PR #1212, opened 2020-11-21
Interface to lammps move

liamhuber:
A new lammps control interface to access [`fix move`](https://lammps.sandia.gov/doc/fix_move.html).

Currently tucked away hidden in `lammps.input.control` with no direct job access as it's probably just for experts/not something you can (currently) do in other codes(?).

### DIFF
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -822,6 +822,12 @@ class LammpsControl(GenericParameters):
             ids (list/numpy.ndarray): Integer ids of the atoms to move in the job's structure.
             velocity (list/numpy.ndarray/tuple): The velocity in x-y-z-direction for the group. `None` arguments are
                 converted to Lammps 'NULL' values and the velocity in this direction is left unchanged.
+
+        Warning: This fix does not exclude these atoms from
+            [other fixes](https://stackoverflow.com/questions/3402168/permanently-add-a-directory-to-pythonpath). You
+            may wish to combine this call with `selective_dynamics` on your corresponding structure. Future developers
+            can find a more complete discussion [here](https://github.com/pyiron/pyiron/pull/1212) when further
+            modifying this capability.
         """
         self._fix_to_three_vector(ids, velocity, 'move linear', LAMMPS_UNIT_CONVERSIONS[self["units"]]["velocity"])
 
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -799,6 +799,21 @@ class LammpsControl(GenericParameters):
             raise ValueError('Group ids must be non-negative to be parsed by Lammps, but got {}'.format(ids))
         self['group___{}'.format(group_name)] = 'id {}'.format(' '.join(np.array(ids).astype(int).astype(str)))
 
+    def _fix_to_three_vector(self, ids, vector, fix_string, conversion):
+        if len(vector) != 3:
+            raise ValueError('{} must have three components, but got {}'.format(fix_string, vector))
+        vector = list(vector)
+        for i, v in enumerate(vector):
+            if v is None:
+                vector[i] = 'NULL'
+            else:
+                vector[i] = str(v * conversion)
+        name = str(hash(tuple(ids))).replace('-', 'm')  # A unique name for the group
+        self._set_group_by_id(name, ids)
+        self['fix___{}_{}'.format(fix_string.replace(' ', '_'), name)] = '{} {} {}'.format(
+            name, fix_string, ' '.join(vector)
+        )
+
     def fix_move_linear_by_id(self, ids, velocity):
         """
         Displace atoms at each timestep. Creates a new group with a unique name based off the hash of the ids.
@@ -808,18 +823,17 @@ class LammpsControl(GenericParameters):
             velocity (list/numpy.ndarray/tuple): The velocity in x-y-z-direction for the group. `None` arguments are
                 converted to Lammps 'NULL' values and the velocity in this direction is left unchanged.
         """
-        if len(velocity) != 3:
-            raise ValueError('Velocity must have three components, but got {}'.format(velocity))
-        conversion = LAMMPS_UNIT_CONVERSIONS[self["units"]]["velocity"]
-        velocity = list(velocity)
-        for i, v in enumerate(velocity):
-            if v is None:
-                velocity[i] = 'NULL'
-            else:
-                velocity[i] = str(v * conversion)
-        name = str(hash(tuple(ids))).replace('-', 'm')  # A unique name for the group
-        self._set_group_by_id(name, ids)
-        self['fix___velocity_{}'.format(name)] = '{} move linear {}'.format(name, ' '.join(velocity))
+        self._fix_to_three_vector(ids, velocity, 'move linear', LAMMPS_UNIT_CONVERSIONS[self["units"]]["velocity"])
+
+    def fix_setforce_by_id(self, ids, force):
+        """
+        Set the force on a collection of atoms to a specified value.
+
+        ids (list/numpy.ndarray): Integer ids of the atoms to move in the job's structure.
+        force (list/numpy.ndarray/tuple): The force in x-y-z-direction for the group. `None` arguments are
+            converted to Lammps 'NULL' values and the force in this direction is left unchanged.
+        """
+        self._fix_to_three_vector(ids, force, 'setforce', LAMMPS_UNIT_CONVERSIONS[self["units"]]["force"])
 
     def _measure_mean_value(self, key_pyiron, key_lmp, every, atom=False):
         """
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -789,6 +789,28 @@ class LammpsControl(GenericParameters):
             self['dump___1'] += ' c_energy_pot_per_atom'
             self['dump_modify___1'] = self['dump_modify___1'][:-1] + ' %20.15g"'
 
+    def _set_group_by_id(self, group_name, ids):
+        self['group___{}'.format(group_name)] = 'id {}'.format(' '.join(np.array(ids).astype(int).astype(str)))
+
+    def fix_move_linear_by_id(self, ids, velocity):
+        """
+        Displace atoms at each timestep. Creates a new group with a unique name based off the hash of the ids.
+
+        Args:
+            ids (list/numpy.ndarray): Integer ids of the atoms to move in the job's structure.
+            velocity (list/numpy.ndarray/tuple): The velocity in x-y-z-direction for the group. `None` arguments are OK.
+        """
+        conversion = LAMMPS_UNIT_CONVERSIONS[self["units"]]["velocity"]
+        velocity = list(velocity)
+        for i, v in enumerate(velocity):
+            if v is None:
+                velocity[i] = 'NULL'
+            else:
+                velocity[i] = str(v * conversion)
+        name = str(hash(tuple(ids))).replace('-', 'm')  # A unique name for the group
+        self._set_group_by_id(name, ids)
+        self['fix___velocity_{}'.format(name)] = '{} move linear {}'.format(name, ' '.join(velocity))
+
     def _measure_mean_value(self, key_pyiron, key_lmp, every, atom=False):
         """
             Args:
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -792,6 +792,10 @@ class LammpsControl(GenericParameters):
     def _set_group_by_id(self, group_name, ids):
         if len(ids) < 1:
             raise ValueError('Group ids must have at least length one, but got {}'.format(ids))
+        if not np.all([isinstance(id_, (int, np.int64)) for id_ in ids]):
+            raise TypeError('Group ids must be integers, but got {}'.format(ids))
+        if np.any(ids < 0):
+            raise ValueError('Group ids must be non-negative to be parsed by Lammps, but got {}'.format(ids))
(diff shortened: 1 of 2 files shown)


liamhuber:
Me to @samwaseda after trying to get a low enough strain rate but still use interactive jobs:

you were right

Interactive = quick and easy = dark side 😝

pyiron/lammps/control.py
```diff
@@ -789,6 +789,32 @@ def energy_pot_per_atom(self):
...
+        """
+        if len(velocity) != 3:
+            raise ValueError('Velocity must have three components, but got {}'.format(velocity))
+        conversion = LAMMPS_UNIT_CONVERSIONS[self["units"]]["velocity"]
+        velocity = list(velocity)
+        for i, v in enumerate(velocity):
+            if v is None:
+                velocity[i] = 'NULL'
+            else:
+                velocity[i] = str(v * conversion)
```
samwaseda, line 813: Don't we have to rotate the velocity? (MOUWAHHAHAHAHAHAAAA!)

samwaseda (dismissed):

pyiron/lammps/control.py
```diff
@@ -789,6 +789,32 @@ def energy_pot_per_atom(self):
             self['dump___1'] += ' c_energy_pot_per_atom'
             self['dump_modify___1'] = self['dump_modify___1'][:-1] + ' %20.15g"'
 
+    def _set_group_by_id(self, group_name, ids):
+        if len(ids) < 1:
+            raise ValueError('Group ids must have at least length one, but got {}'.format(ids))
+        self['group___{}'.format(group_name)] = 'id {}'.format(' '.join(np.array(ids).astype(int).astype(str)))
```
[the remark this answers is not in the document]
liamhuber: Good concerns, both of those would fail in really ugly ways if allowed to pass here. Will fix it.

pyiron/lammps/control.py
liamhuber: Nope, it's fine 😉 

It is, of course, not fine. But the rotation matrix gets passed in explicitly to the `LammpsControl.calc_*` methods from modules that know something about the rotation matrix. Here there is no interface outside the control module, so we're not obligated to provide rotation support -- if it goes wrong it's the user's fault 😈 
> Currently tucked away hidden in lammps.input.control with no direct job access as it's probably just for experts/not something you can (currently) do in other codes(?).

pyiron/lammps/control.py
```diff
@@ -789,6 +789,32 @@ def energy_pot_per_atom(self):
...
+
+    def fix_move_linear_by_id(self, ids, velocity):
+        """
+        Displace atoms at each timestep. Creates a new group with a unique name based off the hash of the ids.
+
+        Args:
+            ids (list/numpy.ndarray): Integer ids of the atoms to move in the job's structure.
+            velocity (list/numpy.ndarray/tuple): The velocity in x-y-z-direction for the group. `None` arguments are OK.
```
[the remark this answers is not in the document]
liamhuber: Will change it to reflect both of these 👍

t-brink (changes requested):
First thing I see is that you can [find the following in the LAMMPS manual](https://lammps.sandia.gov/doc/fix_move.html):

> The atoms affected by this fix should not normally be time integrated by other fixes (e.g. fix nve, fix nvt), since that will change their positions and velocities twice.

I don't see any code to exclude the atoms affected by this fix from the normal time integration.

I also have the feeling that the tests are a bit superficial and should be extended. To catch the above issue, one would need to run a short simulation, though.

liamhuber:
> I don't see any code to exclude the atoms affected by this fix from the normal time integration.

Hmm, this is a very good note! Currently I use it in conjunction with selective dynamics, so this becomes a non-issue as the atoms are excluded anyhow. Since this is anyhow buried inside `job.input.control` with no direct access for non-expert users, I propose to cover this with a warning rather than fixing it outright. The warning will explain the need for selective dynamics and link to the documentation (and PR?) so a future developer gets a head start of she/he wants to do a deeper fix of the issue. Will patch and push as soon as I get a chance :)

> I also have the feeling that the tests are a bit superficial and should be extended. To catch the above issue, one would need to run a short simulation, though.

Yes, this is a fair concern. Right now we have a standard procedure to avoid running actual calculations in the tests for the sake of speed. I think there are a number of places where we have this exact same concern though, and here I'll hold off for a more general solution to all of these.

pyiron/lammps/control.py
```diff
@@ -789,6 +789,58 @@ def energy_pot_per_atom(self):
...
+        if len(ids) < 1:
+            raise ValueError('Group ids must have at least length one, but got {}'.format(ids))
+        if np.any([isinstance(id_, bool) or not isinstance(id_, (int, np.int64)) for id_ in ids]):
+            # Note: it turns out bool is a subclass of int. Weird, eh.
+            raise TypeError('Group ids must be integers, but got {}'.format(ids))
+        if np.any(np.array(ids) < 0):
+            raise ValueError('Group ids must be non-negative to be parsed by Lammps, but got {}'.format(ids))
+        self['group___{}'.format(group_name)] = 'id {}'.format(' '.join(np.array(ids).astype(int).astype(str)))
+
+    def _fix_to_three_vector(self, ids, vector, fix_string, conversion):
```
t-brink, line 802: shouldn't that be named `_three_vector_to_fix` or `_fix_from_three_vector`?

pyiron/lammps/control.py
```diff
@@ -789,6 +789,58 @@ def energy_pot_per_atom(self):
...
+        if len(vector) != 3:
+            raise ValueError('{} must have three components, but got {}'.format(fix_string, vector))
+        vector = list(vector)
+        for i, v in enumerate(vector):
+            if v is None:
+                vector[i] = 'NULL'
+            else:
+                vector[i] = str(v * conversion)
```
t-brink, line 810: Maybe a question of taste (got too much into functional programming once), but I'd avoid explicit indices in Python whenever possible and use a list comprehension:

```
vector = [str(v * conversion) if v is not None else 'NULL'
          for v in vector]
```

pyiron/lammps/control.py
```diff
@@ -789,6 +789,58 @@ def energy_pot_per_atom(self):
...
+        """
+        Displace atoms at each timestep. Creates a new group with a unique name based off the hash of the ids.
+
+        Args:
+            ids (list/numpy.ndarray): Integer ids of the atoms to move in the job's structure.
+            velocity (list/numpy.ndarray/tuple): The velocity in x-y-z-direction for the group. `None` arguments are
+                converted to Lammps 'NULL' values and the velocity in this direction is left unchanged.
+
+        Warning: This fix does not exclude these atoms from
+            [other fixes](https://stackoverflow.com/questions/3402168/permanently-add-a-directory-to-pythonpath). You
```
t-brink, line 827: that's not the right link, is it?

pyiron/lammps/control.py
```diff
@@ -789,6 +789,58 @@ def energy_pot_per_atom(self):
...
+        )
+
+    def fix_move_linear_by_id(self, ids, velocity):
+        """
+        Displace atoms at each timestep. Creates a new group with a unique name based off the hash of the ids.
+
+        Args:
+            ids (list/numpy.ndarray): Integer ids of the atoms to move in the job's structure.
+            velocity (list/numpy.ndarray/tuple): The velocity in x-y-z-direction for the group. `None` arguments are
+                converted to Lammps 'NULL' values and the velocity in this direction is left unchanged.
```
t-brink, line 824: Should that at least get a warning that `velocity` has to be rotated if the simulation box is rotated?

pyiron/lammps/control.py
```diff
@@ -789,6 +789,58 @@ def energy_pot_per_atom(self):
...
+
+        Warning: This fix does not exclude these atoms from
+            [other fixes](https://stackoverflow.com/questions/3402168/permanently-add-a-directory-to-pythonpath). You
+            may wish to combine this call with `selective_dynamics` on your corresponding structure. Future developers
+            can find a more complete discussion [here](https://github.com/pyiron/pyiron/pull/1212) when further
+            modifying this capability.
```
t-brink, line 830: This warning basically means that this fix must currently not be used except for very special cases. Is it hidden enough that the warning suffices? Or should trying to use it with `calc_md` raise a `NotImplementedError`?

pyiron/lammps/control.py
liamhuber: The problem is that at the `LammpsControl` level we have no access to whether or not the box is rotated; this only comes in when the rotation matrix is explicitly passed. We could certainly make this general in the future (and then allow accessing this power directly at the job level) but right now this command knows nothing about it.

pyiron/lammps/control.py
liamhuber: Yikes, nope! I will let the tests finish, then fix this.

pyiron/lammps/control.py
liamhuber: Yep, I like that better. Again, I'll let the tests finish first but then I'll make this change.

pyiron/lammps/control.py
liamhuber: Or `_fix_with_three_vector` or `_three_vector_fix`... 😛 TBH I'm not 100% satisfied with any of these, nor 100% unhappy with any of them. I'll let it stew while the tests run and might rename in the next set of pushes. Since you also provide more than one option I'll assume you're not totally passionate about it either 😉

t-brink: I understand the meaning of all of them, except `_fix_to_three_vector`. That has the opposite meaning of all the others to me, i.e., I'd expect to pass in a fix and get a three vector.

pyiron/lammps/control.py
t-brink: I meant a warning in the docstring to reduce the surprise.

pyiron/lammps/control.py
liamhuber: > This warning basically means that this fix must currently not  be used except for very special cases. 

Yep, 100%.

> Is it hidden enough that the warning suffices? 

This is a good point. Instead of just a docstring warning I'll add a `warnings.warn` as well.

> Or should trying to use it with `calc_md` raise a `NotImplementedError`?

No dice. I actually did all this specifically for moving a few atoms in an MD run (toy nanopillar indentation), so breaking it in conjunction with MD would be shooting myself in the foot 😝

pyiron/lammps/control.py
liamhuber: Ah -- You bet!

liamhuber:
@jan-janssen can you offer any insight why the windows build (and only windows) is failing??

jan-janssen:
```
  File "D:\a\pyiron\pyiron\tests\lammps\test_control.py", line 159, in test_fix_move_linear_by_id
    fix(good_ids, good_velocities)
  File "D:\a\pyiron\pyiron\pyiron\lammps\control.py", line 835, in fix_move_linear_by_id
    self._fix_with_three_vector(ids, velocity, 'move linear', LAMMPS_UNIT_CONVERSIONS[self["units"]]["velocity"])
  File "D:\a\pyiron\pyiron\pyiron\lammps\control.py", line 812, in _fix_with_three_vector
    self._set_group_by_id(name, ids)
  File "D:\a\pyiron\pyiron\pyiron\lammps\control.py", line 797, in _set_group_by_id
    raise TypeError('Group ids must be integers, but got {}'.format(ids))
TypeError: Group ids must be integers, but got [0 1 2 3]
```

Maybe it is better to copy the job object when you apply these commands, it is possible that some tests are executed in parallel.

liamhuber:
> Maybe it is better to copy the job object when you apply these commands, it is possible that some tests are executed in parallel.

@jan-janssen I'm afraid I don't understand what you mean; afaik there is no job object involved, the tests directly instantiate a `LammpsControl` object. I had seen the error you posted, but this why I was confused because `[0 1 2 3]` sure look like integers to me. I even when in and added a `dtype=int` to the `np.arange(4)` that's causing the failure.

pyiron/lammps/control.py
```diff
@@ -789,6 +789,65 @@ def energy_pot_per_atom(self):
             self['dump___1'] += ' c_energy_pot_per_atom'
             self['dump_modify___1'] = self['dump_modify___1'][:-1] + ' %20.15g"'
 
+    def _set_group_by_id(self, group_name, ids):
+        if len(ids) < 1:
+            raise ValueError('Group ids must have at least length one, but got {}'.format(ids))
+        if np.any([isinstance(id_, bool) or not isinstance(id_, (int, np.int64)) for id_ in ids]):
```
jan-janssen, line 795:
```suggestion
        if np.any([isinstance(id_, bool) or not isinstance(id_, (int, np.integer)) for id_ in ids]):
```

liamhuber: *sigh* right, so the windows systems are not 64 bit or something silly like this. Ugh. But thank you! I don't think I ever would have thought down this line myself!

jan-janssen: Programming for windows is annoying - we had the same issue before https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/structure/sparse_list.py#L174 and https://github.com/pyiron/pyiron/blob/master/pyiron/atomistics/structure/atoms.py#L136

jan-janssen: Maybe it makes sense to have a function in `pyiron_base` which provides this kind of utility tasks like https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/generic/util.py#L37

jan-janssen (approved):
Looks good to me

PR #1215, opened 2020-11-22
Interface method use pyscal

jan-janssen:


jan-janssen:
Requires https://github.com/pyiron/pyiron/pull/1216

jan-janssen:
I faced a strange bug when I tested the melting point protocol - so I put this pull request on hold until we figured out what caused pyscal to crash https://github.com/srmnitc/pyscal/issues/90

PR #1219, opened 2020-11-22
protect master branch

jan-janssen:


pmrv:
What does this do?

niklassiemer:
If I understand this correctly, merging a PR now deletes the corresponding branch automatically. This is to ensure that the master branch is always excluded and that closed PR are not deleted.

PR #1220, opened 2020-11-22
Cluster neighbors

samwaseda:
@pmrv and I were nearly coming to the conclusion that `DBSCAN` is a better method for the clustering of neighbor shells, but then I realized that it doesn't work reliably when the number of atoms is very small, so in the end I used `AgglomerativeClustering`.


samwaseda:
To merge after [analyse](https://github.com/pyiron/pyiron/pull/1211) (because I'd like to put `get_average_of_unique_labels` in `analyse.py`)

pyiron/atomistics/structure/neighbors.py
```diff
@@ -21,6 +21,27 @@
 
 s = Settings()
 
+def get_average_of_unique_labels(labels, values):
```
pmrv, line 24: I find this version

```python
return np.array([ values[ labels == s ].mean(axis = 0)
                                  for s in np.unique(labels) ])
```

much clearer, but it does construct a mostly empty index array for every label that your version avoids with the sparse matrix.  I did some mini-benchmarks and the sparse version is faster above ~10,000 values, but only by ~10% (or <1µs per value).  

No hard feelings either way, if you want to keep this version that's also fine by me.

pyiron/atomistics/structure/neighbors.py
```diff
@@ -695,7 +716,9 @@ def find_neighbors_by_vector(self, vector, deviation=False):
             return indices[np.arange(len(dist)), np.argmin(dist, axis=-1)], np.min(dist, axis=-1)
         return indices[np.arange(len(dist)), np.argmin(dist, axis=-1)]
 
-    def cluster_by_vecs(self, bandwidth=None, n_jobs=None, max_iter=300):
+    def cluster_by_vecs(
+        self, distance_threshold=None, n_clusters=None, linkage='complete', affinity='euclidean'
+    ):
```
pmrv, line 721: Maybe we want to group the clustering specific arguments into a dictionary like scipy functions do sometimes [see here under options](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize).  Not necessary if you're convinced that this clustering method will stay for a while.

pmrv: Do we ever want to change the metric though?

pyiron/atomistics/structure/neighbors.py
```diff
@@ -704,17 +727,40 @@ def cluster_by_vecs(self, bandwidth=None, n_jobs=None, max_iter=300):
...
+    def cluster_by_distances(
+        self,
+        distance_threshold=None,
+        n_clusters=None,
+        linkage='complete',
+        affinity='euclidean',
+        use_vecs=False,
+    ):
         """
         Method to group vectors which have similar values. This method should be used as a part of
```
pmrv, line 765: Shoul read "Method to group vectors which have similar *lengths*."

pmrv (approved):
Looks good to me, except for the small stuff I commented.

pmrv:
> To merge after [analyse](https://github.com/pyiron/pyiron/pull/1211) (because I'd like to put `get_average_of_unique_labels` in `analyse.py`)

If this is a user facing function I'd prefer a quick like
```
values.shape[0] == labels.shape[0]
```
to throw a clean error in case you pass weird stuff.

pyiron/atomistics/structure/neighbors.py
samwaseda: Oh, I agree that in things I'd show to other people I wouldn't do the same - I agree entirely with you that that's how it should be written in Notebooks.

On the other hand, as long as it's clear what the function does, and if it just does one single thing, I don't really mind having complex functions.

And here's the result of my speed comparison (beware of the log scales):

Unknown

After having seen this, I might still prefer the current implementation, also because we might find applications in other areas whenever we do clustering.

pyiron/atomistics/structure/neighbors.py
samwaseda: > Maybe we want to group the clustering specific arguments into a dictionary like scipy functions do sometimes [see here under options](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize). Not necessary if you're convinced that this clustering method will stay for a while.

That's actually a really good idea. Probably the best is to use `AgglomerativeClustering` for systems with a small number of atoms and `DBSCAN` with larger systems. I don't know exactly what would be the best way though. Do you perhaps have a specific idea?

samwaseda: > Do we ever want to change the metric though?

One thing is by using the Manhattan norm (or L1 norm), you can potentially enhance the angular variation, although that's orientation dependent, so I wouldn't generalise my claim here. Another thing is, if for some reason the items are periodic, you might want to put a norm with periodicity, even though I saw that [it makes the algorithm extremely slow](https://stackoverflow.com/questions/32938494/dbscan-with-metric-only-in-scikit-learn). So, to answer your question: No, probably not, but it's not impossible that we see such a case.

samwaseda:
> > To merge after [analyse](https://github.com/pyiron/pyiron/pull/1211) (because I'd like to put `get_average_of_unique_labels` in `analyse.py`)
> 
> If this is a user facing function I'd prefer a quick like
> 
> ```
> values.shape[0] == labels.shape[0]
> ```
> 
> to throw a clean error in case you pass weird stuff.

Actually it's a bit like a feature that the shape doesn't matter a lot here. So for example you can have `(N, N)` for `labels.shape` and `(N^2, 3)` for `values.shape`. This was from my point of view necessary, because the shapes I got from the clustering algorithms were not quite the ones that I had expected.

pyiron/atomistics/structure/neighbors.py
samwaseda: P.S. I'm gladly overruled if you still want me to change it.

pyiron/atomistics/structure/neighbors.py
pmrv: Will open a separate PR for that.

samwaseda: I'm personally very much in favour of either user being able to switch between algorithms or pyiron automatically detecting which one is better. The one I implemented here is not much more than a mere replacement of `MeanShift` by `AgglomerativeClustering`.

pyiron/atomistics/structure/neighbors.py
pmrv: I agree, but playing around some more I found this version

```python
def get_average_of_unique_labels(labels, values):
    u, counts = np.unique(labels, return_counts = True)
    means = np.zeros( (len(u), *values.shape[1:]) )
    means[labels] += values
    #manually broadcast counts to have the same dimension as means
    means /= np.reshape(counts, (-1,) + (1,)*len(values.shape[1:]))
    return means
```

which is twice as fast than the sparse version.  Except for the reshaping, I find it quite readable too.

samwaseda: Here, I don't get the same results as the other ones..

pmrv: Mh, true. `means[labels] += values` doesn't actually do what I though it would, my bad.

pmrv: Turns out you have to use `np.add.at`, but it's way too slow.  I'm ok with the sparse implementation then, but I think the output should have the same shape as values you pass in.  Right now the returned means are flattened and that might not be what you expect when you pass in e.g. matrices or other multidimensional stuff.

samwaseda: Return values can have the shape of the labels, but since we don't consider multidimensional labels, I would say it makes the most sense if we flatten the array. One option I see here is to return an array of single elements, instead of returning a flattened array, when there's only one element in each component. I didn't consider this as a particularly important case, but do you want me to change this?

samwaseda:
I was about to change the status from Draft to Ready, but now I realized that the recent changes made it possible for the vectors+distances to have `np.inf`, so I have to make changes to the clustering algorithms too...

samwaseda:
The shells for `np.inf` are now -1. I hope this is acceptable.

tests/atomistics/structure/test_neighbors.py
```diff
@@ -223,6 +223,42 @@ def test_get_local_shells(self):
         self.assertTrue(np.array_equal(shells, neigh.get_local_shells(cluster_by_vecs=True)))
         self.assertFalse(np.array_equal(shells, neigh.get_local_shells()))
 
+    def test_get_global_shells_ragged(self):
+        structure = CrystalStructure(
+            elements='Al', lattice_constants=4, bravais_basis='fcc'
+        ).repeat(2)
+        del structure[0]
+        neigh = structure.get_neighbors(cutoff_radius=3.5, num_neighbors=None)
+        self.assertEqual(np.sum(neigh.get_global_shells()==-1), 12)
```
pmrv, line 232: Why should it be -1 for 12 neighbors? Shouldn't those be correctly classified?

pmrv (approved):
I'd like a bit more documentend intent in the tests, but generally looks good, feel free to merge it.

tests/atomistics/structure/test_neighbors.py
samwaseda: Oh that's probably an important point that should have been better documented in the docstring. -1 is the equivalent of `np.inf` for `distances`. So there are now three conventions that coexist:

- `np.inf` for `distances` and `vecs`
- `n_atoms+1` for `indices`
- -1 for `shells`

I hope we can live with these ones.

PR #1227, opened 2020-11-25
Interface Method: Small updates 

jan-janssen:


samwaseda:
If there's no unit test and no DocString, this module should not be in `pyiron` but in `pyiron_contrib`.

jan-janssen:
> If there's no unit test and no DocString, this module should not be in `pyiron` but in `pyiron_contrib`.

I agree - but we published the corresponding paper before the introduction of `pyiron_contrib` that is why it was implemented here in the first place. Still as the functions are very specific it maybe makes even more sense to publish a separate module for the interface method.

liamhuber:
The fix itself is obviously fine...but -1 to the PR name 🤮 I think basically every dev who sees it in their notifications needs to come look at it when all they read is "another small fix" 😉

samwaseda:
> I agree - but we published the corresponding paper before the introduction of `pyiron_contrib` that is why it was implemented here in the first place. Still as the functions are very specific it maybe makes even more sense to publish a separate module for the interface method.

I'm really totally perfectly fine with this one being where it is right now. The problem is that the file looks like this:

```
def some_function():
    ...

def another_function():
    ...

def more_function():
    ...

def random_function():
    ...
```

And nowhere it's clear which one is the one that I should begin with, or which ones are relevant to me. So I would like to have at least one large DocString at the beginning of the file, explaining what it does with a short example (I presume since you have the paper it should be more or less copy&paste?). Then if possible DocString for each public function+unit tests.

Issue #1173, opened 2020-11-12
Lammps **langevin** MD -- interactive differs from regular


liamhuber:
**Summary**

If you set up a Lammps MD job, you get different physics depending whether or not you run the job interactively when using the Langevin thermostat.

Since it still works OK with the Nose-Hoover thermostat I think this is low-priority, but we should at least get a warning/error added when `langevin=True` pretty quickly.

**Steps to Reproduce**

```python
from pyiron import Project
import matplotlib.pyplot as plt

pr = Project('scratch')
pr.remove_jobs_silently(recursive=True)

regular = pr.create_job(pr.job_type.Lammps, 'regular')
regular.structure = pr.create_ase_bulk('Al', cubic=True).repeat(3)
regular.potential = job.list_potentials()[0]
regular.calc_md(temperature=300, n_ionic_steps=1000, n_print=1, langevin=True)

interactive = regular.copy_to(new_job_name='interactive', new_database_entry=False)

regular.run()

interactive.interactive_open()
interactive.run()
interactive.interactive_close()

plt.plot(regular.output.temperature, label='regular')
plt.plot(interactive.output.temperature, label='interactive', linestyle='-.')
plt.legend()
plt.xlabel('Frame')
plt.ylabel('Temerpature [K]')

equil_frames = 100
print(
    regular.output.temperature[equil_frames:].mean(), 
    interactive.output.temperature[equil_frames:].mean()
)
```

**Expected Behavior**

The same statisical behaviour for both run styles.

**Actual Behavior**

Interactive lammps gives half the temperature that the regular job does. 

Note that as you increase `n_print`, the two results begin to converge. I guess this has to do with the fact that we really set the lammps number of steps to the print value [here](https://github.com/pyiron/pyiron/blob/7e48f661656b894b04a792e2c9be64ad0be220bb/pyiron/lammps/interactive.py#L314), so the lammps library is somehow starting a new calculation at each frame or something? 

I find the *exact* behaviour hard to understand, but the fact that they differ is really unsurprising given how `LammpsInteractive.run_if_interactive` reads. What is really blowing my mind right now is that with the Nose-Hoover thermostat it works fine, it's just Langevin that kills it.

**Further Information, Files, and Links**

With `langevin=True`:
![langevin](https://user-images.githubusercontent.com/20283329/98928017-8bfc4100-24d9-11eb-808c-39df3cffc030.png)

With `langevin=False`:
[...]

PR #1228, opened 2020-11-26
Warn users about combining Lammps interactive and langevin 

liamhuber:
interactive + langevin gives physics that depends on n_print.

Full discussion in #1173

### DIFF
--- a/pyiron/lammps/interactive.py
+++ b/pyiron/lammps/interactive.py
@@ -311,13 +311,6 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
 
     def run_if_interactive(self):
         if self._generic_input["calc_mode"] in ["md", "vcsgc"]:
-            if 'fix___langevin' in self.input.control.keys():
-                warnings.warn(
-                    "Langevin thermostatted MD in interactive mode only gives correct physics in the limit that the "
-                    "n_print variable goes to infinity. A more in-depth discussion can be found "
-                    "[here](https://github.com/pyiron/pyiron/issues/1173).",
-                    stacklevel=2
-                )
             self.input.control["run"] = self._generic_input["n_print"]
             super(LammpsInteractive, self).run_if_interactive()
             self._reset_interactive_run_command()
@@ -335,6 +328,18 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
             super(LammpsInteractive, self).run_if_interactive()
             self.interactive_execute()
             self.interactive_collect()
+
+    def validate_ready_to_run(self):
+        if self.server.run_mode.interactive \
+                and self._generic_input["calc_mode"] in ["md", "vcsgc"] \
+                and 'fix___langevin' in self.input.control.keys():
+            warnings.warn(
+                "Langevin thermostatted MD in interactive mode only gives correct physics in the limit that the "
+                "n_print variable goes to infinity. A more in-depth discussion can be found "
+                "[here](https://github.com/pyiron/pyiron/issues/1173).",
+                stacklevel=2
+            )
+        super().validate_ready_to_run()
 
     def run_if_interactive_non_modal(self):
         if not self._interactive_fetch_completed:
--- a/pyiron/lammps/interactive.py
+++ b/pyiron/lammps/interactive.py
@@ -311,6 +311,13 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
 
     def run_if_interactive(self):
         if self._generic_input["calc_mode"] in ["md", "vcsgc"]:
+            if 'fix___langevin' in self.input.control.keys():
+                warnings.warn(
+                    "Langevin thermostatted MD in interactive mode only gives correct physics in the limit that the "
+                    "n_print variable goes to infinity. A more in-depth discussion can be found "
+                    "[here](https://github.com/pyiron/pyiron/issues/1173).",
+                    stacklevel=2
+                )
             self.input.control["run"] = self._generic_input["n_print"]
             super(LammpsInteractive, self).run_if_interactive()
             self._reset_interactive_run_command()


samwaseda:
Wouldn't it be better to put it in `validate_ready_to_run`? I guess it's enough if it's shown at the first run.

liamhuber:
> Wouldn't it be better to put it in `validate_ready_to_run`? I guess it's enough if it's shown at the first run.

I definitely see where you're coming from, it's just more logic to put it there. We need to check (1) is interactive, (2) is calc_md (or calc_vcsgc) and (3) is langevin. Putting it here means the first two checks are already handled. Normally I prefer to fail earlier, but since this is such a niche topic I favoured development ease rather than ease of use.

samwaseda:
> (1) is interactive, (2) is calc_md (or calc_vcsgc) and (3) is langevin

I guess they cannot be changed after `validate_ready_to_run` is called, so having the check in `validate_ready_to_run` will safely capture the problem.

> since this is such a niche topic I favoured development ease rather than ease of use.

It is a niche topic only if `n_print` is extremely small, but the warning could be shown more often than wanted, because at least @jan-janssen and I both use Langevin and interactive at the same time.

liamhuber:
> > (1) is interactive, (2) is calc_md (or calc_vcsgc) and (3) is langevin
> 
> I guess they cannot be changed after `validate_ready_to_run` is called, so having the check in `validate_ready_to_run` will safely capture the problem.

Yes, it is safe in either location. I got off my lazy butt and put it in `validate_ready_to_run`.

liamhuber:
> > since this is such a niche topic I favoured development ease rather than ease of use.
> 
> It is a niche topic only if `n_print` is extremely small, but the warning could be shown more often than wanted, because at least @jan-janssen and I both use Langevin and interactive at the same time.

First, I think you+Jan still counts as niche 😝 

Second, my sympathy for you seeing the warning is extremely limited. You're getting the warning because the code is broken, and you're using a broken part of it. I can get behind your right as a super-user to go ahead and do this, but if you want the warning to go away you'll have to fix the code.

Right now interactive Lammps+Langevin silently modifies the *physics* based on an argument called `n_print`! The code doesn't follow through on its promise, and it behaves in a *fundamentally* different way for Langevin vs Nose-Hoover thermostats. You might have a super-user loophole to avoid paying the consequences for this trouble, but trouble it is and with a warning it must come.

samwaseda:
> 
> First, I think you+Jan still counts as niche 😝
> 
> Second, my sympathy for you seeing the warning is extremely limited. You're getting the warning because the code is broken, and you're using a broken part of it. I can get behind your right as a super-user to go ahead and do this, but if you want the warning to go away you'll have to fix the code.
> 
> Right now interactive Lammps+Langevin silently modifies the _physics_ based on an argument called `n_print`! The code doesn't follow through on its promise, and it behaves in a _fundamentally_ different way for Langevin vs Nose-Hoover thermostats. You might have a super-user loophole to avoid paying the consequences for this trouble, but trouble it is and with a warning it must come.

Wait we're not talking about the same - I'm saying it's enough to have **one** warning regardless of how often `run` is called, instead of having a warning for each `run`. Besides, a super user like me would simply `unfix` and `fix` Langevin with a different seed if they really want to have a new set of random numbers, in which case the problem stated in the warning is not pertinent anymore, but still we would have to see it come every time. So here's what's gonna happen when the first warning is issued:

Normal user: "Oh I shouldn't be doing this" -> stops using interactive, calc_md or Langevin
Super user: "Oh I have to redefine seeds" -> includes `fix` + `unfix` in the script.
Anakin Skywalker: "Oh I have to save my code" -> moves to the dark side.

In any of these cases, you just need the first warning. That's what I'm saying.

liamhuber:
> Wait we're not talking about the same 

I'm increasingly worried we're not, but perhaps not for the same reason you think 😬 

> - I'm saying it's enough to have **one** warning regardless of how often `run` is called, instead of having a warning for each `run`.

In my tests this is (sort of) the default behaviour anyhow. In my tests, the default behaviour of the warnings module + Jupyter is that the warning is only shown once. I thought it would be once *per job*, but in my notebook the following gives a *single* warning:

```python
from pyiron import Project
pr = Project('scratch')
pr.remove_jobs_silently(recursive=True)
job = pr.create.job.Lammps('lmp')
job.structure = pr.create.structure.ase_bulk('Al', cubic=True)
job.potential = job.list_potentials()[0]
job.calc_md(temperature=300, langevin=True, n_ionic_steps=100, n_print=20)
job.interactive_open()
job.run()
job.run()
job.interactive_close()
job2 = pr.create.job.Lammps('lmp2')
job2.structure = pr.create.structure.ase_bulk('Al', cubic=True)
job2.potential = job.list_potentials()[0]
job2.calc_md(temperature=300, langevin=True, n_ionic_steps=100, n_print=20)
job2.interactive_open()
job2.run()
job2.interactive_close()
```

> Besides, a super user like me would simply `unfix` and `fix` Langevin with a different seed if they really want to have a new set of random numbers, in which case the problem stated in the warning is not pertinent anymore, but still we would have to see it come every time. So here's what's gonna happen when the first warning is issued:
> 
> Normal user: "Oh I shouldn't be doing this" -> stops using interactive, calc_md or Langevin
> Super user: "Oh I have to redefine seeds" -> includes `fix` + `unfix` in the script.
> Anakin Skywalker: "Oh I have to save my code" -> moves to the dark side (and doesn't even manage to save his code).
> 
> In any of these cases, you just need the first warning. That's what I'm saying.

Either I don't understand what you're saying, or you don't understand what I'm saying. My point is this:

```python
from pyiron import Project
pr = Project('scratch')
pr.remove_jobs_silently(recursive=True)
job = pr.create.job.Lammps('lmp')
job.structure = pr.create.structure.ase_bulk('Al', cubic=True)
job.potential = job.list_potentials()[0]
job.calc_md(temperature=300, langevin=True, n_ionic_steps=100, n_print=1)
job.interactive_open()
job.run()
job.interactive_close()
```

Gives the **wrong physics**. It is not obvious this should be the case. It is *not* the case when `langevin=False`. This is terrible behaviour and should be avoided.

My understanding is that you and @jan-janssen have a different use-case, and you're exploiting the interactive jobs to avoid creating a bunch of job files. Something sort of like this:

```python
from pyiron import Project
pr = Project('scratch')
pr.remove_jobs_silently(recursive=True)
job = pr.create.job.Lammps('lmp')
job.structure = pr.create.structure.ase_bulk('Al', cubic=True)
job.potential = job.list_potentials()[0]
job.calc_md(temperature=300, langevin=True, n_ionic_steps=100, n_print=100)  # steps = print --> A-OK
job.interactive_open()
for _ in range(10):
    **TOTALLY CHANGE STUFF SO WE DON'T HAVE A TEMPORAL RELATION TO THE LAST STEP**
    job.run()
job.interactive_close()
```

I'm fine with that use case, but it should still throw a warning.

I *think* it's OK as long as `n_print==n_ionic_steps`, but the only way to check is to run a huge number of tests since the only data we get is the first and last frames. And even then, it's only "right" as long as you don't expect continuity between multiple `run` calls. Thus, I'm sticking with the safer warning I'm confident in: it's fine when `n_print` goes to infinity.

samwaseda:
> I'm increasingly worried we're not, but perhaps not for the same reason you think 😬

Hmmm, that’s possible but after reading your comments I still don’t know what was the reason I had missed....

> In my tests this is (sort of) the default behaviour anyhow. In my tests, the default behaviour of the warnings module + Jupyter is that the warning is only shown once. I thought it would be once per job, but in my notebook the following gives a single warning:

That’s good. So are you saying it doesn’t matter where to put the warning because it issues it anyway only once? If so, I have a slight preference for `validate_ready_to_run`, because I want to keep `run_if_interactive` as light as possible.

> Either I don't understand what you're saying, or you don't understand what I'm saying. My point is this:
> 

Ok I probably should have said that I would rewrite the input file to change the run command, so that it includes a redefinition of seeds at each step, so my run command would consist of `unfix langevin`, `fix langevin` and `run`. I know (or at least knew) how to do it, but it’s definitely something we don’t want to offer to a normal user.

> I'm fine with that use case, but it should still throw a warning.

I’m also fine with that, but I certainly don’t need it every time I call `run`.

> 
> I _think_ it's OK as long as `n_print==n_ionic_steps`, but the only way to check is to run a huge number of tests since the only data we get is the first and last frames. And even then, it's only "right" as long as you don't expect continuity between multiple `run` calls. Thus, I'm sticking with the safer warning I'm confident in: it's fine when `n_print` goes to infinity.

I’m a little bit confused here. All I’m saying is it’s enough to have one warning and not multiple warnings... can you help me understand the relationship?

liamhuber:
> So are you saying it doesn’t matter where to put the warning because it issues it anyway only once? If so, I have a slight preference for `validate_ready_to_run`, because I want to keep `run_if_interactive` as light as possible.

Yep and done! I'm glad you pushed me on it a bit, I've come fully around to liking having it in `validate...` 

> Besides, a super user like me would simply unfix and fix Langevin with a different seed if they really want to have a new set of random numbers, in which case the problem stated in the warning is not pertinent anymore, but still we would have to see it come every time. 

> Ok I probably should have said that I would rewrite the input file to change the run command, so that it includes a redefinition of seeds at each step, so my run command would consist of `unfix langevin`, `fix langevin` and `run`. I know (or at least knew) how to do it, but it’s definitely something we don’t want to offer to a normal user.

> All I’m saying is it’s enough to have one warning and not multiple warnings...

I don't know how Jupyter + `warnings` is collaborating to make this the case, but indeed there is only one warning! I'm relieved because I wouldn't have known how to elegantly set it up to track whether or not  the warning had already been thrown.

My worry was that I don't understand how using `unfix`+`fix` solves the bad physics, and I was concerned that you were asserting that the physics isn't bad. As long as you're not objecting to the latter, I'm willing accept the former with blissful ignorance 😉

samwaseda:
Ok great so I guess we found a common ground 👍 

Let me explain maybe one more time what I was saying with fix and unfix, because I’m pretty sure that this is a possibility that you could take into account when you encounter the same problem in the future: whenever a new `run` is to be transmitted to LAMMPS, it essentially uses the seed we put to create a new set of random numbers and the corresponding random force fields, right? This seed is written inside `fix langevin`. So if we annul the old seed and set a new one before we transmit `run` to LAMMPS at each run, we wouldn’t have this problem. So in practice, it would look like this:

```
def run_if_interactive(self)
    self.command(‘unfix langevin’)
    self.command(‘fix langevin all langevin {} {} {} {}’.format(temperature, temperature, damp, new_seed))
    self.command(‘run {}’.format(n_steps))
```

liamhuber:
> Ok great so I guess we found a common ground 👍

Yep, I think we're good here! Thanks for all the feedback 😄 

> So if we annul the old seed and set a new one before we transmit `run` to LAMMPS at each run, we wouldn’t have this problem.

Haha, I'll believe this when I see it work! On the other hand, I believe you in general, so if I really need this to work this will be the first thing I try 😉

Issue #1229, opened 2020-11-26
ImportError: SQSJob relies on sqsgenerator.core.sqs.ParallelSqsIterator, but this is unavailable


danjuzam:
When I tried to create an SQS job, an error occured:
> 
> from pyiron import Project
> pr = Project("sqs")
> job_sqs = pr.create_job(
...     job_type=pr.job_type.SQSJob, 
...     job_name="sqscalculation"
... )
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/isi/w/mut/Programme/miniconda3/lib/python3.8/site-packages/pyiron/project.py", line 167, in create_job
    job = JobType(
  File "/isi/w/mut/Programme/miniconda3/lib/python3.8/site-packages/pyiron_base/job/jobtype.py", line 92, in __new__
    job = job_class(project, job_name)
  File "/isi/w/mut/Programme/miniconda3/lib/python3.8/site-packages/pyiron/atomistics/job/sqs.py", line 144, in __init__
    _fail_if_imports_missing()
  File "/isi/w/mut/Programme/miniconda3/lib/python3.8/site-packages/pyiron/atomistics/job/sqs.py", line 64, in _fail_if_imports_missing
    raise ImportError("SQSJob relies on sqsgenerator.core.sqs.ParallelSqsIterator, but this is unavailable.")
ImportError: SQSJob relies on sqsgenerator.core.sqs.ParallelSqsIterator, but this is unavailable.

Any ideas what could be the reason for that?

jan-janssen:
The SQS code is an external tool which can be installed as separate conda package: 
```
conda install -c conda-forge sqsgenerator 
```
https://anaconda.org/conda-forge/sqsgenerator
https://github.com/dgehringer/sqsgenerator

danjuzam:
Ok, thank you. Is there a list where one can see which of the available job_types are included in the pyiron installation and which have to be installed separately (and then how)? That would be very helpful.

jan-janssen:
@danjuzam That is a great idea, here is a first list of codes and how to install them, the rest of the job types are available out of the box, but many of them require a quantum engine to evaluate energies and forces for a given structure. These quantum engines are commonly external codes. Still at least the open source codes can all be installed using conda. 

| code | job_type | How to install ? |
|------|----------|----------------|
| sqsgenerator | SQSJob, ParallelSqsIterator | `conda install -c conda-forge sqsgenerator` | 
[...]

PR #1230, opened 2020-11-26
Ever more detailed error messages

liamhuber:
In response to #1229 

There are a few python packages whose import we wrap in a `try/except`. These should fail as early as possible and with as clear an error message as possible.

I couldn't quickly find a way to wrap the other jobs @jan-janssen nicely collated in #1229 , but this PR could be either merged as-is or extended to give users early warning in even more cases.


liamhuber:
If we start using this sort of `_fail_if_imports_missing` in more places, we should refactor it out. I didn't see an existing `utils` module or any other good home, so for now we just duplicate the code (once).

liamhuber:
So this failure is actually not totally surprising; the trouble comes from the gpaw tests:

```python
1955
ERROR: setUpClass (gpaw.test_gpaw.TestGpaw)
1956
1957
1958
  File "/home/runner/work/pyiron/pyiron/tests/gpaw/test_gpaw.py", line 19, in setUpClass
1959
    job = Gpaw(
1960
  File "/home/runner/work/pyiron/pyiron/pyiron/gpaw/gpaw.py", line 45, in __init__
1961
    _fail_if_imports_missing()
1962
  File "/home/runner/work/pyiron/pyiron/pyiron/gpaw/gpaw.py", line 34, in _fail_if_imports_missing
1963
    raise ImportError(
1964
ImportError: Gpaw relies on the gpaw module but th is unavailable. Please ensure your python environment contains gpaw, e.g. by running `conda install -c conda-forge gpaw`.
1965

1966
```

Basically what's happening is that this fix makes the gpaw job type fail at instantiation instead of runtime (because failing earlier is better than later), but the gpaw tests don't actually run anything! So before this fix, the gpaw tests instantiated a gpaw job, wrote it to hdf, etc, but *never actually used any of the functionality from the `gpaw` library*!

So we have two choices: 
1) Allow instantiating a gpaw job, even when it will fail if you try to run it (or do some other things with it?)
2) Exclude the gpaw tests from the test systems that don't have gpaw available

I am obviously in favour of (2), but I wanted to get feedback...and maybe help with the exclusion setup.

jan-janssen:
> So we have two choices:
> 
> 1. Allow instantiating a gpaw job, even when it will fail if you try to run it (or do some other things with it?)
> 2. Exclude the gpaw tests from the test systems that don't have gpaw available
> 
> I am obviously in favour of (2), but I wanted to get feedback...and maybe help with the exclusion setup.

But that applies to most of our tests, we run VASP tests without VASP installed, LAMMPS test with LAMMPS installed and GPAW tests without GPAW installed. So I am wondering if we could provide a special flag for tests.

liamhuber:
> But that applies to most of our tests, we run VASP tests without VASP installed, LAMMPS test with LAMMPS installed and GPAW tests without GPAW installed. So I am wondering if we could provide a special flag for tests.

Hmm, let's think about it a bit. I think it's important to maintain failure at instantiation for the users, but I see the merit to doing non-running tests even when the modules aren't there. I'm sure there is an elegant solution, but the only patches I can think of ATM are very ham-fisted.

liamhuber:
So I've been thinking about this, and I feel the best solution is to uniformly switch over to giving a warning on the instantiation of an un-runable job instead of throwing an error. This satisfies my desire to have things fail (in the case warn) very early for users, but shouldn't break our tests.

I'll try to get this implemented by the end of the week...

liamhuber:
Ok, they're all just warnings.

I also think I caught all the relevant cases. `LammpsInteractive` was a red herring -- the jobs will run either way and we're just letting Windows machines have a free pass; for the relevant UNIX machines the try/except-wrapped package should automatically be installed when lammps is.

Pending on [pyiron/pyiron_base #116](https://github.com/pyiron/pyiron_base/pull/116)

liamhuber:
@jan-janssen the dependencies for this PR are [merged into pyiron_base](https://github.com/pyiron/pyiron_base/pull/116). Next time we have a base release and get the dependency updated on this repo, I'll merge master into this PR and it's ready to go.

jan-janssen:
@liamhuber I would like to merge https://github.com/pyiron/pyiron_base/pull/121 as it fixes https://github.com/pyiron/pyiron/issues/1267 once this is done I would suggest @pmrv or @niklassiemer or @muh-hassani take the chance and create a new release.

liamhuber:
With #1271 I expect this to now work.

jan-janssen (approved):
Looks good to me

PR #1236, opened 2020-11-30
Use new deprecate decorator

pmrv:
As anticipated there are a lot of instances where we only deprecate some arguments to a function, I will handle this in the next iteration of the decorator in pyiron_base and then update here.

There also some cases that I don't know how to handle, like this

https://github.com/pyiron/pyiron/blob/d410e585babd3b07326853de38134bd182b528bd/pyiron/lammps/interactive.py#L113

It'd be nice to also have them be handled by the `Deprecator`, to keep the messages uniform.  I'm thinking to expose a method there that just logs the warning and this method would be called in the decorator, but it would also be available manually.  If you have opinions on this, let me know.

### DIFF
--- a/pyiron/atomistics/job/interactivewrapper.py
+++ b/pyiron/atomistics/job/interactivewrapper.py
@@ -5,6 +5,7 @@
 from datetime import datetime
 import warnings
 from pyiron_base import GenericParameters, GenericJob, GenericMaster
+from pyiron_base.generic.util import deprecate
 from pyiron.atomistics.structure.atoms import ase_to_pyiron
 from pyiron.atomistics.structure.atoms import Atoms as PAtoms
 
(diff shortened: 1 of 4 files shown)


pmrv:
Force pushed to fix missing import.

liamhuber:
> There also some cases that I don't know how to handle, like this
> 
> 
> 
> https://github.com/pyiron/pyiron/blob/d410e585babd3b07326853de38134bd182b528bd/pyiron/lammps/interactive.py#L113

In this particular case I'd refactor i the names but not used 'set_relative', have it do nothing but pass, call it inside the if clause where the message is currently being printed, and decorate that new function. No need to adapt the Decorator class when we can shuffle away what would currently be really weird calls to it.

liamhuber:
Ah I read all and see my solution differs only slightly from yours. I think we're headed down the same path on this one 🙂

pmrv:
> > There also some cases that I don't know how to handle, like this
> > https://github.com/pyiron/pyiron/blob/d410e585babd3b07326853de38134bd182b528bd/pyiron/lammps/interactive.py#L113
> 
> In this particular case I'd refactor i the names but not used 'set_relative', have it do nothing but pass, call it inside the if clause where the message is currently being printed, and decorate that new function. No need to adapt the Decorator class when we can shuffle away what would currently be really weird calls to it.

Ah, that also works.  If there's no other comments, I'll merge this today and pick this up in the next round with the changes to capture deprecated arguments.

PR #1237, opened 2020-11-30
Switching off k-mesh spacing by k_mesh_spacing=None

yury-lysogorskiy:
Switching off by `k_mesh_spacing=None` would **preserve** the current k-mesh, not reset it, i.e.:
```python
ref_job.set_kpoints(k_mesh_spacing=0.1)
# kmesh=[15,15,15]
ref_job.set_kpoints(k_mesh_spacing=None)
# kmesh remains [15,15,15]
```

### DIFF
--- a/pyiron/dft/job/generic.py
+++ b/pyiron/dft/job/generic.py
@@ -270,6 +270,8 @@ class GenericDFTJob(AtomisticGenericJob):
             if mesh is not None:
                 warnings.warn("mesh value is overwritten by k_mesh_spacing")
             mesh = self.get_k_mesh_by_cell(k_mesh_spacing=k_mesh_spacing)
+        elif self.k_mesh_spacing is not None: # switching off k_mesh_spacing would preserve the current mesh
+            mesh = self.get_k_mesh_by_cell(k_mesh_spacing=self.k_mesh_spacing)
         self.k_mesh_spacing = k_mesh_spacing
         self.k_mesh_center_shift = center_shift
         self.reduce_kpoint_symmetry = symmetry_reduction


samwaseda:
Can you maybe make it a bit clear what was the problem and how it should behave? I don't really understand the changes in the current form.

yury-lysogorskiy:
> Can you maybe make it a bit clear what was the problem and how it should behave? I don't really understand the changes in the current form.

Yes, sure. Before, if one switches off `k_mesh_spacing` setting, then 
```python
ref_job.set_kpoints(k_mesh_spacing=0.1)
# kmesh=[15,15,15]
ref_job.set_kpoints(k_mesh_spacing=None) # switches off `k_mesh_spacing` 
# kmesh becomes [1,1,1] or None
```
But I suggest that  expected behaviour should be that after removing the `k_mesh_spacing`, current value of `kmesh` should remain
```python
ref_job.set_kpoints(k_mesh_spacing=0.1) # switches off `k_mesh_spacing`
# kmesh=[15,15,15]
ref_job.set_kpoints(k_mesh_spacing=None)
# kmesh remains [15,15,15]
```

samwaseda:
> > Can you maybe make it a bit clear what was the problem and how it should behave? I don't really understand the changes in the current form.
> 
> Yes, sure. Before, if one switches off `k_mesh_spacing` setting, then
> 
> ```python
> ref_job.set_kpoints(k_mesh_spacing=0.1)
> # kmesh=[15,15,15]
> ref_job.set_kpoints(k_mesh_spacing=None) # switches off `k_mesh_spacing` 
> # kmesh becomes [1,1,1] or None
> ```
> 
> But I suggest that expected behaviour should be that after removing the `k_mesh_spacing`, current value of `kmesh` should remain
> 
> ```python
> ref_job.set_kpoints(k_mesh_spacing=0.1) # switches off `k_mesh_spacing`
> # kmesh=[15,15,15]
> ref_job.set_kpoints(k_mesh_spacing=None)
> # kmesh remains [15,15,15]
> ```

I'm a bit confused because the behaviour that you mentioned doesn't occur when I test it in the current implementation anyway.

Screenshot 2020-11-30 at 16 08 25

pyiron/dft/job/generic.py
```diff
@@ -270,6 +270,8 @@ def set_kpoints(
             if mesh is not None:
                 warnings.warn("mesh value is overwritten by k_mesh_spacing")
             mesh = self.get_k_mesh_by_cell(k_mesh_spacing=k_mesh_spacing)
+        elif self.k_mesh_spacing is not None: # switching off k_mesh_spacing would preserve the current mesh
+            mesh = self.get_k_mesh_by_cell(k_mesh_spacing=self.k_mesh_spacing)
```
samwaseda, line 274: This would overwrite `mesh` regardless of what was set, so you have to make sure that `mesh` is not `None`

samwaseda (changes requested):

PR #1239, opened 2020-12-01
Remove redundant lines in LAMMPS input control

samwaseda:
remove one redundant self.modify and forbid changing thermo value directly. I'm not really sure if it was okay to delete `thermo=value`, but since we are defining `thermotime` I guess we shouldn't change `thermo` directly.

### DIFF
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -337,8 +337,10 @@ class LammpsControl(GenericParameters):
             + str(int(max_evaluations))
         )
         self.remove_keys(["run", "velocity"])
-        self.modify(variable___dumptime="equal " + str(n_print))
-        self.modify(variable___thermotime="equal " + str(n_print))
+        self.modify(
+            variable___dumptime="equal " + str(n_print),
+            variable___thermotime="equal " + str(n_print)
+        )
 
     def calc_static(self):
         self.set(run="0")
--- a/pyiron/lammps/control.py
+++ b/pyiron/lammps/control.py
@@ -337,8 +337,8 @@ class LammpsControl(GenericParameters):
             + str(int(max_evaluations))
         )
         self.remove_keys(["run", "velocity"])
-        self.modify(variable___dumptime="equal " + str(n_print), thermo=n_print)
-        self.modify(variable___thermotime="equal " + str(n_print), thermo=n_print)
+        self.modify(variable___dumptime="equal " + str(n_print))
+        self.modify(variable___thermotime="equal " + str(n_print))
 
     def calc_static(self):
         self.set(run="0")
@@ -578,14 +578,7 @@ class LammpsControl(GenericParameters):
         self.modify(
             fix___ensemble=fix_ensemble_str,
             variable___dumptime=" equal {} ".format(n_print),
-            thermo=int(n_print),
-            run=int(n_ionic_steps),
-            append_if_not_present=True,
-        )
-        self.modify(
-            fix___ensemble=fix_ensemble_str,
             variable___thermotime=" equal {} ".format(n_print),
-            thermo=int(n_print),
             run=int(n_ionic_steps),
             append_if_not_present=True,
         )


liamhuber:
I'm afraid I don't know how to review this. The changes are small and I don't have any code quality/python concerns...but I'm not really familiar enough with what this part of the code is doing to understand the meaning of these changes. Can you give some more insight or what sort of feedback you want?

pmrv:
I worry that this might break my workflow.  I often want to have much more samples from the thermodynamic properties than structures.  If there's another way to change that I'm flexible, but right now I reach directly into the input file to change `thermotime`.

liamhuber:
> I worry that this might break my workflow. I often want to have much more samples from the thermodynamic properties than structures. If there's another way to change that I'm flexible, but right now I reach directly into the input file to change `thermotime`.

I didn't even realize this was possible. Actually, it would be great to eventually extract this generally so that you can ask for it no matter what code you're using to generate your thermodynamic values (LAMMPS, VASP, etc), even if it only makes things *more efficient* for select codes. At least it will save on memory. I guess this would be helpful for literally all the TILD protocols, yeah @raynol-dsouza ?

samwaseda:
> I worry that this might break my workflow. I often want to have much more samples from the thermodynamic properties than structures. If there's another way to change that I'm flexible, but right now I reach directly into the input file to change `thermotime`.

Actually in the current implementation this should not be possible anyway. But I agree that this is something we should think about.

samwaseda:
> I'm afraid I don't know how to review this. The changes are small and I don't have any code quality/python concerns...but I'm not really familiar enough with what this part of the code is doing to understand the meaning of these changes. Can you give some more insight or what sort of feedback you want?

There are two things I changed: One is the removal of `self.modify` which appears twice for pretty much no reason (or at least I don't know why it should be executed twice). Actually I saw that I was the one who made this mistake around a year ago. The other one is to forbid direct manipulation of `thermo` which comes from `n_print`. So this line:

```
self.modify(variable___thermotime="equal " + str(n_print), thermo=n_print)
```

makes `thermotime` redundant because it changes `thermo ${thermotime}` to `thermo 100` for example if `n_print` is 100 (while `thermotime` is also changed to 100 at the same time). I was not sure whether to rather remove `thermotime`, but then I saw that it's actually used in the averaging algorithm, so I removed `thermo=n_print`.

raynol-dsouza:
>  I guess this would be helpful for literally all the TILD protocols, yeah @raynol-dsouza ?

I am not very sure, tbh. The protocols don't save any of the dumps (unless specified) anyway. But it will definitely speed up the protocols if we can only extract the necessary quantities (positions, forces, energy_pot) every step, and not bother about the others.

samwaseda:
... is there further feedback?

pyiron/lammps/control.py
```diff
@@ -337,8 +337,8 @@ def calc_minimize(
             + str(int(max_evaluations))
         )
         self.remove_keys(["run", "velocity"])
-        self.modify(variable___dumptime="equal " + str(n_print), thermo=n_print)
-        self.modify(variable___thermotime="equal " + str(n_print), thermo=n_print)
+        self.modify(variable___dumptime="equal " + str(n_print))
+        self.modify(variable___thermotime="equal " + str(n_print))
```
jan-janssen, line 341:
```suggestion
        self.modify(
            variable___dumptime="equal " + str(n_print),
            variable___thermotime="equal " + str(n_print)
        )
```

jan-janssen (approved):
Looks good to me

pmrv:
I must admit, that my lammps understanding is bad enough, that I still don't quite understand how this will change the default lammps input file.

What I'm doing is along the lines of
```python
job.calc_md(temperature = T, n_ionic_steps = steps, n_print = steps)
job.input.control['thermo'] = thermo
```
This way I get no structures and thermodynamic information every `thermo` steps.  Does that still work?  If not, is there another way to do it?

samwaseda:
> I must admit, that my lammps understanding is bad enough, that I still don't quite understand how this will change the default lammps input file.
> 
> What I'm doing is along the lines of
> 
> ```python
> job.calc_md(temperature = T, n_ionic_steps = steps, n_print = steps)
> job.input.control['thermo'] = thermo
> ```
> 
> This way I get no structures and thermodynamic information every `thermo` steps. Does that still work? If not, is there another way to do it?

It should work (although again, I guess pyiron doesn't really support this kind of situation...)

pmrv:
> > I must admit, that my lammps understanding is bad enough, that I still don't quite understand how this will change the default lammps input file.
> > What I'm doing is along the lines of
> > ```python
> > job.calc_md(temperature = T, n_ionic_steps = steps, n_print = steps)
> > job.input.control['thermo'] = thermo
> > ```
> > 
> > 
> > This way I get no structures and thermodynamic information every `thermo` steps. Does that still work? If not, is there another way to do it?
> 
> It should work 

Sorry for the hold up, then.

> (although again, I guess pyiron doesn't really support this kind of situation...)

Why?  I haven't had problems.  Everything ends up in the HDF5 correctly, it's just less structures than energy values.

samwaseda:
> > > I must admit, that my lammps understanding is bad enough, that I still don't quite understand how this will change the default lammps input file.
> > > What I'm doing is along the lines of
> > > ```python
> > > job.calc_md(temperature = T, n_ionic_steps = steps, n_print = steps)
> > > job.input.control['thermo'] = thermo
> > > ```
> > > 
> > > 
> > > This way I get no structures and thermodynamic information every `thermo` steps. Does that still work? If not, is there another way to do it?
> > 
> > 
> > It should work
> 
> Sorry for the hold up, then.
> 
> > (although again, I guess pyiron doesn't really support this kind of situation...)
> 
> Why? I haven't had problems. Everything ends up in the HDF5 correctly, it's just less structures than energy values.

It's just certain functionalities simply assume a direct correspondence between energy values and forces etc. But from my point of view, if the user does that it's really up to them to know what they're doing.

pmrv:
> > > > I must admit, that my lammps understanding is bad enough, that I still don't quite understand how this will change the default lammps input file.
> > > > What I'm doing is along the lines of
> > > > ```python
> > > > job.calc_md(temperature = T, n_ionic_steps = steps, n_print = steps)
> > > > job.input.control['thermo'] = thermo
> > > > ```
> > > > 
> > > > 
> > > > This way I get no structures and thermodynamic information every `thermo` steps. Does that still work? If not, is there another way to do it?
> > > 
> > > 
> > > It should work
> > 
> > 
> > Sorry for the hold up, then.
> > > (although again, I guess pyiron doesn't really support this kind of situation...)
> > 
> > 
> > Why? I haven't had problems. Everything ends up in the HDF5 correctly, it's just less structures than energy values.
> 
> It's just certain functionalities simply assume a direct correspondence between energy values and forces etc. But from my point of view, if the user does that it's really up to them to know what they're doing.

That sounds like defect to me then.  For system of a certain size or thermodynamic calculations there's simply no point in saving all the structures.

Anyway, that is a different discussion.  With this out of the way, feel free to merge.

pmrv (approved):

samwaseda:
> 
> That sounds like defect to me then. For system of a certain size or thermodynamic calculations there's simply no point in saving all the structures.

It certainly is. It's just we are too stupid to come up with a good idea.

> 
> Anyway, that is a different discussion. With this out of the way, feel free to merge.

liamhuber (approved):

jan-janssen:
@samwaseda Can you merge this? then we can include it in the next pyiron release.

PR #1240, opened 2020-12-01
Shift analyze related stuff to the new analyze class

niklassiemer:
Shift analyze functions towards the new analyze class introduced by Sam. I omitted `analyse_phonopy_equivalent_atoms` since this is completely obsolete if I am not mistaken. 
I did not (yet) update the tests to check if everything is redirected correctly before going on.

Which functions should be moved to the `analyse` class in addition to the analyze_* functions?


pyiron/atomistics/structure/analyse.py
```diff
@@ -130,4 +130,91 @@ def get_layers(self, distance_threshold=0.01, id_list=None, wrap_atoms=True, pla
...
+        """
+        from pyiron.atomistics.structure.pyscal import analyse_cna_adaptive
+        return analyse_cna_adaptive(atoms=self._structure, mode=mode, ovito_compatibility=ovito_compatibility)
+
+    def pyscal_centro_symmetry(self, num_neighbors=12):
+        """
+        Analyse centrosymmetry parameter
+
+        Args:
+            atoms: Atoms object
```
pmrv, line 182: If use 
```
:class:`.Atoms`
```
or

```
:class:`pyiron.structure.atoms.Atoms`
```

Sphinx turns this into links automatically.

pyiron/atomistics/structure/analyse.py
```diff
@@ -130,4 +130,91 @@ def get_layers(self, distance_threshold=0.01, id_list=None, wrap_atoms=True, pla
...
+            job (job): pyiron job
+            neighbor_method (str) : can be ['cutoff', 'voronoi']
+            cutoff (float) : can be 0 for adaptive cutoff or any other value
+            n_clusters (int) : number of clusters for K means clustering
+            q (list) : can be from 2-12, the required q values to be calculated
+            averaged (bool) : If True, calculates the averaged versions of the parameter
+            clustering (bool) : If True, cluster based on the q values
+
+        Returns:
+            q (list) : calculated q parameters
```
pmrv, line 148: There's no need name return values, just use e.g.
```
list: calculated q parameters
```

pyiron/atomistics/structure/analyse.py
niklassiemer: Thanks for the explanation! Actually, I did not yet have any corrections of the Docstrings included - there are some wich have to much arguments listed...
Is there actually a way to copy Docstrings? For now, I copied Docstrings as they are which is a dublication of code.

pmrv: I figured that you just copied them, but since you're already touching them I thought I point it out. ;)

@samwaseda used something like this

https://github.com/pyiron/pyiron/blob/182f47289332d1b4bba51beef732e9f3841985d7/pyiron/atomistics/structure/atoms.py#L1169

to avoid copying the strings by hand.

niklassiemer: Sure, updating them is on the list :) I will use the copy __doc__ in the deprecated methods.

pyiron/atomistics/structure/analyse.py
```diff
@@ -130,4 +130,91 @@ def get_layers(self, distance_threshold=0.01, id_list=None, wrap_atoms=True, pla
...
+
+        """
+        from pyiron.atomistics.structure.pyscal import get_steinhardt_parameter_structure
```
jan-janssen, line 151: I guess you can import all functions at the top of the module, as `pyscal` is available on all three operation system it should not cause any issue. The only limitation is that it is currently not available as pip package.

niklassiemer: Apparently, this results in circular dependencies.

niklassiemer:
Did I miss some functions to be transferred?

pyiron/atomistics/structure/analyse.py
pmrv: I just had a brief look and couldn't see the issue.  Is there a detailed error?  I guess pyscal shouldn't have an import dependency on the atoms module.

niklassiemer: Well if I shift this upwards some tests fail (da86ce6) see with the error
`ImportError: cannot import name 'pyiron_to_ase' from partially initialized module 'pyiron.atomistics.structure.atoms' (most likely due to a circular import) (/home/runner/work/pyiron/pyiron/pyiron/atomistics/structure/atoms.py)`
In pyscal we import pyiron_to_ase from atoms, in atoms we import analyze, thus, we close the circle by importing pyscal in analyze.

pmrv: I pushed one commit that changes the import statement in the pyscal module.  This seems to fix the problems in my local testing.  There's an explanation [here](https://stackoverflow.com/questions/7336802/how-to-avoid-circular-imports-in-python/37126790#37126790), but I didn't fully understand.  It seems to be something weird about pythons import machinery.

If it doesn't work, feel free to
```
git reset --hard HEAD~
git push --force
```

niklassiemer: I also could just revert it - only the commit history would be messier... Speaking of commit history - is there a nice way to revert the revert?

pmrv:
> Did I miss some functions to be transferred?

I would also suggest to move the following methods
```python
get_symmetry()
symmetrize_vectors()
group_points_by_symmetry()
get_equivalent_voronoi_vertices()
get_equivalent_points()
get_symmetry_dataset()
get_spacegroup()
refine_cell() # this should have a docstring, preferably someone who used this before
get_primitive_cell()
get_ir_reciprocal_mesh() # moving this might inconvenient from DFT codes?
```

I haven't used any of these before, so this is a very uninformed list.  I hope whoever is using these will shout at me.  Most of these seem to bind to spglib, so moving them sounds reasonable.

There's some more I wonder why they are on `Atoms` at all
```python
get_distance()
find_mic()
get_distances_array()
```

pyiron/atomistics/structure/analyse.py
pmrv: The easiest would be an interactive rebase + force push as [here at the bottom](https://github.com/eisenforschung/cm-onboarding/blob/master/Git.md).

niklassiemer: Well, I do not like force push very much. We could cleanup in the end this way if we really want to. Now I just revert the revert...

niklassiemer: Hopefully, this works now on every system! Another revert would really look extremely awkward...

niklassiemer:
Thanks very much for the list albeit its not based on experience! 
@samwaseda could you please also skim over the methods which should be shifted?

samwaseda:
> Thanks very much for the list albeit its not based on experience!
> @samwaseda could you please also skim over the methods which should be shifted?

I saw your changes and I'm already very happy with what has been done!

I just want to say that the suggestions coming from @pmrv are good, but before moving them blindly to the new class, I would like to talk about them in detail, because some of them are frankly super redundant, inflexible, not well documented, inefficient, useless or ill-placed.

niklassiemer:
Thanks a lot for the insight! 
I am happy to take some extra work to get this situation better (also in this PR), so we could briefly discuss tomorrow.

niklassiemer:
As discussed I just added a `deprecate_soon` warning to the functions Marvin mentioned since we will either drop, rework, or move them. 
Thus, I do not introduce a new `Symmetry` sub class in this PR and leave this for the next round after we received complaints about deprecating the important function x, and gathering volunteers for a 
sensible rework.

pyiron/atomistics/structure/atoms.py
```diff
@@ -1016,125 +1017,53 @@ def repeat_points(self, points, rep, centered=False):
     def reset_absolute(self, is_absolute):
         raise NotImplementedError("This function was removed!")
 
+    @deprecate("use analyse.pyscal_cna_adaptive() with ovito_compatibility=True instead.")
```
samwaseda, line 1020: Shouldn't it be `Atoms.analyse.pyscal_cna_adaptive()` ? I'm not so sure about the convention though...

pmrv (approved):
lgtm

pyiron/atomistics/structure/atoms.py
```diff
@@ -21,7 +21,8 @@
     ChemicalElement
 )
 from pyiron_base import Settings
-from pyiron_base.generic.util import deprecate
+from pyiron_base.generic.util import deprecate, deprecate_soon
```
jan-janssen, line 24: To keep the interface clean I would suggest to import directly from `pyiron_base`: 
```
from pyiron_base import deprecate, deprecate_soon
```
As implemented in https://github.com/pyiron/pyiron_base/pull/122

pyiron/atomistics/structure/pyscal.py
```diff
@@ -4,7 +4,7 @@
 
 import numpy as np
 from pyiron_base import Settings
-from pyiron.atomistics.structure.atoms import pyiron_to_ase
+import pyiron.atomistics.structure.atoms
```
jan-janssen, line 7: Why is this change required?

pyiron/atomistics/structure/atoms.py
niklassiemer: I an also not aware of a convention. I will switch to your version though.

pyiron/atomistics/structure/pyscal.py
niklassiemer: Otherwise we end up in the circular dependencies:
In atoms we import analyze, in analyze we import pyscal, and if we also import  a function from atoms here, we close the cycle. https://github.com/pyiron/pyiron/pull/1240#discussion_r535205748 https://github.com/pyiron/pyiron/pull/1240#discussion_r535222360

niklassiemer:
I reactivated an older ovito test to use pyscal, however, the outcome of the pyscal function is different. I adopted the test to have the new value ( c6e825e). Could someone please verify that this result is correct and I am not testing to be consistently wrong?!

jan-janssen:
> I reactivated an older ovito test to use pyscal, however, the outcome of the pyscal function is different. I adopted the test to have the new value ( [c6e825e](https://github.com/pyiron/pyiron/commit/c6e825edc34a47d27753089c70be4577501daa11)). Could someone please verify that this result is correct and I am not testing to be consistently wrong?!

There are two parts to it, one make sure you have to `pyscal` version defined in our dependencies and two the pyscal functions have a `ovito_compatibility` flag, so in case you call `pyscal` directly you have to add this.

jan-janssen (approved):
Looks good to me - ready to be merged.

niklassiemer:
I won't merge anything as long as the tests are failing. However, I do not understand what I am missing here. Could you please have a look @jan-janssen or @pmrv ? Thanks

pmrv:
Tests fail because `sklearn.neighbors.kde` is not a module in new scikit versions.  It seems @jan-janssen fixed this in #1277.

niklassiemer:
Thanks a lot! I will merge once all tests are passed.

PR #1243, opened 2020-12-02
Fix the failing pipeline notebook

jan-janssen:
For now I just disable interactive mode


jan-janssen:
It fails for lammps `2020.10.29` but works fine for lammps `2020.03.03`

PR #1250, opened 2020-12-04
Density of state implemented in SPHInX

samwaseda:
@yuyuanjingxuan requested DOS in SPHInX. For Fe I got the following results (the difference probably comes from the dos smearing, but I'm not 100% sure):

```
for name, job in zip(['SPHInX', 'VASP'], [spx_Fe, vasp_Fe]):
    dos = job['output/electronic_structure/dos/tot_densities']
    energies = job['output/electronic_structure/dos/energies']
    efermi = job['output/electronic_structure/efermi']
    plt.plot(energies-efermi, dos[0], label='{} Spin up'.format(name))
    plt.plot(energies-efermi, -dos[1], label='{} Spin down'.format(name))
plt.grid()
plt.ylabel('DOS')
plt.xlabel('Energy eV')
plt.legend();
```

Unknown


freyso:
Sorry - the "from-eps.dat" mode always assumes spin non-polarized calculations (cause you cannot in read up and down and the same time), and therefore has an extra factor 2. (as expected for spin non-polarized)
But that's probably easy to build in, because you have to make the case separation (1 vs 2 spin channels) anyway because of the different filenames...

samwaseda:
Beautiful! (Thanks @freyso)

Unknown

(Now the difference comes only from the number of empty states, since we have a much higher value for SPHInX)

ashtonmv (approved):

samwaseda:
I'm really super sorry to those who lost time reviewing this PR, but since there was much manipulation required for the current implementation of pyiron to use the SPHInX internal algorithm, I decided to calculate DOS through pyiron. I'm especially super sorry, @freyso, for stealing your time...

PR #1259, opened 2020-12-13
Fix delete

sudarsan-surendralal:
Deleting tags properly when passing boolean arguments

### DIFF
--- a/pyiron/atomistics/structure/atoms.py
+++ b/pyiron/atomistics/structure/atoms.py
@@ -2159,7 +2159,10 @@ class Atoms(ASEAtoms):
         if isinstance(key, (int, np.integer)):
             key = [key]
         key = np.array(key).flatten()
-        new_length = len(self) - len(key)
+        if isinstance(key[0], (bool, np.bool_)):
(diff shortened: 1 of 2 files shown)


pyiron/atomistics/structure/atoms.py
```diff
@@ -2159,7 +2159,10 @@ def __delitem__(self, key):
         if isinstance(key, (int, np.integer)):
             key = [key]
         key = np.array(key).flatten()
-        new_length = len(self) - len(key)
+        if isinstance(key[0], (bool, np.bool_)):
```
samwaseda, line 2162: I'm fine with this implementation, but if you want to reduce the number of lines, you could also do

```
new_length = len(self) - len(np.arange(len(key))[np.asarray(key)])
```

samwaseda (approved):
Thanks @sudarsan-surendralal for taking care of this!

pyiron/atomistics/structure/atoms.py
sudarsan-surendralal: This is much nices (although a small change is required `new_length = len(self) - len(np.arange(len(self))[np.asarray(key)])`)

Issue #1262, opened 2020-12-15
sphinx can nit copied specific part from an old project 


mostafajoulaian:
I use the latest version of pyiron which is installed on the cluster.
I try to recover or coping some important setting from an old project (such as energy cut off, K point sampling, etc.). and try to run Murnaghan calculation. (for sphinx)
 It works very well, but what is not work is setting up a sphinx reference job with all the settings I want to make. it is deeper than the sphinx interior. Such as checking overlap.
All setting is set not in generic part, but sphinx specific part is not copied. In conclusion, no matter what you are doing in a reference job, it will never appear in the job is ultimately run.
Here is the piece of code 

def launch_sphinx_vasp (old_project):
    old_ref=old_project.load("strain_1_0")
    elem=old_ref.initial_structure.elements[0].symbol
    if isinstance(old_ref.input.KpointFolding,Group):
        kpoints=  [k for k in old_ref.input.KpointFolding.values()]
    else:
        kpoints=old_ref.input.KpointFolding 
    job_name=elem +  '_k' + 'x'.join([str(k) for k in kpoints])
    sphinx_vasp_ref = new_project.create_job(new_project.job_type.Sphinx, job_name +'_sphinx_vasp')
    sphinx_vasp_ref.structure = old_ref.initial_structure
    sphinx_abinit_ref.calc_static()
    #print(sphinx_vasp_ref.input.sphinx.to_sphinx())
    sphinx_vasp_ref.input['KpointCoords'] = [0.0, 0.0, 0.0]
    sphinx_vasp_ref.set_kpoints(kpoints)
    sphinx_vasp_ref.set_encut(old_ref.encut)
    sphinx_vasp_ref.input['EmptyStates'] = old_ref.input['EmptyStates']
    sphinx_vasp_ref.input['KJxc'] = True 
    sphinx_vasp_ref.server.queue = 'cm'
    sphinx_vasp_ref.server.cores = old_ref.server.cores
    sphinx_vasp_ref.set_occupancy_smearing(width=0.1)
    sphinx_vasp_ref.input.sphinx.pawPot.species={"checkOverlap":0}
    
    
    ########
    murn_sphinx_vasp = new_project.create_job(new_project.job_type.Murnaghan, job_name+str("_sphinx_vasp_murn"))
    murn_sphinx_vasp.input['vol_range'] = 0.06
    murn_sphinx_vasp.input['num_points'] = 7
[...]

PR #1263, opened 2020-12-15
restore set_check_overlap

samwaseda:
Problem raised [here](https://github.com/pyiron/pyiron/issues/1262)

### DIFF
--- a/pyiron/sphinx/base.py
+++ b/pyiron/sphinx/base.py
@@ -431,7 +431,7 @@ class SphinxBase(GenericDFTJob):
             else:
                 raise ValueError('Potential must be JTH or VASP')
         if not check_overlap:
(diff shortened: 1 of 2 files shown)


pyiron/sphinx/base.py
```diff
@@ -431,7 +431,7 @@ def load_species_group(self, check_overlap=True, potformat='VASP'):
             else:
                 raise ValueError()
         if not check_overlap:
-            self.input.sphinx.pawPot["species"][-1]["checkOverlap"] = "false"
+            self.input.sphinx.pawPot["species"]["checkOverlap"] = "false"
```
freyso: No. "checkOverlap" resides in the species group (because it overrides the check on a per PAW pot basis). I believe the original version is correct here.

jan-janssen (approved):
Looks good to me

PR #1265, opened 2020-12-15
update structure setter for interactive jobs

samwaseda:
I wanted to make sure that also for interactive jobs the tags are copied. But then I realised that the `get_structure` in `atomistic.py` has little utility, and moreover in the interactive jobs the structure is copied at each step, which significantly reduces the efficiency. So I removed `get_structure` from `atomistic.py` (which makes you use only the one defined in `GenericAtomisticJob`) and introduced `_update_previous_structure`, which only updates the information and does not copy an entire structure.

### DIFF
--- a/pyiron/atomistics/job/atomistic.py
+++ b/pyiron/atomistics/job/atomistic.py
@@ -629,8 +629,8 @@ class AtomisticGenericJob(GenericJobCore):
         Returns:
             pyiron.atomistics.structure.atoms.Atoms: The required structure
         """
-        if not (self.structure is not None):
-            raise AssertionError()
+        if self.structure is None:
+            raise AssertionError('Structure not set')
         snapshot = self.structure.copy()
         conditions = list()
         if isinstance(self.output.cells, (list, np.ndarray)):
--- a/pyiron/atomistics/job/atomistic.py
+++ b/pyiron/atomistics/job/atomistic.py
@@ -637,7 +637,7 @@ class AtomisticGenericJob(GenericJobCore):
             snapshot.cell = self.output.cells[iteration_step]
         except IndexError:
             if wrap_atoms:
-                raise IndexError('cell at step ', iteraction_step, ' not found')
+                raise IndexError('cell at step ', iteration_step, ' not found')
             snapshot.cell = None
         try:
             snapshot.indices = self.output.indices[iteration_step]
--- a/pyiron/atomistics/job/atomistic.py
+++ b/pyiron/atomistics/job/atomistic.py
@@ -622,6 +622,7 @@ class AtomisticGenericJob(GenericJobCore):
         """
         Gets the structure from a given iteration step of the simulation (MD/ionic relaxation). For static calculations
         there is only one ionic iteration step
+
         Args:
             iteration_step (int): Step for which the structure is requested
             wrap_atoms (bool): True if the atoms are to be wrapped back into the unit cell
@@ -632,15 +633,20 @@ class AtomisticGenericJob(GenericJobCore):
         if self.structure is None:
             raise AssertionError('Structure not set')
         snapshot = self.structure.copy()
-        snapshot.cell = self.output.cells[iteration_step]
+        try:
+            snapshot.cell = self.output.cells[iteration_step]
+        except IndexError:
+            if wrap_atoms:
+                raise IndexError('cell at step ', iteraction_step, ' not found')
+            snapshot.cell = None
         try:
             snapshot.indices = self.output.indices[iteration_step]
         except IndexError:
             pass
         if wrap_atoms:
             snapshot.positions = self.output.positions[iteration_step]
-            return snapshot.center_coordinates_in_unit_cell()
-        if len(self.output.unwrapped_positions) > max([iteration_step, 0]):
+            snapshot.center_coordinates_in_unit_cell()
+        elif len(self.output.unwrapped_positions) > max([iteration_step, 0]):
             snapshot.positions = self.output.unwrapped_positions[iteration_step]
         else:
             snapshot.positions += self.output.total_displacements[iteration_step]
--- a/pyiron/atomistics/job/atomistic.py
+++ b/pyiron/atomistics/job/atomistic.py
@@ -632,25 +632,19 @@ class AtomisticGenericJob(GenericJobCore):
         if self.structure is None:
             raise AssertionError('Structure not set')
         snapshot = self.structure.copy()
-        try:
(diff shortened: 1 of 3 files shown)


pyiron/atomistics/job/atomistic.py
```diff
@@ -629,8 +629,8 @@ def get_structure(self, iteration_step=-1, wrap_atoms=True):
         Returns:
             pyiron.atomistics.structure.atoms.Atoms: The required structure
         """
-        if not (self.structure is not None):
```
samwaseda, line 632: Is this the same as `self.structure is None`?

pyiron/atomistics/job/atomistic.py
```diff
@@ -629,34 +629,28 @@ def get_structure(self, iteration_step=-1, wrap_atoms=True):
...
-            snapshot.positions = self.output.positions[iteration_step]
-        indices = self.output.indices
-        if indices is not None and len(indices) > max([iteration_step, 0]):
-            snapshot.indices = indices[iteration_step]
-        if wrap_atoms:
-            return snapshot.center_coordinates_in_unit_cell()
-        else:
+        except IndexError:
+            pass
+        try:
```
liamhuber, line 636: Stylistically I would prefer a line of whitespace separating repeated independent try/except clauses (same as my preference for independent if/else clauses), i.e.

```python
try:
    foo()
except TypeError:
    pass
except IndexError:
    bar()

try:
    baz()
except IndexError:
    return 3

**Comment by samwaseda** [comment]
By the way, I have to admit that this is a very controversial PR - the new `get_structure` silently returns a structure with initial values if the requested iteration step is not found. However, the same thing was also happening before (for positions, indices etc.) and some of the functionalities inside pyiron (such as random atomistic) were created based on the assumption that this returns a snapshot.

File: pyiron/atomistics/job/atomistic.py [hunk 3]
```diff
@@ -629,34 +629,28 @@ def get_structure(self, iteration_step=-1, wrap_atoms=True):
...
+            snapshot.indices = self.output.indices[iteration_step]
+        except IndexError:
+            pass
+        try:
+            if wrap_atoms:
+                snapshot.positions = self.output.positions[iteration_step]
+                return snapshot.center_coordinates_in_unit_cell()
```
liamhuber: I have a mild stylistic preference to omit the `return` here so this clause is in line with the other two. `center_coordinates_in_unit_cell` looks like it will work in-place so I don't think we need a `snapshot = ...` but just to remove `return`.

pyiron/atomistics/job/interactive.py
```diff
@@ -270,65 +268,26 @@ def interactive_time_getter(self):
...
-            else:
-                return None
+        try:
+            indices = self.output.indices[-1]
+            positions = self.output.positions[-1]
+            cell = self.output.cells[-1]
+        except IndexError:
+            return
+        if len(self._interactive_species_lst) == 0:
+            el_lst = [el.Abbreviation for el in self.structure.species]
```
liamhuber, line 284: I thought that this was already available in `self.structure.get_chemical_symbols()`, but the implementation there is slightly different and uses the indices. I suspect that's redundant though, and if so you should just use the method instead of manually constructing the list.

pyiron/atomistics/job/interactive.py
```diff
@@ -270,65 +268,26 @@ def interactive_time_getter(self):
     def interactive_volume_getter(self):
         return self.initial_structure.get_volume()
 
-    def get_structure(self, iteration_step=-1, wrap_atoms=True):
```
liamhuber, line 273: I don't have any objection to this, but I also don't understand what the ramifications are. So don't feel any need to change anything on my account, but know that I am not wisely and with great understanding approving this change 😉

liamhuber (approved):
There are some stylistic changes I would prefer but do not insist on.

I also think that while you're here and understand this it would be nice to add a test. I guess either [this test](https://github.com/pyiron/pyiron/blob/da28459af5070c8496e6faf8b17466183329c1d0/tests/atomistics/job/test_atomistic.py#L39) can be updated to somehow use `_update_previous_structure`? Even better IMO would be a new test where you manually push different bits of data to the output, call `_update_previous_structure`, and then make sure `structure` is what you expect. This too is requested rather than demanded, so overall I simply approve.

pyiron/atomistics/job/interactive.py
samwaseda: I don't know exactly how it's implemented, but pyiron has a quite complex convention when it comes to symbols/species/elements/indices. Some of them simply list elements, others per-atom elements, and sometimes they also account for original elements and not the user-defined elements etc. I took a brief look at different functions but it looks like the safest way is really to keep what was already in place.

pyiron/atomistics/job/interactive.py
samwaseda: Yeah I also don't really know what could happen, but what's already happening is bad enough (i.e. depending on the `run_mode`, `get_structure` was returning different things). Now there's only one `get_structure` in `atomistic.py`, which I hope will simply it.

pyiron/atomistics/job/atomistic.py
samwaseda: Done!

pyiron/atomistics/job/atomistic.py
samwaseda: Hmmmmm since having line breaks makes it more difficult for vi, I might prefer not to introduce new ones...

samwaseda:
> I also think that while you're here and understand this it would be nice to add a test. I guess either [this test](https://github.com/pyiron/pyiron/blob/da28459af5070c8496e6faf8b17466183329c1d0/tests/atomistics/job/test_atomistic.py#L39) can be updated to somehow use `_update_previous_structure`? Even better IMO would be a new test where you manually push different bits of data to the output, call `_update_previous_structure`, and then make sure `structure` is what you expect. This too is requested rather than demanded, so overall I simply approve.

I added a test!

pmrv (approved):
I'm not sure I can give a qualified opinion, but seems ok.

pyiron/atomistics/job/atomistic.py
liamhuber: dinosaur

😉

pyiron/atomistics/job/interactive.py
liamhuber: I 100% accept that rational for this PR, but because of this knowledge I just threw up a little in my mouth.

PR #1266, opened 2020-12-16
Use deprecator

pmrv:


samwaseda (approved):
Looks good to me (although to be honest I also don't mind getting rid of `get_final_structure` which has its own deprecation warning already for ages)

pmrv:
Am open to throwing it out, I'll discuss with the others on Friday.

pyiron/atomistics/job/atomistic.py
```diff
@@ -10,6 +10,7 @@
 
 from pyiron.atomistics.structure.atoms import Atoms
 from pyiron_base import GenericParameters, GenericMaster, GenericJob as GenericJobCore
+from pyiron_base.generic.util import deprecate
```
jan-janssen, line 13: I would like to import it directly from `pyiron_base` this would allow us to change the structure in `pyiron_base` independent of changing `pyiron` and at the same time we can make sure that only the top level functions are used in other modules.

jan-janssen: Addressed in https://github.com/pyiron/pyiron_base/pull/122

pmrv: Fill fix once we update the version of pyiron_base.

pmrv:
> Looks good to me (although to be honest I also don't mind getting rid of `get_final_structure` which has its own deprecation warning already for ages)

So we decides to keep it for now and throw out deprecated function on the next major update.

jan-janssen (approved):
Looks good to me

PR #1271, opened 2020-12-18
Update to pyiron_base>=0.1.41 and use pip check in tests

pmrv:
I made a mistake and pushed this directly on master by accident.  I reverted it and then reverted the revert to create this PR, hope it didn't break anything in between.


jan-janssen (approved):
Looks good to me

PR #1287, opened 2021-01-08
Fix import from pyiron_base

jan-janssen:


jan-janssen:
Once https://github.com/pyiron/pyiron_base/pull/130 is merged, then the import of the import error can be fixed as well.

PR #1288, opened 2021-01-08
pyiron base update 

jan-janssen:


jan-janssen:
Fixed in https://github.com/pyiron/pyiron/pull/1286

PR #1313, opened 2021-02-26
Import all modules starting with pyiron_*

jan-janssen:
```
from pyiron_base import Project
pr = Project(".")
print(dir(pr.job_type))
import pyiron
print(dir(pr.job_type))
```

### DIFF
--- a/pyiron/__init__.py
+++ b/pyiron/__init__.py
@@ -13,6 +13,8 @@ create_structure = _StructureFactory.crystal
 
 # Fix modules for backwards compatibility
 import sys
+import pkgutil
+import importlib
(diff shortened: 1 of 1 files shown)


liamhuber (approved):
I guess we'll need to edit the exclusion list once pyiron is no longer an alias for pyiron_atomistics, but right now this is perfect.

Can we add your demo snippet to the test suite?

jan-janssen:
> Can we add your demo snippet to the test suite?

It is difficult to write a test without depending on one of the add-on modules like `pyiron_contrib`, `pyiron_gpl`, `pyiron_continuum` or `pyiron_experimental`.

PR #1366, opened 2021-08-23
Add function to get versions of all pyiron_modules

niklassiemer:


jan-janssen:
Can this be combined with https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/project/generic.py#L481 ?

niklassiemer:
Hmm... Combining is rather hard since pyiron_base should not depend on pyiron. We could, however, easily define the same package lookup and version collection as a method/property on pyiron_base.Project

jan-janssen:
If the lookup is only called when the function is called rather than calling it during the import then I am fine with adding it to pyiron_base.

niklassiemer:
I added a version lookup to `get_repository_status` in https://github.com/pyiron/pyiron_base/pull/412.

niklassiemer:
Do we want to provide this `from pyiron import get_pyiron_versions` in addition or is a lookup via the project sufficient?

PR #1371, opened 2021-09-07
Add backlinks to pyiron_{base,atomistics}

pmrv:
As we discussed yesterday, we were missing some links to the detailed docs of atomistics and base.


liamhuber (approved):
lgtm

PR #1394, opened 2022-02-02
Update Readthedocs

jan-janssen:


jan-janssen:
The tqmd progress bar uses none unicode characters, therefore it can not be converted to Latex: 
```
! Package inputenc Error: Unicode character █ (U+2588)
(inputenc)                not set up for use with LaTeX.

See the inputenc package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.1935 ...�███| 5/5 [00:00<00:00, 27.02it/s]
``` 
So I disabled the build of the Latex handbook for now.

Issue #1446, opened 2022-12-21
Creating a custom job


jkitchin:
**Summary**

I want to create a custom job that uses ase.calculators.emt.EMT to calculate the energy of a bulk structure. This is just for a tutorial to show how pyiron works. The end goal is to use this to calculate an EOS for a few metals in a few crystal structures. Ideally, this would end up as a new example notebook for setting up a custom calculation, loop over some metals and structures, and show how to extract the data for analysis afterwards.

I followed the pythonjobtemplate.ipynb in the Docker image and made this.:

```
from pyiron_base import PythonTemplateJob
from ase.calculators.emt import EMT

class EMTJob(PythonTemplateJob):
    def __init__(self, project, job_name):
        super(EMTJob, self).__init__(project, job_name) 
        
    # This function is executed 
    def run(self):
        self.atoms.set_calculator(EMT())
        with self.project_hdf5.open("output/generic") as h5out: 
            h5out["energy"] = self.atoms.get_potential_energy()
        self.status.finished = True
```

Then this works ok:

```
from pyiron import Project
pr = Project('emt')

from ase.build import bulk
atoms = bulk('Al')

job = pr.create_job(job_type=EMTJob, job_name="Al_1")
job.atoms = atoms
job.run()
job['output/generic/energy']
```

I get an energy output that makes sense, and I can see that emt/Al_1.h5 was created.

`pr.job_table()` is empty though.

and it seems I can't use  the job in a Murnaghan task

```
job = pr.create_job(job_type=EMTJob, job_name="Al_2")
job.atoms = atoms

murn = pr.create_job(job_type=pr.job_type.Murnaghan, job_name='murnaghan')
murn.ref_job = job
murn.run()
```

fails with `AttributeError: murnaghan tried to find child job structure, but getattr failed to find the item.`. I guess that means something isn't defined right in the EMTJob class.

If there is a better place to ask these kinds of questions let me know. Thanks!

liamhuber:
Hi @jkitchin, welcome to pyiron!

First: I had never thought of just slapping an ASE calculator onto a structure and making it a job -- that's absolutely brilliant. Pyiron team, I think we can make a whole family of jobs very quickly with this!

[...]

PR #1447, opened 2022-12-22
Add notebook showing how to setup an EMT job and use it with pyiron

jkitchin:
See https://github.com/pyiron/pyiron/issues/1446 for context.


liamhuber (commented):
Hi @jkitchin, looks good! I found all the extra material between the cells easy to read, and the pyiron table at the end is a nice touch. The comments and docstrings retain some of my brutal honesty about the current hiccups in pyiron, but IMO that's ok to leave in 😂 

Please change `from ase.build import bulk` and future calls of `bulk` with the project-based call: `pr.atomistics.structure.bulk`. You convert to pyiron `Atoms` inside the job anyhow (nice touch, btw), and since it's a pyiron tutorial I'd like to get people used to creating structures pyironically. In this case we are just wrapping ASE's bulk call, but we have other structure creators (e.g. for GBs), so it's nice if people start to see where in pyiron to look when building new structures.

Optionally, we might consider putting the volume on the same footing as energy in the new class, i.e. in `run_static` add `self.output.volume = self.structure.get_volume()` right after setting the energy, and then later modifying the bit where we duplicate the result in HDF for Murn's sake to `h5out["volume"] = [self.output.volume]`. It's not strictly necessary, but I think it adds some nice symmetry and consistency to the output.

jkitchin:
I think I made those changes in [e268ae1](https://github.com/pyiron/pyiron/pull/1447/commits/e268ae17e46b800f4f41c10de095b24c9c810776).

notebooks/emt-pyiron.ipynb
```diff
@@ -0,0 +1,2634 @@
...
+    "        # The template job comes with `input` and `output` attributes that will  \n",
+    "        # handle all the (de)serialization for you\n",
+    "        self.input.atoms = pr.atomistics.structure.bulk('Cu')  # Or whatever default\n",
+    "        # self.output.energy = None  # Let's just populate the field for now\n",
+    "        \n",
+    "    @property\n",
+    "    def structure(self):\n",
+    "        \"\"\"\n",
+    "        'structure' is the canonical name for an `Atoms` object in pyiron.\n",
+    "        The Murnhaghan job is expecting(?) a `structure` attribute on its input jobs.\n",
```
pmrv, line 71:
```suggestion
    "        The Murnhaghan job is expecting a `structure` attribute on its input jobs.\n",
```

That's correct!

jkitchin: I changed this.

liamhuber:
I can't figure out how to change my review status to "approve" here on mobile, but I do. I started the tests and as long as we don't have any surprises there, I'm happy to merge. Nice contribution!

jkitchin:
No hurry for me.

jkitchin:
I think I did what you requested.

jan-janssen:
Everything looks alright, but there seems to be an issue with our conda environment. Somehow the dependencies are not resolved correctly, this is not related to this pull request still I would like to fix it before we merge this.

jan-janssen:
@jkitchin Can you once merge with the main branch? We had some issues with the resolving of the dependencies, but those should be solved now.

PR #1468, opened 2023-04-26
quickstart installation

ligerzero-ai:
New quick-install instructions

### DIFF
--- a/docs/source/installation_quickstart.rst
+++ b/docs/source/installation_quickstart.rst
@@ -1,0 +1,132 @@
+.. _installation:
+
+============
+Install pyiron on your cluster in 10 minutes!
+============
+
+1. First, fetch mambaforge install script from the web:
+
+.. code-block:: bash  
+    wget https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-Linux-x86_64.sh
+
+2. Execute the script to install mambaforge:
+.. code-block:: bash
+    bash Mambaforge-Linux-x86_64.sh
+
+.. note:: 
+    NOTE: There are systems with a very tight filequota/memory quota on the home directory. In that case, you may need to install on a different directory. Usually there is a /software or /group directory that users can have permanent storage for software on. 
+
+3. Refresh/restart your shell:
+.. code-block:: bash
+    source .bashrc
+
+4. Now you have the option of installing pyiron in an environment with:
+.. code-block:: bash
+    mamba create -n YOURENVNAME
+Change `YOURENVNAME` to your liking.
+
+5. Then activate your environment with:
+.. code-block:: bash
+    mamba activate YOURENVNAME
+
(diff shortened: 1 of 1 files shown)


docs/source/installation_quickstart.rst
```diff
@@ -0,0 +1,134 @@
...
+    mkdir /home/abc123/pyiron_resources
+
+8. Now, create our pyiron configuration file, `.pyiron` in the home folder. Paste the following lines into the file:
+.. code-block:: bash
+    [DEFAULT]
+    RESOURCE_PATHS = /home/abc123/pyiron_resources, /software/abc123/mambaforge/envs/pyiron/share/pyiron
+    PROJECT_CHECK_ENABLED = False
+    #DISABLE_DATABASE = True
+    FILE = ~/pyiron.db
+```
```
niklassiemer, line 52: Is thin not already automated using the install_dialog function?

And pyiron should automatically add the approptiate directories in the `$CONDAHOME` to the resources. From there on you could go on to configure :)

docs/source/installation_quickstart.rst
```diff
@@ -0,0 +1,134 @@
...
+The first is the directory we just made. The second is where pyiron's environment is located on the filesystem. You can find where it is using `which python` with the environment activated, which yields something like:
+`/scratch/pawsey0380/cxc562/mambaforge/bin/python`
+And you can replace the `bin/…` bit onwards with `envs/YOURENVNAME/share/pyiron`
+
+9. Now enter the `pyiron_resources` folder and make the `queues` folder:
+.. code-block:: bash
+    cd /home/abc123/pyiron_resources
+    mkdir queues
+
+Configure the queue on your supercomputer. Edit/create a queue.yaml file in the `queues` folder, with contents of:
```
niklassiemer, line 68:
```suggestion
Configure the queue on your supercomputer. Edit/create a queue.yaml file in the `queues`, for SLURM setup with contents of:
```

niklassiemer (commented):
Reads quite well, a few comments and we should probably streameline with the other install docs.

docs/source/installation_quickstart.rst
ligerzero-ai: In my experience (installed like 5x over the last week), a fresh install from mamba according to the instructions up to this point does not trigger an automatic generation of the `.pyiron` file. It does require the user to write the file in the home directory.

ligerzero-ai: Is the `install_dialog` function triggered upon loading pyiron modules for the first time? Or is it something that needs to be triggered on the python side? At any rate I don't think it does the user harm to know about the `.pyiron` configuration file even at the beginning.

docs/source/installation_quickstart.rst
ligerzero-ai: addressed in new commit

PR #1479, opened 2023-05-04
Update energy-volume curve tutorial

samwaseda:
Based on [this discussion](https://github.com/pyiron/pyiron_atomistics/issues/1014), I updated the energy-volume curve notebook. In particular, I removed all the manual part in the beginning, because it appears extremely technical, especially only to be replaced by the pyiron Murnaghan module anyway. Instead, I complemented the energy cutoff convergence using Murnaghan, which is probably more important for those who do real physics.


notebooks/energy_volume_curve.ipynb
```diff
@@ -30,14 +30,28 @@
     "ExecuteTime": {
```
liamhuber, line 30: <p>I'm afraid I don't see how to do a github "suggestion" in NB Review, so these will all be old-school comments!</p><p> </p><p>Nit: "allows to [verb]" is a deutschism. "Allows" is a super funny verb that requires an extra noun to work when(ever? I think. Proof by lack of imagination) combined, i.e. "seatbelts allow drivers to not die in low-speed crashes", "the energy volume curve allows us to calculate...", etc. I don't know all the grammatical terms, but <a href="https://english.stackexchange.com/questions/85069/is-the-construction-it-allows-to-proper-english" target="_blank">here</a> is an interesting english.stackexchange commentary on the topic.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyiron/pyiron/pull/1479/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

liamhuber, line 30: <pre class="ql-syntax" spellcheck="false">Line #1.    pr = Project(path='thermo')
</pre><p>Nit: I would omit the keyword, especially because you explicitly call it the "name" above (which I think is a fair term, we don't need to be 100% consistent on that, it's just mildly jarring to see both terms so close together)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyiron/pyiron/pull/1479/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

liamhuber, line 30: <p>Gpaw is not a mandatory requirement for the installation, so it would be good to remind readers how to install it, either by pointing to the install link or explicitly mentioning <code>conda install -c conda-forge gpaw</code></p><p> </p><p>Nit: Since "module" has a technical meaning in python, I would prefer to use "the Murnaghan job can be used"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyiron/pyiron/pull/1479/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

liamhuber, line 30: <p>Nit: Explain why we're changing the input, i.e. to make it run faster</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyiron/pyiron/pull/1479/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

liamhuber, line 30: <p>Great re-direct to related docs! But this is just pointing to the docs main page...the reader then needs to dig through and find the particular topic themselves? I think <a href="https://pyiron.readthedocs.io/en/latest/source/installation_remote.html" target="_blank">this</a> is a better page to point to</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyiron/pyiron/pull/1479/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

liamhuber, line 30: <p>Nit: build in -&gt; built-in</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pyiron/pyiron/pull/1479/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

samwaseda: I kept this cell for no particular reason, but to be honest I find the way it shows the data super clumsy, inconsistent with other jobs and somewhat unrelated to the storyline, so I might remove it altogether.

samwaseda: I might prefer it to be shown in the error message, if Gpaw is used without being installed. But I don't have a very strong feeling, so I can also add it in the notebook.

samwaseda: ok I just realized that I didn't even write this text XD

samwaseda: I removed it in the end.

liamhuber: That's a good point. If it's already in the error message then no further action is required, otherwise touching up the gpaw failure message in this PR would be good 👍

liamhuber: <p>Then I would just soften the language to something like “can be found i the pyiron docs, the homepage for which is here”</p>

liamhuber: <p>Mostly I meant the npoints, which I assume are being changed to make things run a bit faster for this demo and we should be explicit about that.</p><p> </p><p>but now this is a good point, why <em>are </em>you changing the volume range? Isn’t the default 0.1 good enough? It might make one of the fit types perform less well later, but I don’t really see a reason to change it. It feels a bit weird if we’re changing a default sans necessity</p>

samwaseda: Ah, I thought the intention was to show how to change those parameters, because after all, it's a tutorial, so it should cover various scenarios. There was already the phrase "For this tutorial, we change the input parameters as follows", so I extended it by "but in principle you can run the simulation without touching the input", which I hope tells there's no need to change those parameters if the user doesn't want to.

jan-janssen:
> Based on [this discussion](https://github.com/pyiron/pyiron_atomistics/issues/1014), I updated the energy-volume curve notebook. In particular, I removed all the manual part in the beginning, because it appears extremely technical, especially only to be replaced by the pyiron Murnaghan module anyway. 

I am a bit confused by this statement. The intention of the previous tutorial in addition to the virtual course is to demonstrate how for loops that the users calculate these days translate into GenericMaster Jobs:
https://pyiron.org/phasediagram-workshop-2020/Exercise3.html
This not only applies to the calculation of energy volume curves but is more a general concept that pyiron provides multiples levels for users to use pyiron depending on their skill level and prior experience. Especially we do not want to convey the perspective that pyiron is just a black box, which hides the scientific part. 

Removing this part from the functionality of this notebook and hence the tests is very dangerous from my perspective especially as this is discussed in so many places, starting with the initial paper, the virtual course mentioned above and even this week I just gave a tutorial discussing exactly the same functionality again.

notebooks/energy_volume_curve.ipynb
liamhuber: Ah, that's a good idea too. Why don't we do both: "You can change these defaults in the same way as setting new values in a dictionary. In this case, let's decrease the number of volumes we run calculations for to speed the demo up:" (then reset only num points)

jan-janssen:
> ok can we then do a polynomial fit after launching a set of simulations, without using Murnaghan? The cryptic part is the fitting part, which loads functions from the Murnaghan class out of nowhere. It sounds somewhat dull, but at least we can say "hey you can use the Murnaghan class to do all these tasks in one go".

That sounds like a compromise to me.

liamhuber:
@samwaseda please revert the changes to data_mining.ipynb; it looks like this was just a case of the notebook being accidentally open/run/saved/whatever and the notebook metadata updating, not actually part of this PR.

liamhuber (approved):
Looks great! The addition of the by-hand loop calculation following @jan-janssen's comments was a good addition. This also nicely gives context for why we update the murn input to 7 and 5%.

Please pull off the accidental data mining changes, but otherwise this is ready to go!

PR #1495, opened 2023-06-06
Fix queue.yaml example and refer to pysqa documentation.

niklassiemer:

PR #1498, opened 2023-06-08
Fix queue.yaml example and refer to pysqa documentation.

ligerzero-ai:
https://github.com/pyiron/pyiron_base/issues/1114 fixes this (hopefully)

https://github.com/pyiron/pyiron/pull/1495 and makes this one redundant - @niklassiemer


niklassiemer (approved):
Ah, now I start to get it. Right now all the different pages are actually just from one file... I did not expect that TBH. I do not like the redundancy in that case...
As long as this fixes the issue, I am fine with this :)

PR #1500, opened 2023-06-22
Remove pypi support for python versions not allowed by conda

liamhuber:
Synchronizes the pypi representation of pyiron with the hard limit of python >3.8 expressed on [the conda forge feedstock](https://github.com/conda-forge/pyiron-feedstock/blob/c954babdba3bcd313738db2feeae0de6412a9123/recipe/meta.yaml#L24)


srmnitc:
Looks all okay to me.

PR #1508, opened 2023-09-11
Versioning and api

liamhuber:
Just some adjustments to the "contributing" docs


niklassiemer (approved):
This entirely dropped my notion. Changes LGTM. The only 'concern' I have is that we have probably to many places such information should be added... We need some concept about abstract/pyiron specific/package specific documentation and get them together in a sensible way for each package...

Issue #1516, opened 2023-11-14
Outdated (pr.create) commands in tutorials


pkruzikova:
In one of the older [workshops](https://pyiron.org/phasediagram-workshop-2020/Introduction.html), there is information about an update including pyiron modules (such as atomistic,...) and `pr.create` for structure, job, table, etc. object creation. This is in the exercises where new commands are used, as the older commands (e.g. `pr.create_ase_bulk()`) are used in the videos. Some of the other workshops also (understandably) feature the older ones, as well as the example in GitHub pyiron _readme_, and the [tutorials](https://pyiron.readthedocs.io/en/pyiron-0.3.23/source/examples.html), where there is no mention of the update.

The older commands do not cause the code to fail - the structure ones throw a depreciation warning, and the job ones seem to work exactly as the updated ones, with no warnings. 
Still, it would be useful (especially if the older commands are not to work in the future) to update the tutorials accordingly.

PR #1517, opened 2023-11-14
Replace deprecated `pr.create_job` in `first_steps.ipynb`

samwaseda:
Mentioned in [this issue](https://github.com/pyiron/pyiron/issues/1516)


samwaseda:
To be honest it's a horrible idea to show an example to heat up Aluminium to 1,000 K...

pkruzikova (approved):
Looks good to me. I am assuming the replacing pr.create for structures will be done elsewhere (if planned) as it is not in the PL title, and it's good to see the warnings about it now.

notebooks/first_steps.ipynb
```diff
@@ -41,9 +41,30 @@
...
+      },
+      "text/plain": []
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
    "source": [
-    "from pyiron import Project"
+    "from pyiron_atomistics import Project"
```
niklassiemer, line 67: Since we are testing this from the `pyiron` repo, I would stick to `from pyiron import Project`

samwaseda: Hmmm I see your point, but `pyiron` takes so much more time that it might be a bit discouraging to see this in the very first line

niklassiemer: Yes, but this is the only test-space to check that the `from pyiron import` is working correctly! Maybe we have to make the lookup faster if it is just too annoying? Or is it because you also have all the other pyiron modules installed which are not loaded in a nice way (especially contrib...). I mean if only
```
conda create --name pyiron pyiron
```
was used it should essentially only find and load `pyiton_atomistics` at a moderate overhead?

liamhuber: No, I'm with Niklas here -- this is basically the only place that we might reasonably `from pyiron import...` and then expect that to be providing us with sub module support (atomistics, continuum, etc). So if that's something we want to provide, IMO this is the right place to test it. The fact it's painfully slow is related, but a separate issue.

samwaseda: > The fact it's painfully slow is related, but a separate issue.

Well, as long as we cannot solve this problem, it doesn't change the fact that this is the very first line of the very first notebook that we present on our website, and it is extremely demoralising and unsettling to see a notebook frozen for a significant amount of time in the first cell.

Anyway shouldn't we anyway have integration tests if we want to test `from pyiron import Project`?

liamhuber: The notebooks run on the CI, so I feel they _are_ the integration tests (albeit by another name and also wearing at least one other hat.)

If importing from pyiron is really such a pain that we don't want people to do it, and we can't fix it, then perhaps we should stop supporting that syntax?

samwaseda: They run on the CI, but they are at the same time the example notebooks of the website. For me the CI in this case is there only to guarantee that the example notebooks run, but not really to test pyiron.

I don't think we should stop using `from pyiron import Project`. Instead, I think it's a rather special case where a user wants to use tools across different length scales. After all, we introduced `pyiron_atomistics` to make it possible to import only the modules for the users who are only interested in atomistic tools, and since the notebook is all about atomistic calculations, I think it's correct to present only `pyiron_atomistic` at this point.

samwaseda: This being said, I don't have a very strong feeling for this, so I can also put it back if you feel strongly that it should be there.

liamhuber: Those are fair points. I still come down on the side of leaving it as from pyiron, but you've sowed enough doubt that I also don't feel strongly. @niklassiemer I'll leave the responsibility of tie breaking to you, but it doesn't sound like either Sam or I will write you out of the will for your choice 😂

niklassiemer: I am in favor of leaving it `from pyiron` here. However, it might be reasonable to use the notebooks from the `pyiron_atomistics` repo for the education on the website.
Furthermore, we probably should raise an issue for the slow import. 
Some solution could be:
- not import but only find other `pyiron_modules`  than pyiron_atomistics on import of pyiron and import those upon access.
- provide a module scoped list (dict? ) which is used to omit the finding in the first place, i.e. a list like `['contrib', 'gui', ... ]` which is used for autocompletion and to extend this to the corresponding pyiron_module. If a not listed access pyiron.some_mod is performed try to `import pyiron_some_mod` and use this.
- stop treating pyiron as drop in replacement of pyiron_atomistics and actually use pyiron.atomistics (probably together with one of the first). Breaks backwards compatibility, though!

samwaseda: > I am in favor of leaving it `from pyiron` here. However, it might be reasonable to use the notebooks from the `pyiron_atomistics` repo for the education on the website.

ok that means I should copy this notebook to `pyiron_atomistics` and create an integration tests folder where we'd have essentially the same content with `from pyiron import Project`?

niklassiemer: I did not check the exact content, but aren't the notebooks here exactly such copies from pyiron_atomistics with the `from pyiron` to `from pyiron_atomistics` change? It is a not so nice setup, since near duplication, i.e. please apply the same changes over in pyiron_atomistics 😇

niklassiemer: And I really meant for the purpose of displaying on the webpage. Right now, the tutorial is from this repo, while you claim it to be much faster if we use the notebooks from pyiron_atomistics for the webpage.

samwaseda: > I did not check the exact content, but aren't the notebooks here exactly such copies from pyiron_atomistics with the `from pyiron` to `from pyiron_atomistics` change?

AHA!! I didn't know that it existed TWICE!

> And I really meant for the purpose of displaying on the webpage.

I'm not really sure if I understand what you are referring to. Do I understand it correctly:

- We keep `from pyiron import Project` in this repo
- On the website we use the version from `pyiron_atomistics`, i.e. the same notebook with `from pyiron_atomistics import Project`?

PR #1534, opened 2024-01-19
Update installation.rst

samwaseda:
There were comments coming from PMD project partners, saying it's difficult to understand what the examples here actually do. I removed the parts which are not related to the actual queueing system installation.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -163,15 +163,11 @@ Install the conda-packaged version of LAMMPS:
     import os
 
     pr = Project("test_lammps")
-    basis = pr.create.structure.bulk('Al', cubic=True)
-    supercell_3x3x3 = basis.repeat([3, 3, 3])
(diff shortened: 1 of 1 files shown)


niklassiemer (commented):
In general updating the commands is good! However, do we want to teach everyone not to think about potentials any more?

samwaseda:
> In general updating the commands is good! However, do we want to teach everyone not to think about potentials any more?

I can also think of putting it in the comment. The point is that each unrelated line makes it more difficult to understand what this document tries to tell, so I have the feeling that anything unrelated we can remove should be removed.

ligerzero-ai (approved):
Lgtm

niklassiemer:
> > In general updating the commands is good! However, do we want to teach everyone not to think about potentials any more?
> 
> I can also think of putting it in the comment. The point is that each unrelated line makes it more difficult to understand what this document tries to tell, so I have the feeling that anything unrelated we can remove should be removed.

Fair point! However, a comment that this code will raise a warning would be good such that new users are not confused...

